PR #10, opened 2023-07-25
Enhancement/mvp 176 standardize tag prefix scan searches

eduwercamacaro:
First steps for improve Search mechanism.

`PublicScanRequest` types usually do this logic on tag scan:

1. Find if the search should be Local or Remote
2. Set storage name (Core or repartition)
3. Build attributes from proto Search class
4. Create tag prefix from attributes
5. handle bookmark and timestamps.

The intention for this change is to improve step 3 and 4 in a generic way.


sauljabin (approved):

schemas/internal_server.proto
```diff
@@ -85,9 +85,9 @@ message BookmarkPb { // Used for paginated responses
 
 message InternalScanPb {
     message TagScanPb {
-        repeated AttributePb attributes = 1;
-        optional google.protobuf.Timestamp earliest_create_time = 2;
-        optional google.protobuf.Timestamp latest_create_time = 3;
+        optional google.protobuf.Timestamp earliest_create_time = 1;
```
MateoRojas, line 88: This is fine for now, but after we release for Elevance we would need to ensure compatibility. 

P.S. Tech Huddle question: How can we ensure that we have automated contract tests.

schemas/internal_server.proto
```diff
@@ -85,9 +85,9 @@ message BookmarkPb { // Used for paginated responses
...
+        optional google.protobuf.Timestamp latest_create_time = 2;
+        string key_prefix = 3;
```
MateoRojas, line 90: What does key_prefix stand for? Maybe attribute string?

server/src/main/java/io/littlehorse/server/streamsimpl/lhinternalscan/publicrequests/SearchWfRun.java
```diff
@@ -124,55 +124,46 @@ private List<String> searchAttributes() {
...
         }
         if (getLatestStart() != null) {
             tagScanBuilder.setLatestCreateTime(getLatestStart());
         }
+        tagScanBuilder.setKeyPrefix(getSearchAttributeString());
 
         out.setTagScan(tagScanBuilder.build());
         out.setType(ScanBoundaryCase.TAG_SCAN);
-        new WfRun()
+        TagStorageTypePb storageTypePb = new WfRun()
```
MateoRojas, line 142: Is it better to call it `indexTypeForSearch`?

server/src/main/java/io/littlehorse/server/streamsimpl/lhinternalscan/publicrequests/SearchWfRun.java
```diff
@@ -124,55 +124,46 @@ private List<String> searchAttributes() {
...
-                    out.setPartitionKey(
-                        Tag.getAttributeString(
-                            Getable.getTypeEnum(WfRun.class),
-                            buildTagAttributes()
-                        )
-                    );
-                }
-            });
+            .orElse(null);
+        if (storageTypePb != null) {
```
MateoRojas, line 155: What happens if it's null?

MateoRojas (approved):
Reviewed with @coltmcnealy-lh :)

Good to go with just some minor comments and questions

schemas/internal_server.proto
eduwercamacaro: Key prefix is the RocksDb key prefix that will be used by `BackendInternalCommons#doScan` to perform the scan operation on  the database.

server/src/main/java/io/littlehorse/server/streamsimpl/lhinternalscan/publicrequests/SearchWfRun.java
eduwercamacaro: Addressed in other branch

PR #12, opened 2023-07-26
Enhancement/mvp 176 migrate to new search mechanism

eduwercamacaro:
Another step in order to improve Search mechanism.

PublicScanRequest types usually do this logic on tag scan:

- Find if the search should be Local or Remote
- Set storage name (Core or repartition)
- Build attributes from proto Search class
- Create tag prefix from attributes
- handle bookmark and timestamps.

step 3 and 4 were addressed by this [PR](https://github.com/littlehorse-enterprises/littlehorse/pull/10)

The intention for this PR is reduce the logic that we have in `PublicScanRequest` types. This PR only refactors for Search User Task Run and WfRun. next PR will refactor other types of searches.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streamsimpl/lhinternalscan/publicrequests/ListNodeRuns.java
+++ b/server/src/main/java/io/littlehorse/server/streamsimpl/lhinternalscan/publicrequests/ListNodeRuns.java
@@ -14,6 +14,7 @@ import io.littlehorse.sdk.common.proto.ListNodeRunsReplyPb;
 import io.littlehorse.sdk.common.proto.NodeRunPb;
 import io.littlehorse.server.streamsimpl.ServerTopology;
 import io.littlehorse.server.streamsimpl.lhinternalscan.InternalScan;
+import io.littlehorse.server.streamsimpl.lhinternalscan.ObjectIdScanBoundaryStrategy;
 import io.littlehorse.server.streamsimpl.lhinternalscan.PublicScanRequest;
 import io.littlehorse.server.streamsimpl.lhinternalscan.SearchScanBoundaryStrategy;
 import io.littlehorse.server.streamsimpl.lhinternalscan.publicsearchreplies.ListNodeRunsReply;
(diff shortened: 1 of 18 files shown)


server/src/main/java/io/littlehorse/server/streamsimpl/lhinternalscan/publicrequests/SearchNodeRun.java
```diff
@@ -82,38 +79,22 @@ public static SearchNodeRun fromProto(SearchNodeRunPb proto) {
...
 
     @Override
-    public SearchScanBoundaryStrategy getScanBoundary(String searchAttributeString) {
-        return null;
+    public SearchScanBoundaryStrategy getScanBoundary(String searchAttributeString)
+        throws LHValidationError {
+        if (type == NoderunCriteriaCase.WF_RUN_ID) {
+            return new ObjectIdScanBoundaryStrategy(wfRunId);
+        } else {
+            throw new LHValidationError("Yikes, unimplemented type: " + type);
```
sauljabin, line 97: What if: throw new LHValidationError("Unimplemented type: " + type);

sauljabin (approved):

server/src/main/java/io/littlehorse/server/streamsimpl/lhinternalscan/publicrequests/ListNodeRuns.java
```diff
@@ -40,31 +37,17 @@ public GetableClassEnumPb getObjectType() {
...
+        return TagStorageTypePb.LOCAL;
     }
 
     @Override
     public void validate() throws LHValidationError {}
 
     @Override
     public SearchScanBoundaryStrategy getScanBoundary(String searchAttributeString) {
-        return null;
+        return new ObjectIdScanBoundaryStrategy(wfRunId);
```
coltmcnealy-lh, line 51: Quick question. I don't quite see where we tell the Scan Boundary Strategy that we are looking for `ScanResultType.OBJECT` here as opposed to `ScanResultType.OBJECT_ID`. Where does that happen?

coltmcnealy-lh (approved):
+290, -533 is always good

server/src/main/java/io/littlehorse/server/streamsimpl/lhinternalscan/publicrequests/ListNodeRuns.java
eduwercamacaro: here:

https://github.com/littlehorse-enterprises/littlehorse/pull/12/files#diff-8e78027dbb3fb2fdad58b82d87d187b45f3d5c4c4f91e31defa0be3612b7d834R43

PR #13, opened 2023-07-27
Feature/mvp 190 cache wfspec lhdao

coltmcnealy-lh:


server/src/main/java/io/littlehorse/server/streamsimpl/util/LHCache.java
```diff
@@ -0,0 +1,32 @@
+package io.littlehorse.server.streamsimpl.util;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+
+public class LHCache<K, V> {
+
+    private final ConcurrentHashMap<K, V> cache = new ConcurrentHashMap<>();
```
sauljabin, line 8: Why not to use https://www.baeldung.com/guava-cache

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
```diff
@@ -976,6 +987,7 @@ private void clearThingsToWrite() {
         wfRunPuts.clear();
         wfSpecPuts.clear();
         taskDefPuts.clear();
+        taskRunPuts.clear();
```
sauljabin, line 990: for now this is ok, but if we want to prevent for happening again I recommend to use a **queue** or a similar collection where all objects (variablePuts, nodeRunPuts, .....) , it is not a cache is more like a queue of stuffs to be committed.

server/src/main/java/io/littlehorse/server/streamsimpl/ServerTopology.java
```diff
@@ -75,7 +76,7 @@ public static Topology initCoreTopology(
         KafkaStreamsServerImpl server
     ) {
         Topology topo = new Topology();
-
+        WfSpecCache wfSpecCache = new WfSpecCache();
```
sauljabin, line 79: good stuff

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
```diff
@@ -274,15 +279,21 @@ public void putTaskDef(TaskDef spec) {
...
-        LHROStoreWrapper store = isHotMetadataPartition ? localStore : globalStore;
-        WfSpec out;
-        if (version != null) {
-            out = store.get(new WfSpecId(name, version).getStoreKey(), WfSpec.class);
-        } else {
-            out = store.getLastFromPrefix(name, WfSpec.class);
-        }
-        if (out != null) out.setDao(this);
-        return out;
+        Supplier<WfSpec> findWfSpec = () -> {
```
sauljabin, line 282: findWfSpecFromStore (maybe?)

eduwercamacaro (approved):

sauljabin (approved):
I like it but there is not streams 😢

sauljabin:
0ne test failed

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
coltmcnealy-lh: Yeah, we probably could have just used one thing for all of the types.

PR #18, opened 2023-07-31
Fix remote tags search

eduwercamacaro:
Last PR introduced a bug for `LHROStoreWrapper#prefixTagScanStream`. This PR remove the support for Streams API in  `LHROStoreWrapper` class a `BackendInternalCommons`  use the pre-existing scan method.


server/src/main/java/io/littlehorse/server/streamsimpl/BackendInternalComms.java
```diff
@@ -858,22 +858,28 @@ private InternalScanReplyPb specificPartitionTagScan(InternalScan search) {
...
-                            objectId,
-                            Getable.getIdCls(search.getObjectType())
-                        )
-                        .toProto()
-                        .build()
-                        .toByteString();
-                })
-                .collect(Collectors.toList());
-            out.addAllResults(result);
+            LHKeyValueIterator<Tag> tagScanResultIterator = store.prefixScan(
```
coltmcnealy-lh, line 861: do you call `tagScanResultIterator.close()`? If you don't, then we leak resources on the underlying RocksDB process, which is dangerous.

I think the TagScanStream thing didn't close it either...

coltmcnealy-lh (approved):
Approved if you please put the iterator in a `try / finally` block. Thanks!

coltmcnealy-lh (approved):

PR #23, opened 2023-08-03
Fixes minor SDK bugs:

coltmcnealy-lh:
- missing edge on while loop conditionals
- ensure no outgoing edge on EXIT nodes

Also fixes unit tests that weren't actually checking anything.

### DIFF
--- a/examples/basic/src/main/java/io/littlehorse/examples/BasicExample.java
+++ b/examples/basic/src/main/java/io/littlehorse/examples/BasicExample.java
@@ -1,12 +1,8 @@
 package io.littlehorse.examples;
 
 import io.littlehorse.sdk.client.LHClient;
-import io.littlehorse.sdk.common.LHLibUtil;
 import io.littlehorse.sdk.common.config.LHWorkerConfig;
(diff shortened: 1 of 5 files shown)


examples/basic/src/main/java/io/littlehorse/examples/BasicExample.java
```diff
@@ -57,6 +61,31 @@ public static LHTaskWorker getTaskWorker(LHWorkerConfig config) {
     }
 
     public static void main(String[] args) throws IOException, LHApiError {
+        WorkflowImpl wf = new WorkflowImpl(
+            "asdf",
```
eduwercamacaro, line 65: `basic-example` ??

eduwercamacaro (approved):

examples/basic/src/main/java/io/littlehorse/examples/BasicExample.java
coltmcnealy-lh: Hoboy I did NOT mean to change this.

PR #24, opened 2023-08-04
Feature/pass variables to task context

eduwercamacaro:
Added task context for every task execution. now you can access to the task context using `WorkerContext#getTaskContext` method. In the task context, you will find variables associated to the task execution


server/src/main/java/io/littlehorse/common/model/wfrun/ThreadRun.java
```diff
@@ -666,9 +666,6 @@ public void activateNode(Node node) {
 
         cnr.arrivalTime = arrivalTime;
         cnr.wfSpecId = wfRun.getWfSpec().getObjectId();
-        cnr.nodeName = node.name;
```
eduwercamacaro, line 669: @coltmcnealy-lh just cleaning this because i see it is duplicated

PR #25, opened 2023-08-04
Optimization 2/incremental combine tags cache getable

MateoRojas:
In this first optimization PR we wanted to achieve two things:

1. Store the entities together with the tags cache to reduce 1 get and 1 put every time we inserted/updated something
2. Reduce the complexity of the KafkaStreamsLHDAOImpl by moving the "ORM"/"Transactional" logic into the GetableStorageManager

At first we tried to do this optimization/refactor in a big bang way but we encountered a lot of compilation errors and around 30+ files with changes. So we decided to do it incrementally by first introducing it to the "Run" entities. This PR introduces the optimization to the NodeRun, TaskRun, UserTaskRun, WfRun, and TaskWorkerGroup.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
+++ b/server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
@@ -196,7 +196,7 @@ import io.littlehorse.server.streamsimpl.storeinternals.utils.StoreUtils;
 import io.littlehorse.server.streamsimpl.taskqueue.PollTaskRequestObserver;
 import io.littlehorse.server.streamsimpl.taskqueue.TaskQueueManager;
 import io.littlehorse.server.streamsimpl.util.GETStreamObserver;
-import io.littlehorse.server.streamsimpl.util.GETStreamObserverPepe;
+import io.littlehorse.server.streamsimpl.util.GETStreamObserverNew;
 import io.littlehorse.server.streamsimpl.util.POSTStreamObserver;
 import java.io.File;
 import java.io.IOException;
@@ -566,7 +566,7 @@ public class KafkaStreamsServerImpl extends LHPublicApiImplBase {
 
     @Override
     public void getWfRun(WfRunIdPb req, StreamObserver<GetWfRunReplyPb> ctx) {
-        StreamObserver<CentralStoreQueryReplyPb> observer = new GETStreamObserverPepe<>(
+        StreamObserver<CentralStoreQueryReplyPb> observer = new GETStreamObserverNew<>(
             ctx,
             WfRun.class,
             GetWfRunReplyPb.class,
@@ -584,7 +584,7 @@ public class KafkaStreamsServerImpl extends LHPublicApiImplBase {
 
     @Override
     public void getNodeRun(NodeRunIdPb req, StreamObserver<GetNodeRunReplyPb> ctx) {
-        StreamObserver<CentralStoreQueryReplyPb> observer = new GETStreamObserverPepe<>(
+        StreamObserver<CentralStoreQueryReplyPb> observer = new GETStreamObserverNew<>(
             ctx,
             NodeRun.class,
             GetNodeRunReplyPb.class,
@@ -608,7 +608,7 @@ public class KafkaStreamsServerImpl extends LHPublicApiImplBase {
 
     @Override
     public void getTaskRun(TaskRunIdPb req, StreamObserver<GetTaskRunReplyPb> ctx) {
-        StreamObserver<CentralStoreQueryReplyPb> observer = new GETStreamObserverPepe<>(
+        StreamObserver<CentralStoreQueryReplyPb> observer = new GETStreamObserverNew<>(
             ctx,
             TaskRun.class,
             GetTaskRunReplyPb.class,
@@ -630,7 +630,7 @@ public class KafkaStreamsServerImpl extends LHPublicApiImplBase {
         UserTaskRunIdPb req,
         StreamObserver<GetUserTaskRunReplyPb> ctx
     ) {
-        StreamObserver<CentralStoreQueryReplyPb> observer = new GETStreamObserverPepe<>(
+        StreamObserver<CentralStoreQueryReplyPb> observer = new GETStreamObserverNew<>(
             ctx,
             UserTaskRun.class,
             GetUserTaskRunReplyPb.class,
--- a/server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
+++ b/server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
@@ -608,7 +608,7 @@ public class KafkaStreamsServerImpl extends LHPublicApiImplBase {
 
     @Override
     public void getTaskRun(TaskRunIdPb req, StreamObserver<GetTaskRunReplyPb> ctx) {
-        StreamObserver<CentralStoreQueryReplyPb> observer = new GETStreamObserver<>(
+        StreamObserver<CentralStoreQueryReplyPb> observer = new GETStreamObserverPepe<>(
             ctx,
             TaskRun.class,
             GetTaskRunReplyPb.class,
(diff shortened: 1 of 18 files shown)


server/src/main/java/io/littlehorse/server/streamsimpl/util/GETStreamObserverNew.java
MateoRojas, line 1: I couldn't find a better name for this class :(
The idea is for this class to completely replace the GETStreamObserver class once every entity starts using the new entity wrapper (StoredGetable). Any idea for a better name is welcomed. 
cc: @coltmcnealy-lh @eduwercamacaro

server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/utils/StoredGetable.java
MateoRojas, line 1: I don't love the name `StoredGetable`, maybe we can come up with a better name. The idea of this class is to be a wrapper of the entity that allows to stored common stuff across all different entities (e.g. TagsCache). Maybe we can name it some along the lines of `EntityWrapper` or something.
cc: @coltmcnealy-lh @eduwercamacaro @sauljabin

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
```diff
@@ -68,18 +65,13 @@
 @Slf4j
 public class KafkaStreamsLHDAOImpl implements LHDAO {
 
-    private Map<String, NodeRun> nodeRunPuts;
     private Map<String, Variable> variablePuts;
```
MateoRojas, line 68: I was not able to use the `StoredGetable` with the `Variable`. When I made the change a bunch of things started to fail, so I decided to delay the migration for later.

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
```diff
@@ -68,18 +65,13 @@
...
     private Map<String, ExternalEvent> extEvtPuts;
```
MateoRojas, line 69: `ExternalEvent` was also hard to migrate because of the logic inside `getUnclaimedEvent()` method, so I also delayed this migration. I might need some help from someone who understands that logic :)

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
```diff
@@ -520,22 +490,23 @@ public void abortChanges() {
     // CommandProcessor#process().
     @Override
     public void abortChangesAndMarkWfRunFailed(String message) {
```
MateoRojas, line 492: Sorry, seems like I forgot to uncomment this piece of code. Let me fix it :)

server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
```diff
@@ -196,6 +196,7 @@
 import io.littlehorse.server.streamsimpl.taskqueue.PollTaskRequestObserver;
 import io.littlehorse.server.streamsimpl.taskqueue.TaskQueueManager;
 import io.littlehorse.server.streamsimpl.util.GETStreamObserver;
+import io.littlehorse.server.streamsimpl.util.GETStreamObserverNew;
```
coltmcnealy-lh, line 199: To refresh my memory, this is basically the same change that I made on my branch, but in a new class so it still works with the Metadata, because we didn't migrate the metadata over yet to the new TagsCache storage mechanism?

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
coltmcnealy-lh: Got it, yes, my branch has an implementation of that logic with the `GetableStorageManager` already, but it wasn't hooked up to the rest of the code. We can pair on that together.

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
```diff
@@ -163,69 +150,55 @@ public KafkaStreamsLHDAOImpl(
...
     }
 
     @Override
     public NodeRun getNodeRun(String wfRunId, int threadNum, int position) {
         String key = new NodeRunId(wfRunId, threadNum, position).getStoreKey();
-        if (nodeRunPuts.containsKey(key)) {
-            return nodeRunPuts.get(key);
+        NodeRun nodeRun = storageManager.get(key, NodeRun.class);
+        if (nodeRun != null) {
+            nodeRun.setDao(this);
```
coltmcnealy-lh, line 169: Can we put the `setDao()` inside the `GetableStorageManager`?

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
```diff
@@ -163,69 +150,55 @@ public KafkaStreamsLHDAOImpl(
...
     }
 
     @Override
     public TaskRun getTaskRun(TaskRunId taskRunId) {
         String key = taskRunId.getStoreKey();
-        if (taskRunPuts.containsKey(key)) {
-            return taskRunPuts.get(key);
+        TaskRun taskRun = storageManager.get(key, TaskRun.class);
+        if (taskRun != null) {
+            taskRun.setDao(this);
```
coltmcnealy-lh, line 184: Same here

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
```diff
@@ -163,69 +150,55 @@ public KafkaStreamsLHDAOImpl(
...
-        if (userTaskRunPuts.containsKey(key)) {
-            return userTaskRunPuts.get(key);
-        }
-        UserTaskRun out = localStore.get(key, UserTaskRun.class);
-        if (out != null) {
-            userTaskRunPuts.put(key, out);
-            out.setDao(this);
+        UserTaskRun userTaskRun = storageManager.get(key, UserTaskRun.class);
+        if (userTaskRun != null) {
+            userTaskRun.setDao(this);
```
coltmcnealy-lh, line 199: Same here

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
```diff
@@ -699,7 +676,7 @@ private void deleteAllChildren(WfRun wfRun) {
         ) {
             while (iter.hasNext()) {
                 LHIterKeyValue<ExternalEvent> next = iter.next();
-                getableStorageManager.delete(next.getKey(), ExternalEvent.class);
+                storageManager.deleteGetable(next.getKey(), ExternalEvent.class);
```
coltmcnealy-lh, line 679: what's the difference between deleteGetable and delete?

server/src/main/java/io/littlehorse/server/streamsimpl/util/GETStreamObserverNew.java
coltmcnealy-lh: Let's leave it as `GETStreamObserverNew` for now so that we know to remove the old one when this is ready.

server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/utils/StoredGetable.java
coltmcnealy-lh: Hmm, this class is supposed to be a `Storeable` (which it is), and it's supposed to only be used for Getable's. We store other things in the store, like:
* ScheduledTask
* Tag
* TaskMetricUpdate
* WfMetricUpdate
and those things aren't Getable's, so they don't have Tags, and as such don't go in here.

coltmcnealy-lh (approved):

server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
MateoRojas: Yup, yup exactly that

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/KafkaStreamsLHDAOImpl.java
MateoRojas: I couldn't find a better name but basically the `delete` works with the new StoredGetable and the `deleteGetable` works with the metadata that's not using the StoredGetable yet

PR #29, opened 2023-08-05
Minor:

coltmcnealy-lh:
- fix the counting of number of rocksdb gets in each Command.
- no longer put a commandId onthe Timer Command's which don't have a response observer. This reduces load on the AsyncWaiters.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/LHStoreWrapper.java
+++ b/server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/LHStoreWrapper.java
@@ -43,7 +43,6 @@ public class LHStoreWrapper extends LHROStoreWrapper {
 
     private void put(String storeKey, Storeable<?> thing) {
         totalPuts++;
-        log.trace("Putting {}", storeKey);
         store.put(storeKey, new Bytes(thing.toBytes(config)));
     }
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/LHStoreWrapper.java
```diff
@@ -71,6 +80,7 @@ public void delete(String fullStoreKey) {
 
     public Bytes getRaw(String rawKey) {
         totalGets++;
+        log.warn("Getting: {}", rawKey);
```
sauljabin, line 83: warn?

sauljabin (approved):

server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/LHStoreWrapper.java
coltmcnealy-lh: Whoops.

coltmcnealy-lh: Anyways, I think we should not log (even trace) keys since the Tag keys for Variables might include sensitive data.

PR #32, opened 2023-08-06
minor changes to make red hat's java plugin happier

coltmcnealy-lh:
It appears that IntelliJ and VSCode use different java linters. This fixes some errors and warnings that were present (and quite annoying) in the VSCode one.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/GetableStorageManager.java
+++ b/server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/GetableStorageManager.java
@@ -17,7 +17,7 @@ import org.apache.kafka.streams.processor.api.ProcessorContext;
 @Slf4j
 public class GetableStorageManager {
 
-    private final Map<String, StoredGetable<? extends Message, ? extends Getable<?>>> uncommittedChanges;
+    private final Map<String, StoredGetable<?, ?>> uncommittedChanges;
 
     private final LHStoreWrapper localStore;
 
@@ -108,7 +108,6 @@ public class GetableStorageManager {
         uncommittedChanges.put(getable.getStoreKey(), toPut);
     }
 
-    @SuppressWarnings("unchecked")
     public <U extends Message, T extends Getable<U>> void delete(
         String key,
         Class<T> clazz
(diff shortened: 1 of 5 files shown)


server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/GetableStorageManager.java
```diff
@@ -17,7 +17,7 @@
 @Slf4j
 public class GetableStorageManager {
 
-    private final Map<String, StoredGetable<? extends Message, ? extends Getable<?>>> uncommittedChanges;
+    private final Map<String, StoredGetable<?, ?>> uncommittedChanges;
```
MateoRojas, line 20: Yeah, you are right `Map<String, StoredGetable<? extends Message, ? extends Getable<?>>>` is probably redundant considering that StoredGetable already specifies the types it can receive.

server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/GetableStorageManager.java
```diff
@@ -175,7 +178,7 @@ private <U extends Message, T extends Getable<U>> void insertIntoStore(
 
     private <U extends Message, T extends Getable<U>> void deleteFromStore(
         String key,
-        StoredGetable<U, T> getable
+        StoredGetable<?, ?> getable
```
MateoRojas, line 181: If you do `StoredGetable<?, ?>`, this method (`deleteFromStore`) doesn't need to use generics anymore, so we should be able to get rid off `<U extends Message, T extends Getable<U>>`

MateoRojas (approved):

server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/GetableStorageManager.java
coltmcnealy-lh: If we don't put the `<?, ?>` then VSCode gives a warning ):

coltmcnealy-lh:
> What errors are you getting on VSCode? Just wondering so that we are more careful in the future :)

```
The method insertIntoStore(StoredGetable<U,T>) in the type GetableStorageManager is not applicable for the arguments (StoredGetable<capture#20-of ? extends Message,capture#21-of ? extends Getable<?>>>
```

It's a weird one...it compiles, but VSCode not happy.

server/src/main/java/io/littlehorse/server/streamsimpl/storeinternals/GetableStorageManager.java
MateoRojas: I meant to move from 

`private <U extends Message, T extends Getable<U>> void deleteFromStore(....) {}`

to:

`private void deleteFromStore(...) {}`

PR #33, opened 2023-08-06
Minor/improve healthchecks

coltmcnealy-lh:
Refactors health checks:
 - Creates HealthService.java which has a Javalin service Prometheus metrics and also healthchecks
 - Adds a readiness check which tells whether the service can answer requests
 - Adds a liveness check which tells whether the service should be restarted
 
The readiness and liveness checks run on the same Javalin server as the Prometheus exporter.


docs/SERVER_CONFIGURATIONS.md
```diff
@@ -561,20 +561,40 @@ creating the `ExternalEventDef`.
...
+The port that the healthchecks and prometheus metrics are exposed on.
 
 - **Type:** int
 - **Default:** 1822
 - **Importance:** low
 
 ---
 
-### `LHS_PROMETHEUS_EXPORTER_PATH`
+### `LHS_HEALTH_PATH_PROMETHEUS`
```
sauljabin, line 574: LHS_HEALTH_PATH_PROMETHEUS maybe LHS_HEALTH_PATH_METRICS?

server/src/main/java/io/littlehorse/server/metrics/PrometheusMetricExporter.java
```diff
@@ -69,19 +62,10 @@ public void bind(Map<String, KafkaStreams> streams) {
...
-        log.info("Prometheus started");
-    }
-
-    private Handler handleRequest() {
+    public Handler handleRequest() {
         return ctx -> {
-            log.debug(
-                "Request [from={}, path={}, method={}]",
+            log.trace(
+                "Metrics Request [from={}, path={}, method={}]",
```
sauljabin, line 68: we agreed on removing this log injection

sauljabin (approved):

server/src/main/java/io/littlehorse/server/metrics/PrometheusMetricExporter.java
coltmcnealy-lh: Good catch, I forgot to push the change.

docs/SERVER_CONFIGURATIONS.md
coltmcnealy-lh: Ok I can change it. Will change on the Operator too.

PR #35, opened 2023-08-07
Bugfix/search wfspec and taskdef

eduwercamacaro:
Solved bug on `lhctl search wfSpc` and `lhctl search taskDef` searches


coltmcnealy-lh (approved):
Thank you!

PR #37, opened 2023-08-09
Variables tag cache optimization

MateoRojas:
Starts using the StoredGetable object to wrap Variable entities on the database. We needed to modify how the variables are created and mutated to prevent some bugs and avoid having to include one-of logic on the GetableStorageManager "transaction" logic.


server/src/main/java/io/littlehorse/common/model/wfrun/ThreadRun.java
```diff
@@ -885,17 +883,22 @@ public NodeRun getNodeRun(int position) {
...
 
-    public void putVariable(String varName, VariableValue var) throws LHVarSubError {
+    /**
+     * Traverse the current and parent ThreadRun to run the variable modification on the appropriate ThreadRun
+     * @param varName name of the variable
+     * @param function function that executes the variable modification (e.g. creation, mutation, etc.)
+     * @throws LHVarSubError when the varName is not found either on the current ThreadRun definition
+     * or its parents definition
+     */
+    private void applyOnAppropriateThread(
```
MateoRojas, line 893: Is this to complicated or hard to understand? Since we divided the `putVariable` into a `createVariable` and a `mutateVariable`, I wanted a shared way of applying the same parent traversing logic the `putVariable` was applying. If it's to hard to understand and/or maintain I can change it.
cc: @coltmcnealy-lh

server/src/main/java/io/littlehorse/common/model/wfrun/ThreadRun.java
```diff
@@ -905,6 +908,49 @@ public void putVariable(String varName, VariableValue var) throws LHVarSubError
...
+     * who has the variable on its definition
+     * @param varName name of the variable
+     * @param var value of the variable
+     * @throws LHVarSubError when the varName is not found either on the current ThreadRun definition
+     * or its parents definition
+     */
+    public void mutateVariable(String varName, VariableValue var)
+        throws LHVarSubError {
+        VariableModification mutateVariable = (wfRunId, threadRunNumber, wfRun) -> {
+            Variable variable = wfRun
```
MateoRojas, line 945: Is it safe to assume that given we want to mutate the variable it will exists either on the `uncommittedChanges` or the `localStore`? I can make sure to throw an exception if the variable returned by `getDao().getVariable(...)` is null.
cc: @coltmcnealy-lh

sauljabin (approved):

server/src/main/java/io/littlehorse/common/model/wfrun/ThreadRun.java
```diff
@@ -885,17 +883,22 @@ public NodeRun getNodeRun(int position) {
...
+    /**
+     * Traverse the current and parent ThreadRun to run the variable modification on the appropriate ThreadRun
+     * @param varName name of the variable
+     * @param function function that executes the variable modification (e.g. creation, mutation, etc.)
+     * @throws LHVarSubError when the varName is not found either on the current ThreadRun definition
+     * or its parents definition
+     */
+    private void applyOnAppropriateThread(
+        String varName,
+        VariableModification function
```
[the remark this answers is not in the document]
MateoRojas: Great suggestion! Thanks :)

server/src/main/java/io/littlehorse/common/model/wfrun/ThreadRun.java
MateoRojas: Thanks :)

PR #43, opened 2023-08-16
Separate metadata processor

MateoRojas:
Separates metadata into its own processor. The separation is meant to be done incrementally, right now it was implemented with the fewest code changes possible. In a future commit we will be refactoring the code so that Metadata and Run processors do not share the same DAO and CommandProcessor classes.

**Note:** Search is not working. Everything else is, at least everything that is covered by e2e tests.

### DIFF
--- a/server/src/main/java/io/littlehorse/common/LHConfig.java
+++ b/server/src/main/java/io/littlehorse/common/LHConfig.java
@@ -141,6 +141,14 @@ public class LHConfig extends ConfigBase {
         return clusterId + "-core-cmd";
     }
 
+    public String getPepeCmdTopicName() {
+        return getPepeCmdTopicName(getLHClusterId());
+    }
+
+    public static String getPepeCmdTopicName(String clusterId) {
+        return clusterId + "-pepe-cmd";
+    }
+
     public String getRepartitionTopicName() {
         return getRepartitionTopicName(getLHClusterId());
     }
@@ -273,6 +281,17 @@ public class LHConfig extends ConfigBase {
                 replicationFactor
             )
                 .configs(globalMetaCLConfig)
+        );
+
+        // TODO: define the appropriate configuration for this topic -> For sure it shouldn't have compaction
+        out.add(
+            new NewTopic(
+                getPepeCmdTopicName(clusterId),
+                // This topic is input to a global store. Therefore, it doesn't
+                // make sense to have any more than just one partition.
+                1,
+                replicationFactor
+            )
         );
 
         return out;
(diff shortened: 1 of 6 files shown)


server/src/main/java/io/littlehorse/common/LHConfig.java
```diff
@@ -275,6 +283,17 @@ public static List<NewTopic> getAllTopics(
                 .configs(globalMetaCLConfig)
         );
 
+        // TODO: define the appropriate configuration for this topic -> For sure it shouldn't have compaction
```
MateoRojas, line 286: Is this the correct configuration for the metadata topic? 
@coltmcnealy-lh

coltmcnealy-lh: This is correct for the Metadata Command Topic.

server/src/main/java/io/littlehorse/server/streamsimpl/ServerTopology.java
```diff
@@ -67,45 +69,88 @@ public class ServerTopology {
...
+        topo.addSource(
+            METADATA_SOURCE, // source name
+            Serdes.String().deserializer(), // key deserializer
+            new LHDeserializer<>(Command.class, config), // value deserializer
+            config.getMetadataCmdTopicName() // source topic
+        );
+
+        topo.addProcessor(
+            METADATA_PROCESSOR,
+            () ->
```
coltmcnealy-lh, line 123: I am regretting the decision to use `prettier`

server/src/main/java/io/littlehorse/server/streamsimpl/coreprocessors/CommandProcessor.java
```diff
@@ -28,22 +29,42 @@ public class CommandProcessor
...
+        );
+        dao =
+            new KafkaStreamsLHDAOImpl(
+                this.ctx,
+                config,
+                server,
+                wfSpecCache,
+                localStore,
+                isMetadataProcessorInstance
+            );
```
coltmcnealy-lh, line 67: This is fine for now; however, in a follow-up PR let's add:
* `GlobalMetadataDAO`
* `ROGlobalMetaStore`

That should simplify the `KafkaStreamsLHDAOImpl`

coltmcnealy-lh (approved):
Good work. Thanks!

server/src/main/java/io/littlehorse/server/streamsimpl/ServerTopology.java
MateoRojas: Hahahaha I was thinking the same. I think we need to configure it better :)

PR #44, opened 2023-08-16
Creates metadata changelog topic and refactors getAllTopics

MateoRojas:
I forgot to add the creation of the metadata changelog topic previously so I added it now. Plus, I refactored the getAllTopics() method to explicitly show the topics that are being created and their configurations. It's more lines of codes than before, but I think that they are easier to read and understand what topics are being created. Let me know if you like the refactor, otherwise I can take it back to how it was.

### DIFF
--- a/server/src/main/java/io/littlehorse/common/LHConfig.java
+++ b/server/src/main/java/io/littlehorse/common/LHConfig.java
@@ -310,6 +310,7 @@ public class LHConfig extends ConfigBase {
 
         return List.of(
             coreCommand,
+            metadataCommand,
             repartition,
             observability,
             timer,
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/littlehorse/common/LHConfig.java
```diff
@@ -218,82 +222,104 @@ public List<NewTopic> getAllTopics() {
         );
     }
 
+    // Internal topics are manually created because:
+    // 1.- It makes it possible to manage/create topics using an external tool (e.g terraform, strimzi, etc.)
+    // 2.- It allows to explicitly manage the configuration of the topics
+    // Note: Kafka streams doesn't support disabling automatic internal topic creation. Thus, internal topics
+    // that are not explicitly created here will be automatically created by Kafka Stream. Please make sure to
+    // manually create all internal topics. Kafka has opened KIP-698 to solve this.
```
coltmcnealy-lh, line 230: Way to go, namedropping KIP's!

coltmcnealy-lh (approved):
This is much better than the original. Thank you!

PR #45, opened 2023-08-17
MINOR: allows hyphens in listener name

coltmcnealy-lh:
allows hyphens in listener name on Protocol Map, authorization map, and advertised listeners map.


sauljabin (approved):

sauljabin:
There are some unit test about it, I always like to test regexs, they are tircky

PR #46, opened 2023-08-18
Enhancement/pb refactor

eduwercamacaro:


coltmcnealy-lh (approved):

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Upgrade to Prettier 2.0
+7ae08ef1b92d3c14f9b037a01f4351e38392c8b4
```
sauljabin, line 2: I guess this is not going to work when squashing to master

sauljabin: I guess we can fix it in another commit

sauljabin (approved):

PR #57, opened 2023-08-25
Clean and change server authentication configurations

MateoRojas:
Summary of changes:

- `LHS_LISTENER_<LISTENER NAME>_CA_CERT` changed to `LHS_CA_CERT` and it's now global not per listener
- `LHS_LISTENERS_AUTHORIZATION_MAP` changed to `LHS_LISTENERS_AUTHENTICATION_MAP`
- `LHS_LISTENER_<LISTENER NAME>_AUTHORIZATION_SERVER` changed to `LHS_OAUTH_SERVER_URL` and it's now global not per listener
- `LHS_LISTENER_<LISTENER NAME>_CLIENT_ID` changed to `LHS_OAUTH_CLIENT_ID` and it's now global not per listener
- `LHS_LISTENER_<LISTENER NAME>_CLIENT_SECRET` changed to `LHS_OAUTH_CLIENT_SECRET` and it's now global not per listener


server/src/main/java/io/littlehorse/common/LHConfig.java
```diff
@@ -451,12 +476,18 @@ public List<ServerListenerConfig> getListeners() {
                     AuthorizationProtocol authProtocol =
                             authMap.get(name) == null ? AuthorizationProtocol.NONE : authMap.get(name);
 
+                    if (authProtocol == AuthorizationProtocol.MTLS && protocol != ListenerProtocol.MTLS) {
```
MateoRojas, line 479: This validates that we only allow MTLS as an authentication mechanism when the listener is set to user MTLS as communication protocol.

server/src/test/java/io/littlehorse/unit/common/LHConfigTest.java
MateoRojas, line 1: This file has a bunch of changes because I grouped the existing unit tests using inner classes so that it's easier to know what test if testing what method/functionality.

docs/SERVER_CONFIGURATIONS.md
```diff
@@ -355,17 +360,17 @@ LHS_LISTENER_MY_LISTENER_CLIENT_ID_FILE=/temp/client_id
 
 ---
 
-### `LHS_LISTENER_<LISTENER NAME>_CLIENT_SECRET`
+### `LHS_OAUTH_CLIENT_SECRET`
```
MateoRojas, line 363: On one of our slack chats it was suggested that this should be named `LHS_OAUTH_SERVER_SECRET`. I didn't name it that way because `CLIENT_SECRET` and `CLIENT_ID` are well known oauth industry names, so I thought it would be better to use names that people already know. I can change it to `LHS_OAUTH_SERVER_SECRET` if you all don't like `LHS_OAUTH_CLIENT_SECRET`.

docs/SERVER_CONFIGURATIONS.md
```diff
@@ -278,33 +279,37 @@ LHS_LISTENER_MY_LISTENER_CA_CERT=/tmp/ca
 
 ---
 
-### `LHS_LISTENERS_AUTHORIZATION_MAP`
+### `LHS_LISTENERS_AUTHENTICATION_MAP`
 
-Map between listener names and authorization protocols. Depends on `LHS_LISTENERS`.
+Map between listener names and authentication mechanisms. Depends on `LHS_LISTENERS`.
+
+> `LHS_LISTENERS_PROTOCOL_MAP` should be set to `MTLS` for a listener in order to use `MTLS` as authentication mechanisms.
```
coltmcnealy-lh, line 286: Nit pick—should be:
> as the authentication mechanism.

docs/SERVER_CONFIGURATIONS.md
```diff
@@ -278,33 +279,37 @@ LHS_LISTENER_MY_LISTENER_CA_CERT=/tmp/ca
...
 - **Type:** map
 - **Default:** null
 - **Importance:** high
 
 ---
 
-### `LHS_LISTENER_<LISTENER NAME>_AUTHORIZATION_SERVER`
+### `LHS_OAUTH_SERVER_URL`
+
+Optional OAuth server URL. Used by the server to identify the OpenId Connect endpoints and validate incoming calls.
```
coltmcnealy-lh, line 302: Should we say "authenticate" instead of "validate"?

Additionally, this is probably a question for Saúl, but are you sure OIDC is a thing here? I thought OIDC involved human flows and was something built on top of OAuth to communicate information about what groups a user belongs to.

docs/SERVER_CONFIGURATIONS.md
```diff
@@ -334,18 +339,18 @@ LHS_LISTENER_MY_LISTENER_CLIENT_ID=server
...
 ---
 
-### `LHS_LISTENER_<LISTENER NAME>_CLIENT_ID_FILE`
+### `LHS_OAUTH_CLIENT_ID_FILE`
 
-Optional OAuth2 Client Id. Used by the Worker to identify itself at an Authorization Server. Client Credentials Flow.
-Depends on `LHS_LISTENERS`. If it is different to null it overrides the `LHS_LISTENER_<LISTENER NAME>_CLIENT_ID` config
-and load the client id from the file.
+Optional OAuth2 Client Id. Used by the server to identify incoming calls against an Authorization Server.
+This OAuth client id is used by all the listeners. If it is different to null it overrides the `LHS_OAUTH_CLIENT_ID` config
```
coltmcnealy-lh, line 345: >This OAuth client id is used by all `MTLS` listeners, and authenticates all incoming calls against the OAuth server.

coltmcnealy-lh (approved):
Genial!

sauljabin (approved):

docs/SERVER_CONFIGURATIONS.md
MateoRojas: Thanks :)

PR #63, opened 2023-08-28
Update oauth configurations

sauljabin:
This PR includes:
- Remove openid scope from workers
- Improve the performance in the OAuthClient
- Change config from **LHS_OAUTH_SERVER_URL** to **LHS_OAUTH_INTROSPECT_URL** in the server
- Change config from **LHC_OAUTH_AUTHORIZATION_SERVER** to **LHS_OAUTH_ACCESS_TOKEN_URL** in the workers
- Add **LHC_OAUTH_SERVER_URL** for **lhctl** 
- Remove tokens from debug log


sdk-java/src/main/java/io/littlehorse/sdk/common/auth/OAuthCredentialsProvider.java
```diff
@@ -23,10 +23,10 @@ public void applyRequestMetadata(RequestInfo requestInfo, Executor executor, Met
         executor.execute(() -> {
             try {
                 if (currentToken == null || currentToken.isExpired()) {
-                    log.debug("Token expired, requesting a new one");
                     currentToken = oauthClient.getAccessToken();
+                    log.debug("Token was expired. New token: {}", currentToken);
```
MateoRojas, line 27: @sauljabin is it safe to log the token? I mean, the token is temporary and the access can be revoked but I'm not sure if it would be consider a security breach.

coltmcnealy-lh: Definitely don't `debug` log it. At most a `trace`. And yes, I would rather not log it at all.

sauljabin: good catch
1 All this logs should be removed before we finish (not only the token itself), I think they do not provide to much value
2 I was running into a bug, so we need unit test on this, in python i'm trying to cover as much as i can
3 This is the toString:
```
public String toString() {
        return String.format("IntrospectResponse [expired=%s, expiration date=%s]", isExpired(), expiration);
        return String.format(
                "TokenStatus [client id=%s, expired=%s, expiration date=%s]", clientId, isExpired(), expiration);
    }
```
We are not printing the token itself, just the status, i just needed to see that info not the token

server/src/main/java/io/littlehorse/server/auth/OAuthServerAuthorizer.java
```diff
@@ -69,11 +67,11 @@ private void validateToken(String token) throws Exception {
         TokenStatus tokenStatus = tokenCache.getIfPresent(token);
 
         if (tokenStatus == null) {
-            log.debug("Introspecting a new token");
             tokenStatus = client.introspect(token);
             tokenCache.put(token, tokenStatus);
+            log.debug("New token received: {}", tokenStatus);
         } else {
-            log.debug("Using cached token");
+            log.debug("Using cached token: {}", tokenStatus);
```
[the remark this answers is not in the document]
sauljabin: Let's remove them

coltmcnealy-lh (approved):
:ship:

eduwercamacaro (approved):
Suggestion. add a Non null checks to expiration field in TokenStatus class and/or add a test case when builder receives null for expiration field.

sauljabin:
@eduwercamacaro done, thanks

PR #68, opened 2023-08-30
Reduce the number of needed scripts

sauljabin:
The goal of this is to reduce the number of file and reuse variables and functions

* Merge do-server.sh with do-docker-server.sh
* Merge compile-proto.sh with Dockerfile
* Merge setup.sh with docker-compose.yml
* Merge setup.sh with setup-keycloak.sh
* Merge setup.sh with cleanup.sh


docker/standalone/Dockerfile
```diff
@@ -8,6 +8,7 @@ FROM amazoncorretto:17
 LABEL maintainer="engineering@littlehorse.io"
 
 ENV PATH ${PATH}:/kafka/bin:/lh
+VOLUME /data
```
coltmcnealy-lh, line 11: What happens if you want the data to be stored locally inside the container, not on a mounted volume? Does this still work?

local-dev/compile-proto.sh
```diff
@@ -4,11 +4,31 @@ set -e
...
-PROTOC_DOCKER_DIR=$(cd "$SCRIPT_DIR/../docker/protoc" && pwd)
+PUBLIC_PROTOS=$(ls "$WORK_DIR"/schemas | grep -v -E "^internal")
+INTERNAL_PROTOS=$(ls "$WORK_DIR"/schemas/internal)
+docker_run="docker run --rm -it -v ${WORK_DIR}:/littlehorse protoc:23.4"
 
 # compile protoc
-echo "Compiling protoc" $(docker build -q --file ${PROTOC_DOCKER_DIR}/Dockerfile --tag protoc ${PROTOC_DOCKER_DIR})
-echo "Protoc version" $(docker run --rm -it protoc protoc --version)
+echo "Compiling docker image 'protoc:23.4'"
+docker build -q --tag protoc:23.4 -<<EOF
```
coltmcnealy-lh, line 13: can we tag it `lh-protoc` so that when people do `docker images` they can see it's littlehorse?

eduwercamacaro (approved):

coltmcnealy-lh (approved):
Looks great!

local-dev/compile-proto.sh
sauljabin: sure

docker/standalone/Dockerfile
sauljabin: if the stop de container (manually or because you restarted your compu) and then you run docker start the data will be there, they can run docker rm <dockername litthlehorse in the example> and then run again it will create a new volume

sauljabin: testing without this line again

sauljabin: you are right, not needed, much better without it, removing it

PR #70, opened 2023-08-30
Python Workflow SDK: Phase I

sauljabin:
This PR includes:
* Workflow
* Utils
* ThreadBuilder Execute, Add variable, Format
* WfRunVariable: JSON path, Index
* Examples: Basic, Json


sdk-python/littlehorse/exceptions.py
```diff
@@ -14,3 +17,11 @@ class SerdeException(Exception):
     def __init__(self, message: str):
         self.message = message
         super().__init__(self.message)
+
+
+class GrpcException(Exception):
```
coltmcnealy-lh, line 22: Is there a reason why we don't use the standard grpc exception like we do in the Java SDK?

We wrap it in the server but we don't expose the wrapper to the users in `sdk-java`. The wrapper is just so that it's easier to throw the exception.

sdk-python/littlehorse/workflow.py
```diff
@@ -0,0 +1,471 @@
...
+            raise ValueError(f"Invalid JsonPath: {json_path}. Use $. at the beginning")
+
+        if self._json_path is not None:
+            raise ValueError("Cannot set a json_path twice on same var")
+
+        self._json_path = json_path
+
+    def with_json_path(self, json_path: str) -> "WfRunVariable":
+        """Valid only for output of the JSON_OBJ or JSON_ARR types. Returns a new
+        NodeOutput handle which points to Json element referred to by the json path.
```
coltmcnealy-lh, line 155: `s/NodeOutput/WfRunVariable`

coltmcnealy-lh (approved):
This is fantastic. So exciting!

sdk-python/littlehorse/exceptions.py
sauljabin: this is the reason `from grpc._channel import _InactiveRpcError`

`_channel`? `_InactiveRpcError`? i do not like the python grpc lib, do you want me to remove the GrpcException? if so, ther is no problem

PR #77, opened 2023-08-31
Feature/wait for event exception handler

eduwercamacaro:
This PR add support for propagating user defined exception to parent threads. in that way users can catch the exception an implement its own business use cases after a exception was thrown.  This makes this possible:

```
NodeOutput nodeOutput = thread.waitForThreads(p1Thread, p2Thread, p3Thread);

thread.handleException(nodeOutput, "denied-by-user", xnHandler -> {
    xnHandler.execute("exc-handler");
});
```

Also it adds new E2E cases


sauljabin (approved):

coltmcnealy-lh (approved):
Well done!

PR #79, opened 2023-09-02
--quick option for local docker build

coltmcnealy-lh:
Adds a --quick option to local-dev/build.sh which uses local gradle cache to build.

### DIFF
--- a/local-dev/build.sh
+++ b/local-dev/build.sh
@@ -9,8 +9,6 @@ CONTEXT_DIR=$(cd "$SCRIPT_DIR/.." && pwd)
 
 cd "$CONTEXT_DIR"
 
-CONTAINER=lh-quickbuilder
-
 if [ "$1" == "--quick" ]; then
     # The quick build compiles the jar using gradle on the host machine, which
     # enables usage of the gradle cache. This is much faster than building from
--- a/local-dev/build.sh
+++ b/local-dev/build.sh
@@ -9,10 +9,28 @@ CONTEXT_DIR=$(cd "$SCRIPT_DIR/.." && pwd)
 
 cd "$CONTEXT_DIR"
 
-echo "Building server"
-docker build --file ${LH_SERVER_WORK_DIR}/Dockerfile \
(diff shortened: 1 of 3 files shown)


local-dev/build.sh
```diff
@@ -9,10 +9,28 @@ CONTEXT_DIR=$(cd "$SCRIPT_DIR/.." && pwd)
 
 cd "$CONTEXT_DIR"
 
-echo "Building server"
-docker build --file ${LH_SERVER_WORK_DIR}/Dockerfile \
-    --tag littlehorse/littlehorse-server:latest ${CONTEXT_DIR}
+CONTAINER=lh-quickbuilder
```
sauljabin: What is this for?

local-dev/util/Dockerfile.server-quick
```diff
@@ -0,0 +1,9 @@
+FROM amazoncorretto:17
```
sauljabin, line 1: we can encapsulate this inside the same script:

example:

 docker build -f ${SCRIPT_DIR}/util/Dockerfile.server-quick \
        --tag littlehorse/littlehorse-server:lates -<<EOF
WORKDIR /lh

COPY ./docker/server/docker-entrypoint.sh /lh
COPY ./docker/server/log4j2.properties /lh
COPY ./server/build/libs/server-*-all.jar /lh/server.jar
ENTRYPOINT ["/lh/docker-entrypoint.sh"]
CMD ["server"]
EOF

sauljabin: it's just a suggestion hehe

sauljabin (approved):

local-dev/util/Dockerfile.server-quick
coltmcnealy-lh: Roger Roger. Making that change then merging.

coltmcnealy-lh: I got an error after playing around with it for a few minutes, and I don't see the option here: https://docs.docker.com/engine/reference/commandline/build/

I'm going to merge as-is, and we can improve it later.

PR #83, opened 2023-09-06
Improvements to monitoring on LH Server.

coltmcnealy-lh:
* Remove the Readiness probe, since it turns out we will not use it on K8s.
* Adds a /diskUsage endpoint to HealthService.
* Removes the manual Streams Restoration metrics from prometheus since they're added in KIP-869. TODO: add appropriate ServerFilterRules.
* Removes the TopologyInstanceState rpc endpoint, since LH Server Health data does NOT belong in the public grpc api.
* Adds a /status endpoint to the HealthService which does roughly the same as the TopologyInstanceState endpoint.

NOTE: I discovered two interesting things about Streams when doing this:
1. In the Timer Topology (which is at-least-once semantics), we can observe 'negative lag'. I will ask Matthias and the rest of the folks what that means.

2. When a task is fully down (i.e. restoring), it doesn't show up in either activeTasks() or standbyTasks(). However, we do see the restoration going on.

As per item 2, this commit has a bit more work to do. We need to show it as 'unhealthy task' or add it to the activeTasks.


server/src/main/java/io/littlehorse/server/monitoring/HealthService.java
```diff
@@ -0,0 +1,125 @@
+package io.littlehorse.server.monitoring;
```
coltmcnealy-lh, line 1: Review this package, please :smile:

examples/hundred-tasks/src/main/java/io/littlehorse/examples/MyWorker.java
```diff
@@ -11,7 +11,11 @@ public String task1() {
 
     @LHTaskMethod("task-2")
     public String task2() {
-        return "task-2";
+        StringBuilder out = new StringBuilder();
+        for (int i = 0; i < 100; i++) {
+            out.append("asdfasdfasdfasdfjwth0238ahs9213t8hd");
```
sauljabin, line 16: is this going to be removed?

server/src/main/java/io/littlehorse/server/listener/ServerListener.java
```diff
@@ -43,6 +43,8 @@ public void start() {
 
     @Override
     public void close() {
+        // This forcibly closes all grpc connections rather than waiting for them to
+        // complete. This cuts off all streaming connections too.
         server.shutdownNow();
```
sauljabin, line 48: were you to change this right? to server.shutdown();

server/src/main/java/io/littlehorse/server/monitoring/metrics/ServerFilterRules.java
```diff
@@ -9,6 +9,7 @@ private ServerFilterRules() {}
 
     // relevance ordered
     public static final List<ServerFilterRule> RULES = List.of(
+            // TODO: Wait for KIP-869 and gather state restoration metrics.
```
sauljabin, line 12: here we should remove all the custom metric we created before and are removed now

sauljabin (approved):

server/src/main/java/io/littlehorse/server/listener/ServerListener.java
coltmcnealy-lh: `server.shutDown()` made the server hang because we would need to individually close each client request (especially the bidirectional stream ones with the task workers) before the server is allowed to shut down.

Also `server.isShutDown()` is basically just a flag that returns whether `server.shutDown()` was called...not really a health metric ):

examples/hundred-tasks/src/main/java/io/littlehorse/examples/MyWorker.java
coltmcnealy-lh: Good catch, will remove. I added it so that we could build up state faster so that restoration times would take longer, so that I could actually test everything.

But I don't think it needs to be in the examples :+1:

server/src/main/java/io/littlehorse/server/monitoring/metrics/ServerFilterRules.java
coltmcnealy-lh: Roger Roger

PR #84, opened 2023-09-06
Python SDK Mutation and Ext Event

sauljabin:
* Add wait_for_event
* Add mutate
* Add NodeOutput
* Add utils
* Fix has_context function


sdk-python/littlehorse/user_utils.py
```diff
@@ -100,3 +100,24 @@ def register_task(
             logging.info(f"TaskDef {name} already exits, skipping")
             return
         raise e
+
+
+def register_external_event(
```
coltmcnealy-lh, line 105: This is nice! Java and Go don't have it yet :wink:

coltmcnealy-lh: Actually, could you please rename it to `register_external_event_def`? The `def` is important, `ExternalEvent != ExternalEventDef`.

coltmcnealy-lh (approved):
:pray:

sdk-python/littlehorse/user_utils.py
sauljabin: is it possible "register" a event? isn't like post_external_event? 

register_external_event is for PutExternalEventDef
post_external_event is PutExternalEvent

or is it confuse? 🤔

sauljabin: in lhctl we have postEvent

coltmcnealy-lh: This method creates external event definition, so it should have _def_ at the end.

You "postEvent` an External Event, which is dfferent than `externalEvent`.

It's `lhctl deploy exteranlEventDef` and `lhctl postEvent`... probably shouldbe `lhctl put externalEvent` but oops

sauljabin: what is instead of register we use "create"

PR #89, opened 2023-09-07
Fix/reassign user task on deadline

eduwercamacaro:


sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/ThreadBuilderImpl.java
```diff
@@ -112,10 +112,12 @@ public void reassignToGroupOnDeadline(UserTaskOutput userTaskOutput, int deadlin
...
         }
-        VariableAssignment userGroup = curNode.getUserTaskBuilder().getUserGroup();
-        if (userGroup == null) {
-            throw new IllegalStateException("User task is not assigned to a userGroup");
+        UserTaskNode.UserAssignment userAssignment =
+                curNode.getUserTaskBuilder().getUser();
+        if (userAssignment == null) {
+            throw new IllegalStateException("User task is not assigned to a user");
         }
+        VariableAssignment userGroup = userAssignment.getUserGroup();
```
coltmcnealy-lh, line 120: Do we need to validate that the userGroup isn't null?

server/src/main/java/io/littlehorse/common/model/getable/core/usertaskrun/UserTaskRunModel.java
```diff
@@ -337,6 +339,14 @@ private void scheduleTaskReassign(UTActionTriggerModel action) {
         LocalDateTime localDateTime = LocalDateTime.now().plusSeconds(delayInSeconds);
         Date maturationTime =
                 Date.from(localDateTime.atZone(ZoneId.systemDefault()).toInstant());
+        ReassignUserTask command = buildReassignUserTaskCommandFrom(action);
+        if (command != null) {
```
coltmcnealy-lh, line 343: Nit pick: It's not clear when `command` would be null. Perhaps we could put the test for whether it would be null in this function, and then guarantee that `buildReassignUserTaskCommandFrom()` always returns a value?

coltmcnealy-lh (approved):
Good work

PR #93, opened 2023-09-11
Changes from gradle to gradlew everywhere

MateoRojas:
It's a good practice to use gradle wrapper because:
- The commands can still run even if the computer doesn't have gradle installed
- Ensures that the same version of gradle is used everywhere


coltmcnealy-lh (approved):

sauljabin (changes requested):
rejected, pipeline failed

sauljabin:
@MateoRojas any update on this?

coltmcnealy-lh:
I updated Mateo's PR since he's out. The build was failing because `./gradlew` isn't working inside the Builder Image. However, I think using `gradle` inside the image is fine, especially because we explicitly build from the `gradle:8` docker image, which means we are explicitly choosing the specific gradle stack to use. I reverted the changes on the Standalone and Server Dockerfile. The rest of the PR is untouched; now the pipeline is green.

sauljabin (approved):

PR #95, opened 2023-09-11
Allows Variables to be persistent across WfSpec versions.

coltmcnealy-lh:
* Adds a 'bool persistent' field to VariableDef.
* If a variableDef is persistent, the VariableModel gets an index with just wfSpecName, variable name, and value (no wfSpecVersion).
* Allows searching for persistent variables without wfSpecVersion.
* Validates that a new version of a WfSpec respects all persistent variables from the previous version.
* Adds a WfRunVariable#persistent() method to SDK to allow creating persistent variables.


sauljabin (approved):

sauljabin:
I like from now, every change has to be implemented in all SDKs, that way we prevent having drifts

PR #96, opened 2023-09-11
Python Workflow SDK Conditionals

sauljabin:
* Add do_if
* Add do_if_else
* Add example


sdk-python/littlehorse/workflow.py
```diff
@@ -287,25 +403,49 @@ def __init__(self, workflow: "Workflow", initializer: "ThreadInitializer") -> No
         """
         self.wf_run_variables: list[WfRunVariable] = []
         self._workflow = workflow
-        self._nodes: dict[str, Node] = {}
+        self._nodes: list[WorkflowNode] = []
```
coltmcnealy-lh, line 406: Interesting, this is quite different from the other two SDK's. Probably it's cleaner this way.

sdk-python/littlehorse/workflow.py
```diff
@@ -459,32 +612,74 @@ def add_node(self, name: str, sub_node: NodeType) -> str:
...
+            (which the Library casts to a Variable Value) or a
+            `WfRunVariable` representing the RHS of the expression.
+
+        Returns:
+            WorkflowCondition: a WorkflowCondition.
+        """
+        return WorkflowCondition(left_hand, comparator, right_hand)
+
+    def do_if(
+        self, condition: WorkflowCondition, initializer: "ThreadInitializer"
```
coltmcnealy-lh, line 644: `ThreadInitializer` is much better than `ThreadBuilder`. Perhaps it's worth changing in the other SDK's?

In Go, changing the names won't be hard at all since it uses duck-typing.

In Java, we could have `public interface ThreadInitializer`, and `public interface ThreadBuilder extends ThreadInitializer` but `ThreadBuilder` doesn't have anything different. That way maybe the current API contract works, and we can mark `ThreadBuilder` as `@Deprecated`

sdk-python/littlehorse/workflow.py
```diff
@@ -314,6 +454,24 @@ def _check_if_active(self) -> None:
...
+            raise ReferenceError("No node found")
+        return self._nodes[-1]
+
+    def _find_node(self, name: str) -> WorkflowNode:
+        for node in self._nodes:
+            if node.name == name:
+                return node
+        raise ReferenceError("Node not found")
+
+    def _find_next_node(self, name: str) -> WorkflowNode:
```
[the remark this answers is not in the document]
coltmcnealy-lh: It works!

coltmcnealy-lh (approved):

PR #100, opened 2023-09-12
Lh test refactors

eduwercamacaro:
This PR adds support for lh-test:

1. Inject LhClient to test classes
2. Show which steps failed in test cases
3. Shows expected value and evaluated value on failed tests


lh-test/src/main/java/io/littlehorse/test/WfRunVerifier.java
```diff
@@ -114,36 +114,47 @@ public WfRunVerifier waitForUserTaskRunStatus(int threadRunNumber, int nodeRunNu
...
-        steps.add(new WaitForStatusStep<>(objectLHStatusFunction, taskStatus));
+        steps.add(new WaitForStatusStep<>(objectLHStatusFunction, taskStatus, steps.size() + 1));
         return this;
     }
+
+    private WfRunId wfRunIdFrom(String wfRunId) {
+        return WfRunId.newBuilder().setId(wfRunId).build();
+    }
+
+    private NodeRunId nodeRunIdFrom(String wfRunId, int threadRunNumber, int nodeRunNumber) {
```
coltmcnealy-lh, line 153: I've added a ticket to the backlog to create a class `LHProtoUtil` or something like that with a bunch of methods like this. We should make life easier for customers :smile:

lh-test/src/main/java/io/littlehorse/test/internal/TestContext.java
```diff
@@ -125,4 +140,20 @@ public void registerUserTaskSchemas(Object testInstance) throws IllegalAccessExc
             registerUserTaskDef(taskDefRequest);
         }
     }
+
+    public WfSpec registerWfSpecIfNotPresent(Workflow workflow) {
```
coltmcnealy-lh, line 144: FANTASTIC

coltmcnealy-lh (approved):

PR #109, opened 2023-09-13
Add interruptions and sleeps to python sdk

sauljabin:


sdk-python/examples/interruption/example_interruption.py
```diff
@@ -0,0 +1,59 @@
...
+    return config
+
+
+def get_workflow() -> Workflow:
+    def my_interrupt_handler(thread: ThreadBuilder) -> None:
+        thread.execute("interrupt-handler")
+
+    def my_entrypoint(thread: ThreadBuilder) -> None:
+        thread.add_interrupt_handler("interruption-event", my_interrupt_handler)
+        thread.sleep(30)
```
coltmcnealy-lh, line 27: 30 seconds is good. We should describe what this Workflow does in the README. Say "the parent thread sleeps 30 sec, we can interrupt it while it's sleeping, and it creates child thread, etc"

sdk-python/examples/interruption/example_interruption.py
```diff
@@ -0,0 +1,59 @@
...
+    config = LHConfig()
+    config_path = Path.home().joinpath(".config", "littlehorse.config")
+    if config_path.exists():
+        config.load(config_path)
+    return config
+
+
+def get_workflow() -> Workflow:
+    def my_interrupt_handler(thread: ThreadBuilder) -> None:
+        thread.execute("interrupt-handler")
```
coltmcnealy-lh, line 23: For clarity purposes: let's call it `some-task` not `interrupt-handler`.

sdk-python/examples/interruption/README.md
```diff
@@ -0,0 +1,32 @@
...
+lhctl get wfRun <wf run id>
+
+# Note that it is 'RUNNING'. Next, post an external event using the following:
+lhctl postEvent <wf run id> interruption-event NULL
+
+# Then inspect the wfRun:
+# Note it is 'COMPLETED' whit type 'INTERRUPT'
+lhctl get wfRun <wf run id>
+
+# Then inspect the output of the ExternalEvent and how it completed the nodeRun:
```
coltmcnealy-lh, line 29: This sounds incorrect. In an Interrupt scenario, `ExternalEvent` doesn't complete a `NodeRun`. Rather, the `ExternalEvent` causes the creation of a new `ThreadRun` with type `ThreadType.INTERRUPT`.

Also, on line 27, it may be `COMPLETD` or `RUNNING` depending on how fast you do `lhctl get wfRun`.

In order for the Interrupt to happen, the parent thread needs to be sleeping.

sdk-python/tests/test_workflow.py
```diff
@@ -687,6 +695,59 @@ def my_entrypoint(self, thread: ThreadBuilder) -> None:
...
+    def test_compile_with_interrupt(self):
+        def my_interrupt_handler(thread: ThreadBuilder) -> None:
+            thread.execute("interrupt-handler")
+
+        def my_entrypoint(thread: ThreadBuilder) -> None:
+            thread.add_interrupt_handler("interruption-event", my_interrupt_handler)
+            thread.execute("my-task")
+
+        wf = Workflow("my-wf", my_entrypoint)
+        self.assertEqual(
```
coltmcnealy-lh, line 707: Whoa this is an intense test!

coltmcnealy-lh (approved):
Approved, but it would be great to do some small updates to the example README

eduwercamacaro (approved):

PR #113, opened 2023-09-15
Feature/lh 25 python user tasks

coltmcnealy-lh:
Working commit with scheduling of User Tasks working. Still needs:
* Automatic reassignment
* Schedule Reminder Tasks
* Add Notes to User Task

### DIFF
--- a/sdk-python/littlehorse/workflow.py
+++ b/sdk-python/littlehorse/workflow.py
@@ -494,6 +494,11 @@ class WorkflowInterruption:
         return to_json(self.compile())
 
 
+class UserTaskOutput(NodeOutput):
+    def __init__(self, node_name: str) -> None:
+        super().__init__(node_name)
+
+
 class ThreadBuilder:
     def __init__(self, workflow: "Workflow", initializer: "ThreadInitializer") -> None:
         """This is used to define the logic of a ThreadSpec in a ThreadInitializer.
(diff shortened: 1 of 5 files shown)


sdk-python/littlehorse/workflow.py
```diff
@@ -634,6 +639,26 @@ def execute(self, task_name: str, *args: Any) -> NodeOutput:
         node_name = self.add_node(task_name, task_node)
         return NodeOutput(node_name)
 
+    def assign_user_task(
```
sauljabin, line 642: Some methods are missing right? and an example

sauljabin (approved):
code looks good, also we have talked already about the optimization you did

sdk-python/littlehorse/workflow.py
coltmcnealy-lh: Yes and yes. Still in draft...

sdk-python/littlehorse/workflow.py
```diff
@@ -494,6 +503,48 @@ def __str__(self) -> str:
         return to_json(self.compile())
 
 
+class LHFormatString:
```
sauljabin, line 506: this class already existed i guess

sauljabin:
```
class FormatString:
    def __init__(self, format: str, *args: Any) -> None:
        """Generates a FormatString object that can be understood by the ThreadBuilder.

        Args:
            format (str): String format with variables with curly brackets {}.
            *args (Any): Arguments.

        Returns:
            FormatString: A FormatString.
        """
        self.format = format
        self.args = args
```

PR #115, opened 2023-09-15
Feature/wait for threads refactor

eduwercamacaro:


sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/SpawnedThread.java
```diff
@@ -4,4 +4,7 @@
  * This is the output of `ThreadBuilder::spawnThread()`. It is used as input for
  * `ThreadBuilder::waitForThread()`.
  */
-public interface SpawnedThread {}
+public interface SpawnedThread {
+
+    WfRunVariable getThreadNumberVariable();
```
coltmcnealy-lh, line 9: Hmm. I would prefer not exposing this to the public interface because then the customer might try to use that internal variable.

coltmcnealy-lh (approved):
I see why you added `getThreadNumberVariable` to the `SpawnedThreads` interface...I'd prefer to not have it public, but if it is necessary to be public, that is fine.

PR #122, opened 2023-09-25
Fully support for API search in lh-test

eduwercamacaro:


sauljabin (approved):
This is so pro, thanks.

I would add:
- Java doc to interfaces and mathods
- Change the RuntimeException

PR #123, opened 2023-09-27
Add lhctl image

sauljabin:


MateoRojas:
Quick question, the purpose of this image is being able to run lhctl commands without having to install it on the machine?

coltmcnealy-lh (approved):

coltmcnealy-lh:
Maybe all you need to do to install `lhctl` now is:

```
alias lhctl='docker run -it gallery.ecr.aws/littlehorse/lhctl:latest'
```

sauljabin:
correct, something interesting is we can run a pod with command: tail -f /dev/null

PR #125, opened 2023-09-28
Add fail method to builder

sauljabin:


coltmcnealy-lh (approved):
When the tests are longer than the change :joy:

eduwercamacaro (approved):

sdk-python/littlehorse/workflow.py
```diff
@@ -936,6 +937,36 @@ def handle_error(
         last_node = self._find_node(node.node_name)
         last_node.failure_handlers.append(failure_handler)
 
+    def fail(
+        self, failure_name: str, message: str, input: Optional[Any] = None
```
sauljabin, line 941: @coltmcnealy-lh is input right? or output

PR #132, opened 2023-10-10
Task Workers Health checks

eduwercamacaro:
This PR adds full support for liveness check and a small support for Readiness (Readiness is more complex that just verifying the streams state) for Task workers.
It covers python, java and go sdks


local-dev/configs/server-3.config
```diff
@@ -0,0 +1,26 @@
+LHS_KAFKA_BOOTSTRAP_SERVERS=localhost:9092
```
eduwercamacaro, line 1: The only way to test this new feature on local env is having 3 nodes deployed

schemas/service.proto
```diff
@@ -315,6 +315,7 @@ message TaskWorkerHeartBeatRequest {
 
 message RegisterTaskWorkerResponse {
   repeated LHHostInfo your_hosts = 1;
+  optional bool is_cluster_healthy = 2;
```
eduwercamacaro, line 318: It says if the cluster is able to process hearbeat for the task worker. Implementation of this will be more complex in the future.

coltmcnealy-lh:
I think we may need to have a zoom to discuss the `optional bool is_cluster_healthy`...other than that, this PR looks good.

schemas/service.proto
coltmcnealy-lh: 1. After talking to Edu offline, it seems that the `optional` is so that _new clients_ can talk to _old servers._ Makes sense

coltmcnealy-lh (approved):
Looks good, requires some significant documentation.

PR #136, opened 2023-10-23
Feature/enable sdk retries

coltmcnealy-lh:
I have returned.

### DIFF
--- a/sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
+++ b/sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
@@ -209,11 +209,11 @@ final class WorkflowThreadImpl implements WorkflowThread {
         return new LHFormatStringImpl(this, format, args);
     }
 
-    public NodeOutputImpl execute(String taskName, Object... args) {
+    public TaskNodeOutputImpl execute(String taskName, Object... args) {
         checkIfIsActive();
         TaskNode taskNode = createTaskNode(taskName, args);
         String nodeName = addNode(taskName, NodeCase.TASK, taskNode);
-        return new NodeOutputImpl(nodeName, this);
+        return new TaskNodeOutputImpl(nodeName, this);
     }
 
     private TaskNode createTaskNode(String taskName, Object... args) {
@@ -450,6 +450,20 @@ final class WorkflowThreadImpl implements WorkflowThread {
         mutate(internalStartedThreadVar, VariableMutationType.ASSIGN, new NodeOutputImpl(nodeName, this));
 
         return new SpawnedThreadImpl(this, threadName, internalStartedThreadVar);
+    }
+
+    public void overrideTaskRetries(TaskNodeOutputImpl node, int retries) {
+        checkIfIsActive();
+        Node.Builder nb = spec.getNodesOrThrow(node.nodeName).toBuilder();
+        if (nb.getNodeCase() != NodeCase.TASK) {
+            throw new IllegalStateException("Impossible to not have task node here");
+        }
+
+        TaskNode.Builder taskBuilder = nb.getTaskBuilder();
+        taskBuilder.setRetries(retries);
+
+        nb.setTask(taskBuilder);
+        spec.putNodes(node.nodeName, nb.build());
     }
 
     public void addTimeoutToExtEvt(NodeOutputImpl node, int timeoutSeconds) {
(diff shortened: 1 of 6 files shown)


sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
```diff
@@ -444,19 +452,42 @@ public SpawnedThreadImpl spawnThread(ThreadFunc threadFunc, String threadName, M
         return new SpawnedThreadImpl(this, threadName, internalStartedThreadVar);
     }
 
+    public void overrideTaskRetries(TaskNodeOutputImpl node, int retries) {
+        checkIfIsActive();
+        Node.Builder nb = spec.getNodesOrThrow(node.nodeName).toBuilder();
+        if (nb.getNodeCase() != NodeCase.TASK) {
+            throw new IllegalStateException("Impossible to not have task node here");
```
diablouma, line 459: hey @coltmcnealy-lh what do you think about having a kind of positive statement for this error message, something similar to : "only tasks nodes allowed", I think is easier from the user perspective. wdyt?

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
```diff
@@ -444,19 +452,42 @@ public SpawnedThreadImpl spawnThread(ThreadFunc threadFunc, String threadName, M
...
+            n.setTask(task);
 
-        n.setExternalEvent(evt);
+        } else if (n.getNodeCase() != NodeCase.EXTERNAL_EVENT) {
+
+            ExternalEventNode.Builder evt = n.getExternalEventBuilder();
+            evt.setTimeoutSeconds(timeoutValue);
+            n.setExternalEvent(evt);
+        } else {
+            throw new RuntimeException("Tried to set timeout on non-ext evt node!");
```
diablouma, line 489: what do you think about using the full name instead of the abbreviation for non-ext evt?

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
eduwercamacaro, line 459: Suggestion:

"Retries are allowed only for task nodes" or smth like that.

sdk-java/src/test/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImplTest.java
```diff
@@ -107,4 +107,65 @@ void testWhileLoopConditional() {
...
+        assertThat(defaultNode.getTask().getRetries()).isEqualTo(1);
+
+        Node overridenNode =
+                wfSpec.getThreadSpecsOrThrow(wfSpec.getEntrypointThreadName()).getNodesOrThrow("2-asdf-TASK");
+        assertThat(overridenNode.getTask().getRetries()).isEqualTo(2);
+    }
+
+    @Test
+    void setDefaultTaskTimeout() {
+        WorkflowImpl wf = new WorkflowImpl("asdf", thread -> {
```
diablouma, line 156: I think it is worth to have more real-life names for the variables we have on the test scenarios that makes it more clear the intention of the test and also helps to debug easier if needed

eduwercamacaro (approved):
Approved with a suggestion. Also, I think these defaults retries and timeouts could be easier to implement in the client-side if the server handle this logic.

diablouma (approved):

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
coltmcnealy-lh: It's not possible to construct an input that makes this not be a task node. Which means that if we reach here, then there is a bug in the SDK itself.

coltmcnealy-lh:
> Approved with a suggestion. Also, I think these defaults retries and timeouts could be easier to implement in the client-side if the server handle this logic.

The server already handles the logic, it just hadn't ever been added to the SDK's.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
coltmcnealy-lh: Good suggestion

PR #138, opened 2023-10-26
Two fixes to test harness:

coltmcnealy-lh:
- dont reuse same workflow name parallel-approval
- Ignore already exist errors on external event def's which makes ExternalBootstrapper work better


eduwercamacaro (approved):
Thanks.!

PR #141, opened 2023-10-31
Feature/multitenancy and acls

eduwercamacaro:
Major refactor for LittleHorse storage in order to support multi-tenancy features. This PR includes two new interfaces to interact with native rocksdb store. `ReadOnlyModelStore` API is able to get specific stored getables and execute randoms scans, in addition `ModelStore` extends `ReadOnlyModelStore` to provide write semantics to the store, ModelStores are designed to be used only for DAO classes. In the other hand, there are two implementation for these stores `DefaultModelStore` and `TenantModelStore` these implementations represent logical stores first one is a general store and the last one is a tenant isolated store. 

Also, this PR adds `@Authorize(...)` annotation, you can add this annotation to restrict any Grpc request from clients, when this annotation is not present in a Grpc endpoint, it takes admin acls as required.


lhctl/cmd/principal.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"context"
+	"strings"
+	"github.com/spf13/cobra"
+	"github.com/littlehorse-enterprises/littlehorse/sdk-go/common"
+	"github.com/littlehorse-enterprises/littlehorse/sdk-go/common/model"
+)
+
+
+var createPrincipalCmd = &cobra.Command{
```
coltmcnealy-lh, line 12: Nit pick: let's rename this to `putPrincipalCmd`

lhctl/cmd/principal.go
```diff
@@ -0,0 +1,85 @@
...
+				Resources: allowedResources,
+				AllowedActions: actions,
+			}
+			serverAcls = append(serverAcls, &serverAcl)
+		}
+		putRequest := model.PutPrincipalRequest{
+			Id: id,
+			Acls: serverAcls,
+		}
+		print("hmmm")
```
coltmcnealy-lh, line 31: :eyes:

lhctl/cmd/put.go
```diff
@@ -0,0 +1,15 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+// executeCmd represents the run command
+var putCmd = &cobra.Command{
+	Use:   "put",
+	Short: "Create an object",
```
coltmcnealy-lh, line 10: Nit: It should say "Create or update an object"

lhctl/cmd/tenant.go
```diff
@@ -0,0 +1,28 @@
...
+	"context"
+	"log"
+	"github.com/spf13/cobra"
+	"github.com/littlehorse-enterprises/littlehorse/sdk-go/common"
+	"github.com/littlehorse-enterprises/littlehorse/sdk-go/common/model"
+)
+
+var putTenant = &cobra.Command{
+	Use:   "tenant",
+	Short: "Deploy a tenant.",
```
coltmcnealy-lh, line 13: Maybe:
> Create a Tenant. Currently, updating Tenants is not supported.

server/src/main/java/io/littlehorse/common/ServerContext.java
```diff
@@ -0,0 +1,17 @@
+package io.littlehorse.common;
+
+import io.littlehorse.common.model.getable.global.acl.PrincipalModel;
+
+public interface ServerContext {
+
+    enum Scope {
```
coltmcnealy-lh, line 7: Can you please put a comment here so that someone who doesn't understand the difference between GRPC Server and Streams Command Processor can understand this? It makes 100% sense to me but probably not for @diablouma ... when he does his first LH Server ticket he'd probably get confused :joy:

server/src/main/java/io/littlehorse/common/ServerContextImpl.java
```diff
@@ -0,0 +1,30 @@
...
+        this.scope = scope;
+    }
+
+    @Override
+    public Scope scope() {
+        return scope;
+    }
+
+    @Override
+    public PrincipalModel principal() {
```
coltmcnealy-lh, line 22: Two questions:
1. Are we going to standardize on using `getFoo()` or `foo()`, or does it not matter? I personally don't care, I'm not a java guy.
2. Why `null`?

server/src/main/java/io/littlehorse/common/dao/CoreProcessorDAO.java
```diff
@@ -77,4 +77,6 @@ public Date getEventTime() {
     public abstract LHHostInfo getAdvertisedHost(HostModel host, String listenerName);
 
     public abstract InternalHosts getInternalHosts();
+
+    public abstract void onPartitionClaimed();
```
coltmcnealy-lh, line 81: Note: We can use this method to guarantee RYOW even in the case of a rebalance.

server/src/main/java/io/littlehorse/common/model/getable/core/taskrun/TaskRunModel.java
```diff
@@ -204,8 +204,9 @@ public void processStart(TaskClaimEvent se) {
         taskResult.setTaskRunId(id);
         taskResult.setTime(new Date(System.currentTimeMillis() + (1000 * timeoutSeconds)));
         taskResult.setStatus(TaskStatus.TASK_TIMEOUT);
-
-        LHTimer timer = new LHTimer(new CommandModel(taskResult, taskResult.getTime()), getDao());
+        CommandModel timerCommand = new CommandModel(taskResult, taskResult.getTime());
```
coltmcnealy-lh, line 207: Please make the `tenantId` a required argument for the `CommandModel` constructor so that:

1. It is impossible to have a null tenant id.
1. This code can be one line again

server/src/main/java/io/littlehorse/common/model/getable/global/acl/PrincipalModel.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public PrincipalModel() {}
+
+    public PrincipalModel(final String id, final List<ServerACLModel> acls, final TenantModel tenant) {
+        this.id = id;
+        this.acls.addAll(acls);
+        this.tenant = tenant;
+    }
+
+    public static PrincipalModel anonymous() {
```
coltmcnealy-lh, line 45: We need to have the ability to remove the abilities of the `anonymous` Principal. Is that possible here? I don't like the idea of having this method exposed here, because if we mark in the store that `anonymous` has been deleted, I could easily see how a sloppy method call could result in `anonymous` coming back from the dead with full permissions to do everything.

server/src/main/java/io/littlehorse/common/model/getable/global/acl/TenantModel.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private String id;
+
+    public TenantModel() {}
+
+    public TenantModel(final String id) {
+        this.id = id;
+    }
+
+    public static TenantModel createDefault() {
```
coltmcnealy-lh, line 31: This belongs in the `DAO`, not in the `TenantModel`. That will enable the DAO to determine whether the `DefaultTenant` has been deleted.

And again, creating stuff willy-nilly in memory is not a good idea especially when we add more fields to the `Tenant`.

server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/WfSpecModel.java
coltmcnealy-lh, line 1: As per comments in the other pull request, could we please separate the `WfSpec`/`TaskDef`/ExternalEventDef`/`UserTaskDef` from the `Principal`/`Tenant`?

They should definitely be in different code packages; additionally, I think they should inherit different sub-classes of `AbstractGetable`.

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/DeletePrincipalRequestModel.java
```diff
@@ -0,0 +1,43 @@
...
+        return DeletePrincipalRequest.class;
+    }
+
+    @Override
+    public boolean hasResponse() {
+        return false;
+    }
+
+    @Override
+    public Message process(MetadataProcessorDAO dao, LHServerConfig config) {
```
coltmcnealy-lh, line 39: Can you put some checking here to ensure that the `Principal` isn't the last "admin" left?

coltmcnealy-lh (approved):
Approved, but please address all unresolved comments before merging. Thank you!

server/src/main/java/io/littlehorse/common/ServerContextImpl.java
eduwercamacaro: `foo()` is preferred in this case

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/DeletePrincipalRequestModel.java
eduwercamacaro: This will be addressed here in LH-240

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/DeletePrincipalRequestModel.java
```diff
@@ -0,0 +1,54 @@
...
+        return DeletePrincipalRequest.class;
+    }
+
+    @Override
+    public boolean hasResponse() {
+        return false;
+    }
+
+    @Override
+    public Empty process(MetadataProcessorDAO dao, LHServerConfig config) {
```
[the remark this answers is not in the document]
eduwercamacaro: We will implement this in LH-240

PR #144, opened 2023-11-03
Feature/improve wfrun retention configs

coltmcnealy-lh:


coltmcnealy-lh:
Going to use my CEO power to merge this since it's a holiday in Ecuador so my team is out.

PR #145, opened 2023-11-05
Colt suggestion/multi tenancy acls

coltmcnealy-lh:
Eduwer—

Here are some of the changes that I would like to see based on your PR. I didn't have time this weekend to implement all of them, and I have other work I need to get done for the next few days, so this PR is incomplete.

However, I would like it if you could take a look at the changes and merge it into your branch. I think it gets us closer to our desired end goal. Here are some changes:

1. Update the `PrincipalModel` to have `map<string, ServerACLs> per_tenant_acls`, and also `ServerACLs global_acls`:
     - Global ACL's which enable the "Tyler" user
     - Per-tenant ACL's in a map which allow fine-grained ACL's depending on the tenant
2. Store the `PrincipalModel` outside of the tenant scope
    - makes it easier to do `lhctl whoami`
    - Makes it easier to do the Tyler user
3. Rename `ServerSubCommand` -> `ClusterLevelCommand`
4. Move some logic from `private static class AclVerifier` into the `PrincipalModel`
5. Fix a bunch of `switch` statements that were missing `cases`, and remove some unused private variables
6. Add validation to never delete the `anonymous` principal
    - rather than delete it, what we will do is remove all of its permissions
7. Remove the `PrincipalModel#anonymousFor()` method and move it to the DAO
    - You'll see a comment on the `MetadataProcessorDAOImpl`. I don't really like the way it's done. I think it would be better to do something along the lines of [this comment](https://github.com/littlehorse-enterprises/littlehorse/pull/128#discussion_r1372079502)


coltmcnealy-lh:
I know this is really bad etiquette as a software engineer, but as CEO...it's the best I can do given my other jobs. Edu, could you take it from here? Maybe merge this into your branch and continue on?

PR #152, opened 2023-11-21
Variable Index API Changes

coltmcnealy-lh:
Changes the SDK's for the Variable Index, and changes the server to match the appropriate changes.


schemas/wf_run.proto
```diff
@@ -12,26 +12,22 @@ option java_package = "io.littlehorse.sdk.common.proto";
...
   repeated PendingInterrupt pending_interrupts = 9;
   repeated PendingFailureHandler pending_failures = 10;
-
-  // Introduced now since with ThreadRun-level retention, we can't rely upon
-  // thread_runs.size() to determine the number of ThreadRuns.
-  int32 greatest_threadrun_number = 11;
-
-  // Keeps track of old versions of the WfSpec for this WfRun
-  repeated int32 old_wf_spec_versions = 12;
+  repeated WfSpecId old_wf_spec_versions = 11;
```
eduwercamacaro, line 30: order matters here?

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/WfRunVariable.java
```diff
@@ -1,53 +1,44 @@
...
-     * future versions of the WfSpec have the same variable with
-     * @return
+     * Marks the JSON_OBJ or JSON_ARR Variable as "Searchable", and creates an
+     * index on the specified field.
+     * @param fieldPath is the JSON Path to the field that we are indexing.
+     * @param fieldType is the type of the field we are indexing.
+     * @return same {@link WfRunVariable} instance
      */
-    WfRunVariable persistent();
+    WfRunVariable searchableOn(String fieldPath, VariableType fieldType);
```
eduwercamacaro, line 43: I'm wondering if we do a validation when we are indexing a Getable to make sure that the json attribute type matches with this.

schemas/wf_run.proto
coltmcnealy-lh: Yes, order matters, if we migrate back. This field will serve as a log of what versions we were on, etc.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/WfRunVariable.java
coltmcnealy-lh: We will do that. The question is how do we handle when validation fails. I don't think it should be an error that stops the workflow. We need a way to send a "warning" to the user.

coltmcnealy-lh: We will do that. The question is how do we handle when validation fails. I don't think it should be an error that stops the workflow. We need a way to send a "warning" to the user.

eduwercamacaro (approved):

PR #159, opened 2023-11-29
Add specific ACL resources and actions to grpc service methods

eduwercamacaro:
- Added Java doc for `@Authorize` annotation
- Added specific set of actions and resources for grpc service methods.


sauljabin (approved):
It looks very interesting, I would like to see the logic around approving or rejecting the access

coltmcnealy-lh (approved):

PR #163, opened 2023-12-01
Fix task worker assignor

sauljabin:
- Now the assignor is idempotent
- It assigns the server in each heartbeat


server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/TaskWorkerHeartBeatRequestModel.java
```diff
@@ -40,7 +39,9 @@ public class TaskWorkerHeartBeatRequestModel extends CoreSubCommand<TaskWorkerHe
     @Setter // for unit test
     private String clientId;
 
+    @Setter // for unit test
```
eduwercamacaro, line 42: can we remove this? I don't see any unit test using this setter

eduwercamacaro (approved):

PR #165, opened 2023-12-02
Child Workflows With Strong Parent Reference

coltmcnealy-lh:


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=0.7.0-alpha.3
+version=special-test
```
mijailr, line 1: Is this change intended?

mijailr (approved):
Looks good to me, would be great to have an example though.

gradle.properties
sauljabin, line 1: revert this please

sauljabin (changes requested):

gradle.properties
coltmcnealy-lh: Still in progress this PR, let me update it to draft

coltmcnealy-lh: It's missing tests and examples so I can't merge it. But I will remove that marker.

sauljabin: OK, change it to draft 👍🏽

coltmcnealy-lh:
This is ready for review. Only thing missing is python and go sdk's; I have created a ticket and will implement it this weekend. Let's get this merged to unblock the control plane team.

sauljabin (approved):

PR #167, opened 2023-12-04
Nelson - fixed node placement for paralell approval flow

diablouma:
- fixed node that was misplaced
- fixed width of the node edges


coltmcnealy-lh (approved):
Thanks Nelson!

PR #174, opened 2023-12-11
Propagate authorization for interactive queries

eduwercamacaro:


coltmcnealy-lh (approved):
This looks clean. Maybe a good idea for @sauljabin to look at the Interceptor code, just to check for any potential security flaws? But I'm ok with it being merged as-is.

sauljabin (approved):

PR #175, opened 2023-12-13
Enhancement/cleanup var val proto

coltmcnealy-lh:
Moves the Variable Value Proto to a `oneof`, which is proper:

```
message VariableValue {
  reserved 1;
  oneof value {
    string json_obj = 2;
    string json_arr = 3;
    double double = 4;
    bool bool = 5;
    string str = 6;
    int64 int = 7;
    bytes bytes = 8;
  }
}
```

In the past, it was all optionals.


schemas/internal/job.proto
```diff
@@ -4,8 +4,6 @@ package littlehorse;
 option java_multiple_files = true;
 option java_package = "io.littlehorse.common.proto";
 
-import "object_id.proto";
```
eduwercamacaro, line 7: oops

eduwercamacaro (approved):

PR #176, opened 2023-12-13
updates lhctl version to the next upcoming release.

coltmcnealy-lh:
TODO: this should be automated.


sauljabin (changes requested):
adding it

sauljabin (approved):
The solution that I have in mind is more complicated, so let's create a technical dept ticket. I'm focus in the operator

PR #547, opened 2023-12-22
Idempotency on PutWfSpec

### LINKED ISSUE #544: Idempotency on PutWfSpec
### Context

In order to reduce duplication in the WfSpec versions, it is proposed to implement idempotence in the PutWfSpec service.

### Proposal

Add a new field to `PutWfSpecRequest` proto message, which allows you to define what type of update you want to execute.

Allow all changes:

1. Doesn't exist: just create it
2. Exists, and is the same:
a. Don't change anything
b. Return OK
3. Exists, and is different:
a. Create a new WfSpec with a different version (either major or revision)
b Return OK

Allow only minor revision:

1. Doesn't exist: just create it
2. Exists, and is the same:
a. Don't change anything
b. Return OK
3. Exists, and is different and change the revision:
a. Create a new WfSpec with a different version
b. Return OK
4. Exists, and is different and change the majorVersion:
a. Fail with the FAILED_PRECONDITION response code.

Reject changes:

1. Doesn't exist: just create it
2. Exists, and is the same:
a. Don't change anything
b. Return OK
3. Exists, and is different:
a. Fail with the ALREADY_EXISTS response code.

### Scope

This issue is scoped to the procedure PutWfSpec. Out of scope saving the policy in the storage.

### Component

* [x] LH Server
* [x] Python SDK
* [x] Go SDK
* [x] Java SDK
* [x] C# SDK

### Links

- https://grpc.github.io/grpc/core/md_doc_statuscodes.html

mijailr:
Introduces three update types to PutWFSpec:
- `ALL`
- `MINOR_REVISION_ONLY`
- `NONE`

## Allow all changes: (`ALL`)

- Doesn't exist: just create it
- Exists, and is the same:
  1. Don't change anything
  2. Return OK
- Exists, and is different:
  1. Create a new WfSpec with a different version (either major or revision)
  2. Return OK

## Allow only minor revision: (`MINOR_REVISION_ONLY`)

- Doesn't exist: just create it
- Exists, and is the same:
  1. Don't change anything
  2. Return OK
- Exists, and is different and change the revision:
  1. Create a new WfSpec with a different version
  2. Return OK
- Exists, and is different and change the majorVersion:
  1. Fail with the `FAILED_PRECONDITION` response code.

## Reject changes: (`NONE`)

- Doesn't exist: just create it
- Exists, and is the same:
  1. Don't change anything
  2. Return OK
- Exists, and is different:
  1. Fail with the `ALREADY_EXISTS` response code.

Resolves #544


examples/external-event/src/main/java/io/littlehorse/examples/ExternalEventExample.java
```diff
@@ -35,14 +35,14 @@ public static Workflow getWorkflow() {
         return new WorkflowImpl(
             "example-external-event",
             wf -> {
-                WfRunVariable name = wf.addVariable("name", VariableType.STR).searchable();
+                WfRunVariable name = wf.addVariable("person", VariableType.STR).searchable();
```
eduwercamacaro, line 38: Probably we don't want this

schemas/service.proto
```diff
@@ -33,6 +33,13 @@ message PutWfSpecRequest {
   string entrypoint_thread_name = 6;
   optional WorkflowRetentionPolicy retention_policy = 8;
   reserved 9;
+  optional AllowedUpdateType allowed_updates = 10;
+}
+
+enum AllowedUpdateType {
+  ALL = 0;
```
eduwercamacaro, line 40: Comments will be useful. And I believe this will add javadoc for the generated Java enum.

sdk-go/examples/basic/README.md
```diff
@@ -1,4 +1,4 @@
-# Basic Example
+ # Basic Example
```
eduwercamacaro, line 1: is this expected?

sdk-python/tests/test_workflow.py
```diff
@@ -814,6 +814,7 @@ def my_entrypoint(thread: WorkflowThread) -> None:
         self.assertEqual(
             wf.compile(),
             PutWfSpecRequest(
+                allowed_updates="ALL",
```
eduwercamacaro, line 817: Is not clear if we are covering the case when this function is invoked without the allowed_updates argument

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutWfSpecRequestModel.java
```diff
@@ -61,6 +66,7 @@ public void initFrom(Message proto, ExecutionContext context) {
         PutWfSpecRequest p = (PutWfSpecRequest) proto;
         name = p.getName();
         entrypointThreadName = p.getEntrypointThreadName();
+        allowedUpdateType = p.hasAllowedUpdates() ? p.getAllowedUpdates() : AllowedUpdateType.ALL;
```
eduwercamacaro, line 69: suggestion:
what if we add a private constant for AllowedUpdateType.ALL?

maybe smth like this:

private static AllowedUpdateType defaultUpdateType = ...

examples/external-event/src/main/java/io/littlehorse/examples/ExternalEventExample.java
mijailr: I reverted this file back to original

server/src/main/java/io/littlehorse/common/util/WfSpecUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import io.littlehorse.common.model.getable.global.wfspec.WfSpecModel;
+import io.littlehorse.common.model.getable.global.wfspec.thread.ThreadVarDefModel;
+import io.littlehorse.sdk.common.proto.WfSpec;
+import io.littlehorse.sdk.common.proto.WfSpecId;
+import java.util.Arrays;
+import java.util.Date;
+import java.util.Map;
+import java.util.Set;
+
+public class WfSpecUtil {
```
eduwercamacaro, line 13: suggestion:

add a private constructor for this class to prevent class instances since it only contains static methods

eduwercamacaro (approved):
Good job. just a few suggestions :)

sdk-python/tests/test_workflow.py
mijailr: I'll delegate this to @sauljabin

coltmcnealy-lh: The way protobuf works is that it's not an `optional` field, so if you don't set it, then it is treated as the protobuf "zero" field. That means the enum value `0`; in this case, allowing `ALL` udpates.

sdk-go/examples/basic/workflow.go
```diff
@@ -14,7 +14,7 @@ func Greet(name string) string {
 }
 
 func MyWorkflow(wf *wflib.WorkflowThread) {
-	nameVar := wf.AddVariableWithDefault("name", model.VariableType_STR, "Qui-Gon Jinn")
```
coltmcnealy-lh, line 17: This will break the readme for the example, please revert this change.

schemas/service.proto
```diff
@@ -33,6 +33,17 @@ message PutWfSpecRequest {
...
   optional WorkflowRetentionPolicy retention_policy = 8;
   reserved 9;
+  optional AllowedUpdateType allowed_updates = 10;
+}
+
+// Defines the allowed update type
+// ALL - Creates either a revision or majorVersion when WfSpec changes
+// MINOR_REVISION_ONLY - Allow revision updates but reject majorVersion
+// NONE - Reject any update
+enum AllowedUpdateType {
```
coltmcnealy-lh, line 43: One quirk with protobuf is that we won't be able to have any other enum value in our entire package that is called `ALL`, `MINOR_REVISION_ONLY`, or `NONE`.

I definitely don't think we should reserve `ALL` or `NONE` for this use case. Can we please put a prefix? Or make it `ALL_UPDATES`, `MINOR_REVISION_UPDATES`, and `NO_UPDATES`?

sdk-python/tests/test_workflow.py
```diff
@@ -1160,6 +1166,7 @@ def my_entrypoint(thread: WorkflowThread) -> None:
...
```
coltmcnealy-lh, line 1169: As per the comment above, if you don't set this field, it defaults to `ALL`, so there's not really any need to set it.

schemas/service.proto
```diff
@@ -33,6 +33,17 @@ message PutWfSpecRequest {
   string entrypoint_thread_name = 6;
   optional WorkflowRetentionPolicy retention_policy = 8;
   reserved 9;
+  optional AllowedUpdateType allowed_updates = 10;
```
coltmcnealy-lh, line 36: Please make this not an optional field. It will default to the enum value zero. That is backwards compatible.

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutWfSpecRequestModel.java
```diff
@@ -30,6 +32,9 @@ public class PutWfSpecRequestModel extends MetadataSubCommand<PutWfSpecRequest>
     public Map<String, ThreadSpecModel> threadSpecs;
     public String entrypointThreadName;
     public WorkflowRetentionPolicyModel retentionPolicy;
+    public AllowedUpdateType allowedUpdateType;
+
+    private static AllowedUpdateType defaultUpdateType = AllowedUpdateType.ALL;
```
coltmcnealy-lh, line 37: This shouldn't be necessary.

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutWfSpecRequestModel.java
```diff
@@ -97,10 +105,32 @@ public WfSpec process(MetadataCommandExecution executionContext) {
...
+            return optWfSpec.get().toProto().build();
+        }
+
+        verifyUpdateType(allowedUpdateType, spec, optWfSpec);
         metadataManager.put(spec);
         return spec.toProto().build();
     }
 
+    private void verifyUpdateType(AllowedUpdateType updateType, WfSpecModel spec, Optional<WfSpecModel> oldSpec) {
+        switch (updateType) {
```
coltmcnealy-lh, line 118: I'd like to have a checkstyle rule which throws an error if any enum types are missing from a switch. Could we please address that here?

coltmcnealy-lh (changes requested):
Several small changes requested, mostly around how proto works.

schemas/service.proto
mijailr: Oh understood, will do.

server/src/main/java/io/littlehorse/common/util/WfSpecUtil.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param right WfSpecModel compared against
+     * @return true when it is equal, false otherwise.
+     */
+    public static boolean equals(WfSpecModel left, WfSpecModel right) {
+        WfSpec.Builder copy = left.toProto();
+        Timestamp date = LHUtil.fromDate(new Date());
+        sanitize(copy, date);
+        WfSpec.Builder toCopy = right.toProto();
+        sanitize(toCopy, date);
+        return Arrays.equals(copy.build().toByteArray(), toCopy.build().toByteArray());
```
coltmcnealy-lh, line 29: This is actually a lot nicer than I thought it would come out to be. Good work!

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutWfSpecRequestModel.java
mijailr: You mean this?

image

coltmcnealy-lh: Yes, that. Thanks!

coltmcnealy-lh (approved):
Going to pre-emptively approve this, please merge once you've addressed the prior feedback.

PR #550, opened 2023-12-22
Go SDK Variable Type Fix

coltmcnealy-lh:
Previously, the common.GetVarType() function did not work when a VariableValue was passed in. This was due to improper inspection of the Go Protobuf Oneof, which has been fixed and tested in this PR.


sauljabin (approved):
Is it possible to add a unit test at least?

Approved

coltmcnealy-lh:
Tests already included

PR #556, opened 2023-12-24
REFACTOR: New Set of Stores

coltmcnealy-lh:
The goal of this PR is to introduce a new set of Store classes that wrap over the Kafka Streams `KeyValueStore`. As a refresher, I first introduced store wrappers over a year ago for performance reasons. Rather than have a store for `WfRun`, `NodeRun`, `TaskRun`, `ExternalEvent`, `Tag`, etc (which would now be 21 stores), we use prefixes within the same State Store. This has the following advantages:

- Instead of 21 changelog partitions per input partition, we have 1 changelog partition per input partition.
- More lightweight consumer rebalances.
- Fewer partitions involved in each Kafka Transaction, which means faster Streams Commits and better throughput.
- Fewer open files on the server from RocksDB
- Less stress on the Kafka Cluster (fewer partitions == better).

With @eduwercamacaro 's introduction of Multi-Tenancy, we needed to add additional functionality to the stores: some objects, like WfRun/WfSpec, need to be isolated within a tenant, and others, such as `Principal` and `Tenant`, need to be stored at the cluster level. That introduced a set of stores.

This PR removes some ambiguity about whether a store is tenant or cluster-scoped. To do so, we are proposing the following store structure:

* `ReadOnlyClusterScopedStore` -> allows reading objects without a tenant prefix.
* `ReadOnlyTenantScopedStore` -> allows reading objects with a tenant prefix.
* `ClusterScopedStore` -> allows reading and writing objects without a tenant prefix.
* `TenantScopedStore` -> allows reading and writing objects with a tenant prefix.

These stores will be used internally by all of the implementations of `ExecutionContext`. They will generally not be passed around in the code.


eduwercamacaro:
This PR adds a total replacement for the current store. I think this is important to mention in the PR description.
From a store user* perspective, this has a very similar design.


```
ReadOnlyModelStore  --> ReadOnlyBaseStore
ModelStore -- BaseStore
ReadOnlyModelDefaultStore --> ReadOnlyClusterScopedStore
DefaultModelStore --> ClusterScopedStore
TenantModelStore --> TenantScopedStore
ReadOnlyTenantStore --> ReadOnlyTenantScopedStore
```


In your proposal, you are ignoring this method:

`<U extends Message, T extends AbstractGetable<U>> StoredGetable<U, T> get(ObjectIdModel<?, U, T> id)`

In my opinion, this method is important to include because it adds an easy way to query getable objects; otherwise, users will need to pass the key as a string.
I want to avoid users representing the key as a string.

Before this, I was planning to remove every reference to the existing ModelStore and instead use the specific
Tenant or Default store.

Another point is that I don't really like the term `ClusterStore`; for me, it is not clear what its real intention is. I'm Considering whether to use a name like `SystemStore` or keep the name `DefaultStore`, because in this store we persist system-related objects such as tenant and principal, cache objects and it is highly probably that in the future we will need more system objects.

Please add tests for these classes to show that it actually works and is usable. 


* In this context, the user is a developer that is contributing to the server source code

eduwercamacaro (changes requested):
I know that this is just a proposal and you are not affecting any existing flow yet. But, In order to make the motivation more explicit and how these classes will be used I would like to see some unit test for these changes before merging into master.

coltmcnealy-lh:
> In your proposal, you are ignoring this method:
> 
> `<U extends Message, T extends AbstractGetable<U>> StoredGetable<U, T> get(ObjectIdModel<?, U, T> id)`
> 
> In my opinion, this method is important to include because it adds an easy way to query getable objects; otherwise, users will need to pass the key as a string. I want to avoid users representing the key as a string.

Absolutely not. That method is dangerous since it allows, for example, getting a `WfSpec` from a core store, or getting a `WFRun` from the global store.

The only classes that should interact with the stores in this PR are the four `Context`s: Metadata, Core, Repartition, and Request. (And also the timer processor, but that's a different story).

coltmcnealy-lh:
> I know that this is just a proposal and you are not affecting any existing flow yet. But, In order to make the motivation more explicit and how these classes will be used I would like to see some unit test for these changes before merging into master.

Sure, in a few hours I'll push my local changes which have everything working with the new stores.

coltmcnealy-lh:
@eduwercamacaro it is passing tests and pipeline is green. I am going to continue now fixing the `SearchTaskDef` pagination now that we don't have ambiguity between tenant-scoped and cluster-scoped stores anymore, since I was unable to make progress on that bug with the previous state of the code.

server/src/main/java/io/littlehorse/server/streams/storeinternals/ReadOnlyMetadataManager.java
```diff
@@ -24,20 +25,45 @@
...
+        this.tenantStore = tenantStore;
     }
 
-    public <U extends Message, T extends GlobalGetable<U>> T get(MetadataId<?, U, T> id) {
+    public <U extends Message, T extends MetadataGetable<U>> T get(MetadataId<?, U, T> id) {
         log.trace("Getting {} with key {}", id.getType(), id);
         T out = null;
-        ReadOnlyModelStore specificStore = isClusterLevelObject(id) ? defaultStore : tenantStore;
+        // First check the cache.
+        @SuppressWarnings("unchecked")
```
sauljabin, line 41: in some point we should remove this suppress right?

sauljabin (approved):
I guess from now we should make smaller PR

server/src/main/java/io/littlehorse/server/streams/stores/ReadOnlyBaseStoreImpl.java
```diff
@@ -0,0 +1,123 @@
...
+ */
+abstract class ReadOnlyBaseStoreImpl implements ReadOnlyBaseStore {
+
+    @Getter
+    protected final String tenantId;
+
+    protected final ExecutionContext executionContext;
+    private final ReadOnlyKeyValueStore<String, Bytes> nativeStore;
+
+    ReadOnlyBaseStoreImpl(
```
eduwercamacaro, line 33: This is not a good way to represent optional arguments for constructors. Instead, you should create two constructors:

- `ReadOnlyBaseStoreImpl(ReadOnlyKeyValueStore<String, Bytes> nativeStore, ExecutionContext executionContext)`
- `ReadOnlyBaseStoreImpl(ReadOnlyKeyValueStore<String, Bytes> nativeStore,String tenantId,ExecutionContext executionContext)`

 Optional is not intended to be a replacement for every null reference. 

Sames applies for specific classes of these stores.

eduwercamacaro: Also, I propose reducing the access modifier for these constructors to protected.

server/src/main/java/io/littlehorse/common/model/repartitioncommand/RepartitionSubCommand.java
```diff
@@ -1,10 +1,10 @@
 package io.littlehorse.common.model.repartitioncommand;
 
-import io.littlehorse.server.streams.store.ModelStore;
+import io.littlehorse.server.streams.stores.TenantScopedStore;
 import org.apache.kafka.streams.processor.api.ProcessorContext;
 
 public interface RepartitionSubCommand {
-    public void process(ModelStore repartitionedStore, ProcessorContext<Void, Void> ctx);
+    public void process(TenantScopedStore repartitionedStore, ProcessorContext<Void, Void> ctx);
```
eduwercamacaro, line 7: Metrics are stored in a "Cluster" store. how are you planning to do that with this design?

eduwercamacaro: Or are you planning to store it into the "default" tenant?

server/src/main/java/io/littlehorse/common/model/repartitioncommand/repartitionsubcommand/AggregateWfMetricsModel.java
```diff
@@ -78,11 +76,18 @@ public String getPartitionKey() {
...
 
     @Override
-    public void process(ModelStore repartitionedStore, ProcessorContext<Void, Void> ctx) {
+    public void process(TenantScopedStore repartitionedStore, ProcessorContext<Void, Void> ctx) {
         for (WfMetricUpdateModel metricUpdate : metricUpdates) {
-            StoredGetable<WfSpecMetrics, WfSpecMetricsModel> storedMetrics =
-                    repartitionedStore.get(WfSpecMetricsIdModel.getObjectId(
-                            metricUpdate.getWindowStart(), metricUpdate.getWindowType(), metricUpdate.getWfSpecId()));
+
+            // TODO: We should NOT do this. The RepartitionContext should do this for us. We also shouldn't
```
eduwercamacaro, line 82: Ok, maybe this is related to my other comment about passing TenantStore here

server/src/main/java/io/littlehorse/server/streams/storeinternals/MetadataManager.java
```diff
@@ -3,67 +3,102 @@
...
+import io.littlehorse.server.streams.stores.ClusterScopedStore;
+import io.littlehorse.server.streams.stores.TenantScopedStore;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class MetadataManager extends ReadOnlyMetadataManager {
 
-    private ModelStore defaultStore;
-    private ModelStore tenantStore;
+    private ClusterScopedStore clusterStore;
```
eduwercamacaro, line 20: final-first approach: mark these field as final.

server/src/main/java/io/littlehorse/server/streams/storeinternals/ReadOnlyMetadataManager.java
eduwercamacaro: IMO, `GetableToStore` shouldn't have a generic definition at class level. From what I understand, every class instance for this class only represent a row that will be store into the ModelStore, there is no point on setting a generic here because we are putting these instances in a buffer like this:

`Map<String, GetableToStore<?, ?>> uncommittedChanges.`

But then, you will be forced on casting objectToStore anyways. Unchecked casting are hard to avoid with this design, which tells me that we should think carefully every time that we define a generic class.

coltmcnealy-lh: This needs a ticket, also remove generic from `LHSerializable<>`

eduwercamacaro (approved):

PR #559, opened 2023-12-26
Docs/add comments to grpc

coltmcnealy-lh:
Gets a start adding comments to our GRPC code. This is useful because the comments will now show up in the automatically-generated javadoc and godoc for our GRPC clients.

TODO: we need to figure out whether it's possible to get the Python Protobuf Compiler to automatically generate docs for the python client.


schemas/service.proto
coltmcnealy-lh, line 1: Please review this carefully even though it's a big diff.

eduwercamacaro (approved):
LGTM

PR #560, opened 2023-12-26
PutUserTaskDefRequest idempotency

mijailr:
Adds idempotency to UserTask.

It will accept and compare a new userTask request to the previous version stored and verify equality using `UserTaksUtil.equals` static method.

If equals returns true, then no new version will be created and the current spec will be returned.
If equals returns false, then a new version will be created

It includes utils tests and e2e tests and updates `examples/user-task` to remove no longer required existence checks.

### DIFF
--- a/examples/user-tasks/src/main/java/io/littlehorse/examples/UserTasksExample.java
+++ b/examples/user-tasks/src/main/java/io/littlehorse/examples/UserTasksExample.java
@@ -16,12 +16,8 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.util.Properties;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class UserTasksExample {
-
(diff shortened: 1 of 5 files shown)


examples/user-tasks/src/main/java/io/littlehorse/examples/UserTasksExample.java
```diff
@@ -158,45 +154,20 @@ public void doMain() throws IOException {
         // New worker
         LHTaskWorker worker = getTaskWorker(config);
 
-        // Register task if it does not exist
```
coltmcnealy-lh, line 161: I love red diff's in our examples (:

coltmcnealy-lh (approved):
Well done!

coltmcnealy-lh:
On the merge commit, one thing: please say `PutUserTaskDefRequest Idempotency` as opposed to `User Task Idempotency` as the second one is a bit ambiguous and might be confusing.

PR #565, opened 2023-12-26
Send task def metrics

eduwercamacaro:


coltmcnealy-lh (approved):
Cleaned it up and approved

PR #566, opened 2023-12-27
Metadata Processor Integrity E2E Test

coltmcnealy-lh:
Adds end to end test to verify integrity of metadata processor


coltmcnealy-lh:
In Slack, @sauljabin and @mijailr reported behavior in which we needed to put a `sleep()` between the `PutTaskDef` and `PutWfSpec` RPC calls. I was unable to reproduce that bug on the `master` branch. Perhaps it's been fixed recently?

Anyways, this test should verify what you were concerned about.

mijailr (approved):
Looks good

PR #567, opened 2023-12-27
Fixes Metadata Search Pagination

coltmcnealy-lh:
Metadata Search Pagination was broken by ambiguity between cluster-scoped and tenant-scoped stores when we re-added the `default/` prefix back for the `default` `Tenant`.

This PR is a TEMPORARY fix for this issue which hard-codes correct logic to handle metadata searches. The combination of #556 PR and the planned follow-up will remove this hackiness.

This PR also adds an e2e test which verifies that pagination works for metadata search.


sauljabin:
there is a test failing, flaky?

sauljabin (approved):

coltmcnealy-lh:
Green now :smile:

PR #571, opened 2023-12-27
Fixes if/else without Node on Go SDK

MateoRojas:


coltmcnealy-lh (approved):
This was very tricky. Good work

PR #572, opened 2023-12-27
Wf metrics latency

eduwercamacaro:
This PR calculate the average delay from a WfRun start to WfRun Complete. Similar calculation for task metrics


coltmcnealy-lh (approved):
:ship:

Issue #578, opened 2024-01-04
Allow implementations of the SDK to send tenantId headers in the GRPc calls


eduwercamacaro:
### Context

To provide user isolation per tenancy, the LH server optionally takes the 'tenantId' header on a single grpc request. When tenantId is not present LH server handle the request as a default tenant isolation.

### Description

**As a** LH user
**I want to** store my data in an specific LH tenant
**So that** my data is not accessible to other users in other tenants

### Acceptance Criteria

**Scenario** Register a new WfSpec
**Given** A user belonging to an specific tenant 
**When** register a new WfSpec
**Then** The WfSpec is stored in their specific tenant
**And** is not accessible to other tenants

### Components

- [ ] Dashboard
- [ ] Server
- [X] Python SDK
- [X] Go SDK
- [X] Java SDK
- [ ] C# SDK
- [ ] LH Control
- [ ] LH Tests Utils

PR #579, opened 2024-01-05
Enable SDK Java clients sending tenantId header

eduwercamacaro:
This PR add a new class (TenantMetadataProvider) that will intercept grpc request on the java client side and add metadata for the client tenant
https://github.com/littlehorse-enterprises/littlehorse/issues/578

### DIFF
--- a/sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
+++ b/sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
@@ -7,6 +7,7 @@ import io.grpc.TlsChannelCredentials;
 import io.littlehorse.sdk.common.auth.OAuthClient;
 import io.littlehorse.sdk.common.auth.OAuthConfig;
 import io.littlehorse.sdk.common.auth.OAuthCredentialsProvider;
+import io.littlehorse.sdk.common.auth.TenantMetadataProvider;
 import io.littlehorse.sdk.common.proto.LittleHorseGrpc;
 import io.littlehorse.sdk.common.proto.LittleHorseGrpc.LittleHorseBlockingStub;
 import io.littlehorse.sdk.common.proto.LittleHorseGrpc.LittleHorseStub;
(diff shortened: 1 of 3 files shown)


sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
```diff
@@ -40,6 +41,9 @@ public class LHConfig extends ConfigBase {
     /** The Client Id. */
     public static final String CLIENT_ID_KEY = "LHC_CLIENT_ID";
 
+    /** The Tenant Id for this client, null will be used if not set */
```
coltmcnealy-lh, line 44: Do we want `null` or `default`? How do we handle `null`...in some cases, perhaps we can "infer" the tenant id from the principal.

For example, if Principal `alice` only has access to the `foo` Tenant, then does `alice` need to specify this header? Automatically inferring it would be nice, but it could also be dangerous, because if someone edits the `alice` principal to give the principal permissions to a new Tenant, then existing applications would break.

sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
```diff
@@ -268,6 +272,27 @@ private Channel getChannel(String host, int port) throws IOException {
...
+    private LittleHorseBlockingStub getDefaultBlockingStub(String host, int port) throws IOException {
+        String tenantId = getTenantId();
+        LittleHorseBlockingStub blockingStub = LittleHorseGrpc.newBlockingStub(getChannel(host, port));
+        if (tenantId != null) {
+            return blockingStub.withCallCredentials(new TenantMetadataProvider(tenantId));
+        }
+        return blockingStub;
+    }
+
+    private LittleHorseStub getDefaultAsyncStub(String host, int port) throws IOException {
```
coltmcnealy-lh, line 287: What does "default" mean here?

sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
eduwercamacaro: I suggest keep this as `null`, in that way there is only one place (server side) that determine the tenant Id when the header is not present.

sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
eduwercamacaro: The application default setup to communicate with the server. method clients can add more configuration on top of this default configurations. For example, this method can return the default setup and on top of this we can add the OAuth configuration.

coltmcnealy-lh (approved):

PR #595, opened 2024-01-11
Feature/594 runtime env configs

diablouma:
* Using multi stage docker build for the dashboard
* Using runtime environment variables to configure if the enable/disable the OAUTH for the dashboard
* Use LHD_OAUTH_ENABLED as the config name instead of AUTHENTICATION_ENABLED


dashboard/Dockerfile
coltmcnealy-lh, line 1: Could we move this to the `/build` directory in the server? Just trying to standardize everything.

docker/dashboard/Dockerfile
```diff
@@ -1,14 +1,25 @@
-FROM node:20
+FROM node:20 AS base
```
coltmcnealy-lh, line 1: Ah, I see this is here. Could we delete the other `Dockerfile` then and just use this one? The rest of the repo only has dockerfiles here.

coltmcnealy-lh (approved):
Approved but please address the Dockerfile comments before merging.

docker/dashboard/Dockerfile
diablouma: you are right @coltmcnealy-lh I forgot to delete the other file.

dashboard/Dockerfile
diablouma: just deleted it, as it should be under the docker folder

PR #600, opened 2024-01-16
adds docs to user_tasks.proto so that we get javadoc on the generated…

coltmcnealy-lh:
These docs show up in the generated Java and Go clients.


schemas/user_tasks.proto
coltmcnealy-lh, line 1: This is the only file that requires review. The rest were auto-generated.

schemas/user_tasks.proto
```diff
@@ -34,81 +62,170 @@ message UserTaskField {
...
+  // UserTaskRun. If user_id is set, then the UserTaskRun cannot be in the
+  // UNASSIGNED status.
   optional string user_id = 4;
 
+  // The results of the UserTaskRun. Empty if the UserTaskRun has not yet been completed.
+  // Each key in this map is the `name` of a corresponding `UserTaskField` on the
+  // UserTaskDef.
   map<string, VariableValue> results = 6;
+
+  // Status of the UserTaskRun. Can be UNASSIGNED, ASSIGNED, DONE, or CANCELLED.
```
sauljabin, line 90: this is nice

sauljabin (approved):

schemas/user_tasks.proto
coltmcnealy-lh: That was a team effort...we spent hours in design meetings trying to figure it out :sweat_smile:

PR #611, opened 2024-01-17
More GRPC Docs

coltmcnealy-lh:


dashboard/apps/web/littlehorse-public-api/variable.ts
```diff
@@ -327,7 +327,7 @@ export type Exact<P, I extends P> = P extends Builtin ? P
 
 function toTimestamp(dateStr: string): Timestamp {
   const date = new globalThis.Date(dateStr);
-  const seconds = date.getTime() / 1_000;
+  const seconds = Math.trunc(date.getTime() / 1_000);
```
coltmcnealy-lh, line 330: Why is this changed? Did someone change the `protoc` version?

dashboard/apps/web/littlehorse-public-api/acls.ts
```diff
@@ -1016,7 +1016,7 @@ export type Exact<P, I extends P> = P extends Builtin ? P
 
 function toTimestamp(dateStr: string): Timestamp {
   const date = new globalThis.Date(dateStr);
-  const seconds = date.getTime() / 1_000;
+  const seconds = Math.trunc(date.getTime() / 1_000);
```
coltmcnealy-lh, line 1019: This looks like a `protoc` version change. Weird. @sauljabin any ideas?

coltmcnealy-lh:
Team, this is ready to go. @eduwercamacaro remember when we were on zoom, we made a small change to `LHServerConnectionManager` when trying to fix the way the exception stacktrace was displayed? Well, that ended up breaking two end-to-end tests. One of them was because it was looking for the stacktrace in the wrong place (test issue), and the other because somehow the WfRun didn't fail (which is an actual bug).

I reverted that change, so this PR is strictly a docs PR. But we need to fix that.

eduwercamacaro (approved):

PR #615, opened 2024-01-19
Builds and pushes lh-standalone image for x86 and ARM platforms and fixes dashboard

MateoRojas:
All our images are built for the linux/amd64 platform because that's the platform of the Github Agent. This means that our images do not work on ARM-based Mac computers (e.g. M1, M2, etc.). That is a big problem considering that a lot of developers use Macs for their day to day work.

This PR introduces the `docker/build-push-action` action to help us build a multiplatform image (namely linux/amd64 and linux/arm64) for the `lh-standalone` image. We should probably consider doing all our other images multiplatform, the major drawback is that the image size seems to become quite bigger.

Also, the dashboard was not working on `lh-standalone`. That's also fixed as part of this PR.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -71,7 +71,9 @@ jobs:
           npm install pnpm --global
           pnpm install
           pnpm build
-#      - name: Tests
+      - name: Build server
+        run: ./gradlew server:shadowJar -x test
+#      - name: Tests server
 #        run: ./gradlew server:test
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,7 @@ on:
   push:
     tags:
       - "[0-9]+.[0-9]+.[0-9]+*" # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
+
 jobs:
   publish-sdk-java:
     runs-on: ubuntu-latest
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -71,10 +71,6 @@ jobs:
           npm install pnpm --global
           pnpm install
           pnpm build
-      - name: Setup Gradle
-        uses: gradle/gradle-build-action@v2
-        with:
-          gradle-version: wrapper
       - name: Build server
         run: ./gradlew server:shadowJar -x test
 #      - name: Tests server
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -107,7 +107,7 @@ jobs:
         with:
           push: true
           tags: "${{ steps.login-ecr-public.outputs.registry }}/littlehorse/lh-proof-of-concept:latest"
-#          platforms: "linux/amd64,linux/arm64"
+          platforms: "linux/amd64,linux/arm64"
           cache-from: type=gha
           cache-to: type=gha,mode=max
           file: ./docker/standalone/Dockerfile
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -71,6 +71,10 @@ jobs:
           npm install pnpm --global
           pnpm install
           pnpm build
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+        with:
+          gradle-version: 8.2
       - name: Build server
         run: ./gradlew server:shadowJar -x test
 #      - name: Tests server
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -108,9 +108,7 @@ jobs:
           push: false
           tags: latest
           labels: latest
-          platforms:
-            - linux/amd64
-            - linux/arm64
+          platforms: "linux/amd64,linux/arm64"
           cache-from: type=gha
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -72,8 +72,12 @@ jobs:
           npm install pnpm --global
           pnpm install
           pnpm build
-      - name: Tests
-        run: ./gradlew server:test
+      - name: Build & Test Server
```
MateoRojas, line 75: Building the jar as part of the pipeline drastically reduces the amount of time the `docker build` for the `linux/arm64` platform. Also, this same jar could be leverage for the `lh-sever` image instead of building it there too.

.github/workflows/publish.yml
```diff
@@ -72,8 +72,12 @@ jobs:
...
+        run: ./gradlew server:test server:shadowJar
+      - name: Set up QEMU
```
MateoRojas, line 77: QEMU is needed to emulate an ARM architecture inside x86

.github/workflows/publish.yml
```diff
@@ -72,8 +72,12 @@ jobs:
...
+        uses: docker/setup-qemu-action@v1
+      - name: Set up Docker Buildx
```
MateoRojas, line 79: BuildX (BuildKit) is needed to be able to build multiplatform images

.github/workflows/publish.yml
```diff
@@ -100,16 +104,15 @@ jobs:
...
-          ECR_REGISTRY: ${{ steps.login-ecr-public.outputs.registry }}
-          ECR_REGISTRY_ALIAS: littlehorse
-          ECR_REPOSITORY: lh-standalone
-          IMAGE_TAG: ${{ github.ref_name }}
-        run: |
-          docker build -f docker/standalone/Dockerfile -t $ECR_REGISTRY/$ECR_REGISTRY_ALIAS/$ECR_REPOSITORY:$IMAGE_TAG .
-          docker tag $ECR_REGISTRY/$ECR_REGISTRY_ALIAS/$ECR_REPOSITORY:$IMAGE_TAG $ECR_REGISTRY/$ECR_REGISTRY_ALIAS/$ECR_REPOSITORY:latest
-          docker push $ECR_REGISTRY/$ECR_REGISTRY_ALIAS/$ECR_REPOSITORY:$IMAGE_TAG
-          docker push $ECR_REGISTRY/$ECR_REGISTRY_ALIAS/$ECR_REPOSITORY:latest
+        uses: docker/build-push-action@v5
```
MateoRojas, line 107: Leveraging an already existing and official action to build and push the image. This seems to be better than doing the commands one by one. Maybe we should do this with all the other `docker build`.

docker/standalone/Dockerfile
```diff
@@ -1,8 +1,3 @@
-FROM gradle:8 as builder
```
MateoRojas, line 1: The jar is compiled on the pipeline at a previous step. This ensures two things:

1. Drastically reduces build time for linux/arm64 images. It was taking aroung 3 to 5 mins just to compile the jar
2. The jar can also be used by the `lh-server` image

docker/standalone/Dockerfile
```diff
@@ -26,18 +21,19 @@ COPY ./docker/standalone/dashboard-entrypoint.sh /lh
...
 COPY ./dashboard/apps/web/.next/static ./apps/web/.next/static
 COPY ./dashboard/apps/web/public ./apps/web/public
-COPY ./docker/dashboard/docker-entrypoint.sh ./
 
 WORKDIR /
 
-COPY --from=builder /lh/server/build/libs/server-*-all.jar /lh/server.jar
+COPY ./server/build/libs/server-*-all.jar /lh/server.jar
+
+ENV LHD_API_HOST=localhost
```
MateoRojas, line 36: This environment variables where missing for the dashboard to work inside `lh-standalone`

eduwercamacaro (approved):

docker/standalone/dashboard-entrypoint.sh
```diff
@@ -40,7 +40,7 @@ else
   export AUTH_SECRET=$(uuidgen)
 fi
 
-export HOSTNAME=localhost
+export HOSTNAME=0.0.0.0
```
MateoRojas, line 43: For some reason I don't seem to understand, `localhost` was not working and thus we were not able to stablish a connection to the dashboard on the `lh-standalone` image

.github/workflows/publish.yml
```diff
@@ -72,8 +72,12 @@ jobs:
           npm install pnpm --global
           pnpm install
           pnpm build
-      - name: Tests
-        run: ./gradlew server:test
+      - name: Build & Test Server
+        run: ./gradlew server:test server:shadowJar
```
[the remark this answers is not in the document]
MateoRojas: Oh great question, what do you think @sauljabin? 

TBH, I don't really think we should be running any test at this point. I might be wrong, but my understanding is that we should only reach this `publish` workflow once our other pipeline/workflows have successfully validated that everything is correct.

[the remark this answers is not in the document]
coltmcnealy-lh: I agree with that; I would even say don't put `server:test` here.

.github/workflows/publish.yml
coltmcnealy-lh: This seems like a good idea, I like it.

PR #616, opened 2024-01-19
chore: add TLS configs for keycloak

sauljabin:
@diablouma please test over this branch the dashboard, remember to point to https://localhost:8443

### DIFF
--- a/local-dev/setup.sh
+++ b/local-dev/setup.sh
@@ -38,12 +38,20 @@ services:
   keycloak:
     ports:
       - "8888:8888"
+      - "8443:8443"
     container_name: lh-server-auth
-    image: quay.io/keycloak/keycloak:21.1.1
-    command: ["start-dev", "--http-port=8888"]
+    image: quay.io/keycloak/keycloak:23.0
+    command:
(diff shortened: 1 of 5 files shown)


local-dev/setup.sh
```diff
@@ -171,14 +187,42 @@ EOF
...
+
+"$WORK_DIR/issue-certificates.sh"
+
+if [ ${keycloak} = true ]; then
     setup_keycloak
-    ;;
-*)
+fi
+
+if [ ${kafka} = true ]; then
```
coltmcnealy-lh, line 226: Please let's not do this, there is no use-case for setting up without kafka.

coltmcnealy-lh (approved):
Please revert the `--kafka` change. Everything else is fine.

sauljabin:
This change modify the setup script for configuring keycloak with TLS and that way the dashboard can work

@diablouma please test the dashbord pointing to both
http://localhost:8888 and https://localhost:8443

And let me know if the change is needed

sauljabin:
NOT NEEDED, master works well

Issue #614, opened 2024-01-18
Introduce Task Worker ID separate from OAuth Client ID


coltmcnealy-lh:
## Context

The `TaskWorkerHeartbeat` request is used to tell the server about the liveness of Task Workers. It's proto is as follows:
```
message TaskWorkerHeartBeatRequest {
  string client_id = 1;
  TaskDefId task_def_id = 2;
  string listener_name = 3;
}
```

Originally, the `client_id` was intended more as a `task_worker_id` which uniquely identifies the Task Worker in the `TaskWorkerGroup`. This allows the LH Server to intelligently balance work (in this case, work is defined as a connection to poll one specific LH Server). This originally came from the `LHC_CLIENT_ID` configuration **before** we introduced OAuth to LittleHorse.

_NOTE: when the `LHC_CLIENT_ID` is unset, the Task Worker SDK's auto-generate one for the purpose of having a unique identifier to aid in the task worker group balancing process._

When we introduced OAuth, we re-purposed the `LHC_CLIENT_ID` configuration to use it as the OAuth Client ID. Now, when a user uses OAuth, they must set that config to the client id. Usually, that config is shared among Task Workers in a group. This means that, to the server, it looks like the entire group of Task Workers is only one Task Worker, so all of the workers end up polling all of the servers (this is wasteful).

## Proposal

The OAuth client id (i.e. the `Principal` ID) and the Task Worker ID are two different things and should have two separate configurations. We should:
* introduce the `LHW_TASK_WORKER_ID` config which does the same thing that the `LHC_CLIENT_ID` config did _before_ introducing OAuth
* Send the `LHW_TASK_WORKER_ID` in the `PollTask` and `TaskWorkerHeartbeat` request
* Record the `task_worker_id` in the `TaskAttempt` proto.
* The Java Task Worker SDK logger should put the `task_worker_id` in the stdout logs using some fancy SLF4J thing.

Recording the `task_worker_id` in the `TaskAttempt` proto is important for debugging purposes: a good "best practice" would be to make the Task Worker ID be something like the Kubernetes Pod Name of the Task Worker, so that, for example, you can easily search for your logs in DataDog when you see a failed task run.

Pairing with @eduwercamacaro

coltmcnealy-lh:
Relates to issue #426

coltmcnealy-lh:
[...]

PR #618, opened 2024-01-22
Config Change: LHW_TASK_WORKER_ID

coltmcnealy-lh:
Implements ticket #614 and introduces the LHW_TASK_WORKER_ID config which replaces the confusing LHC_CLIENT_ID config.


sauljabin (approved):
it is just a rename variable

coltmcnealy-lh:
Yes, that's pretty much true haha. But it also renames a configuration, so it required a PR and a ticket.

@eduwercamacaro and I first thought that there was a bug because we thought the `LHC_CLIENT_ID` config was used for Task Worker ID **and** OAuth Client ID, but it turns out that was not the case.

PR #620, opened 2024-01-22
Use github packages to host docker images

mijailr:
This changes our public registry repository from AWS to github

Closes #615 

It already implements what was added on #615


.github/actions/publish-image/action.yml
```diff
@@ -0,0 +1,47 @@
+name: Build and publish docker images
```
sauljabin, line 1: Can we move all actions to a separated repository? and facilitate reuse them?

.github/workflows/publish.yml
```diff
@@ -4,10 +4,17 @@ on:
...
     tags:
       - "[0-9]+.[0-9]+.[0-9]+*" # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
-
+permissions:
+      packages: write
+      contents: read
 jobs:
-  publish-sdk-java:
+  publish-docker:
+    uses: ./.github/workflows/publish-docker.yaml
```
sauljabin, line 12: are we going to publish even if the tests fails?

sauljabin (approved):

.github/workflows/publish.yml
mijailr: It runs server tests inside of it

Issue #626, opened 2024-01-24
Investigate Running E2E Tests with Multiple LH Servers


coltmcnealy-lh:
We should figure out how to run our tests with multiple LH Servers running with multiple Stream THreads each. In the past, we have had bugs slip through our tests because they were only related to the interactive queries between servers, or were related to multiple Stream Thread's accessing shared objects in memory.

Considerations:
* Waiting for metadata to propagate
* Sufficient resources on GitHub Runner

coltmcnealy-lh:
I tried to do this along with #478 and it turns out that getting two servers to run in the same process is quite tricky. I think for now we can put this ticket "on the shelf". Before each release, we will make sure to run the tests manually with a multi-server cluster.

coltmcnealy-lh:
Done

coltmcnealy-lh:
Yeah this is done

Issue #576, opened 2024-01-02
Measure Metrics Aggregator performance.


eduwercamacaro:
### Context

Workflow metrics are calculated on a per-partition basis and then aggregated into a repartition store (see PartitionMetricsModel). To avoid sending many records into the repartition topic, we send the partition metrics every 30 seconds via Kafka Streams Punctuator. It is important to mention that punctuators execute blocking operations over the CoreProcessor, so it is mandatory to keep this as performant as possible to avoid hurting latency for commands. We need a way to verify that we meet performance expectations and weird behaviors under high-load scenarios.

We should ensure that any necessary changes do not impact metric calculations since they need to be accurate.

### Value

Low

### Effort

High

### Current State and Pains

We need to test LH server under high load scenarios and make sure that we meet performance expectations.

### Solution and Benefits

It will depend on the diagnosis.

### Acceptance Criteria

**Scenario** [the name for the behavior that will be described]
**Given** [the beginning state of the scenario]
**When** [specific action that the user makes]
**Then** [the outcome of the action in **When**]
**And** [used to continue any of three previous statements]

### Components

- [ ] Dashboard
- [X] Server
- [ ] Python SDK
- [ ] Go SDK
- [ ] Java SDK
- [ ] C# SDK
- [ ] LH Control
- [ ] LH Tests Utils

coltmcnealy-lh:
I did some benchmarking.

One terminal:
```
gradle example-hundred-tasks:run
```

Another terminal:
```
-> for i in $(seq 1 1000) ; do lhctl run hundred-tasks ; done
```

Wait for the last WfRun to finish.

With current `master` branch: 6 minutes 6 seconds.

Then I disabled the `MetricsUpdater#listen()` method and made it a no-op. 1 minute, 59 seconds.

We will rewrite this with the original proposed implementation.

coltmcnealy-lh:
This spike is closed; we are rewriting the Metrics Aggregator

PR #630, opened 2024-01-28
Expose More Performance-Related Configurations

coltmcnealy-lh:


coltmcnealy-lh:
Doing testing with the defaults.
* Core Stream Threads: 2
* Timer Stream Threads: 2
* Core Stream Commit Interval: 5 seconds
* Timer Stream Commit Interval: 30 seconds
* Core Memtable Size: 64MB
* Timer Memtable Size: 32MB
* Core State Store Cache Size: 32MB
* Timer State Store Cache Size: 64MB
* Shared RocksDB Block Cache Size: 64MB
* Write Buffer Manager Size: 64MB

```
12:54:22 -> for i in $(seq 1 1000); do     lhctl run hundred-tasks; done
```

It completed in 10 minutes 3 seconds. Yikes. Looked like a lot of RocksDB Write Stalls. Most likely caused by the Write Buffer Manager Size being too small.

coltmcnealy-lh:
* Core Stream Threads: 2
* Timer Stream Threads: 2
* Core Stream Commit Interval: 5 seconds
* Timer Stream Commit Interval: 30 seconds
* Core Memtable Size: 64MB
* Timer Memtable Size: 32MB
* Core State Store Cache Size: 32MB
* Timer State Store Cache Size: 64MB
* Shared RocksDB Block Cache Size: 64MB
* Write Buffer Manager Size: 4GB

This also took ~10 minutes. Turns out that this ticket is blocked by #576

coltmcnealy-lh:
I disabled metrics aggregation (`MetricsUpdater#listen()`) and then ran the benchmark again. Prior to the changes in this PR, the benchmark took 119 seconds. After the changes, running with `larger-memory.config` took 114 seconds.

It is a small improvement despite "maxing out" the available memory on the provided configs; however, the difference between `master` and this PR is that we now have a better way to cap the total memory (in-heap and off-heap) used by LittleHorse. This PR should unblock #626 .

I think this PR is ready to review on its own. We should define follow-up issues to:
* Enable `jemalloc` on our docker image
* Determine recommended settings for the various cache sizes based on a given amount of memory.

server/src/main/java/io/littlehorse/common/LHServerConfig.java
coltmcnealy-lh, line 1: We might want to open a tech-debt ticket to make this class smaller.

coltmcnealy-lh:
I disabled metrics aggregation (`MetricsUpdater#listen()`) and then ran the benchmark again. Prior to the changes in this PR, the benchmark took 119 seconds. After the changes, running with `larger-memory.config` took 113 seconds.

I disabled the changes related to the WriteBufferManager and Block Cache size (no shared WBM or Block Cache). Keeping the changes to commit interval and state store cache kept the small performance improvement (113 seconds). My thoughts going into this PR were confirmed:
* The changes to the `WriteBufferManager` and Block Cache are intended to reduce the risk of OOM errors.
* Using a larger commit interval should result in better performance. This performance change should be more drastic when running in a replicated environment with network latency (I am using local dev right now).

I think this PR is ready to review on its own. We should define follow-up issues to:
* Enable `jemalloc` on our docker image
* Determine recommended settings for the various cache sizes based on a given amount of memory.

My profiling has confirmed that this ticket accomplishes what it needed to do—power users can mostly limit the total on-heap and off-heap memory by tuning configurations, and we also improved performance slightly in the case where more resources are provided to the server.

The only danger of this PR is that the new default configurations result in poorer performance than the old default configurations, because for certain types of memory the old configurations did not specify a limit (which is dangerous as it makes OOM more easy to occur)

eduwercamacaro (approved):

server/src/main/java/io/littlehorse/common/util/RocksConfigSetter.java
```diff
@@ -1,39 +1,89 @@
...
 import org.rocksdb.CompactionStyle;
 import org.rocksdb.Options;
 
 @Slf4j
 public class RocksConfigSetter implements RocksDBConfigSetter {
 
+    // Need to inject the LHServerConfig, but Kafka Streams requires we pass in a Class with
+    // a default constructor. So the only way to do that is to have a static singleton which
+    // we set elsewhere in the code.
+    public static LHServerConfig serverConfig;
```
coltmcnealy-lh, line 18: @mjsax this appears to be the only way to "inject" configuration into the `RockDBConfigSetter` implementation. Would it be worth a KIP to add a method:

```
KafkaStreams#setRocksDBConfigurator(RocksDBConfigSetter otterSetter)
```
The method should throw `IllegalStateException` if it is called after `KafkaStreams#start()`, or if the `rocksdb.config.setter` parameter is set on the properties.

mjsax: Don't think it's the only way to inject configs.

There is `RocksConfigSetter.setConfig(..., config)` taking the Kafka Streams `config` object and third parameter. When you create `StreamsConfig` you pass in `Map<?, ?> props` and all entries will be contained in `config`.

Thus, you can take `LHServerConfig` and add all of it to `props` before creating `StreamsConfig` and should have access to all of it in `setConfig(...)`

coltmcnealy-lh: @mjsax Thanks so much! This works great. Much better than submitting a KIP to the Apache Kafka Debate Club :smile: 

CC @eduwercamacaro let's refactor this tomorrow.

PR #633, opened 2024-01-29
Fix standalone image

mijailr:
This PR fixes an issue with the standalone image that prevents the user to access the dashboard.

As an extra this also enables the platforms `linux/amd64` and `linux/arm641 and some minor linter issues.


coltmcnealy-lh (approved):
Looks good. Thanks @mijailr !

PR #635, opened 2024-01-29
Rebuild dashboard for standalone and dashboard docker images

mijailr:
While building the docker image, pnpm creates symbolic links to dependencies that cannot be recovered with a cache action through artifacts.

This disable the artifact action but instead builds the artifact for standalone and dashboard images.


coltmcnealy-lh (approved):
Looks good to me, but I can't say I fully understand all of this. I've not got much experience with anything JS-related

PR #637, opened 2024-01-29
LH Canary

sauljabin:
### Included:

- Metronome (synthetic monitoring simulator)
- Aggregator (kafka stream topology) for calculating metrics
- TaskRunLatency metric
- Configurations
- do-canary.sh script
- Log level improvements
- Protobuf classes
- Docker build

### Future:

- Finish DuplicatedTaskRun metric
- Kafka Stream tests
- gRPC server
- gRPC health check
- Prometheus integration


canary/canary.properties
```diff
@@ -0,0 +1,14 @@
+# Canary configurations
```
coltmcnealy-lh, line 1: This is interesting. Do we want to convert the LH Server to use this? It seems more "java" than what we do with the server. I think it would be best to use one standard for everything...

On the other hand, the dashboard doesn't do this, and this doesn't work well for env vars.

canary/src/main/java/io/littlehorse/canary/Main.java
```diff
@@ -0,0 +1,46 @@
...
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class Main {
+
+    public static void main(String[] args) throws IOException, InterruptedException {
+        CanaryConfig config = args.length > 0 ? ConfigLoader.load(Paths.get(args[0])) : ConfigLoader.load();
```
coltmcnealy-lh, line 19: Do we support environment variables?

canary/src/main/java/io/littlehorse/canary/config/CanaryConfig.java
```diff
@@ -0,0 +1,48 @@
+package io.littlehorse.canary.config;
+
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.stream.Collectors;
+
+public class CanaryConfig implements Config {
+
+    public static final String LH_CANARY_PREFIX = "lh.canary.";
```
coltmcnealy-lh, line 9: Ah, I see, `LHC_` conflicts with LH Client configurations. That is interesting. However, I don't know if I've ever seen environment variables with `.` in them...

I'm not sure how critical it is for the canary to support environment variable configs. The Server does, and we use that in the operator. But our operator could easily make a Properties file for the canary to run as well. What do you think?

canary/src/main/java/io/littlehorse/canary/config/ConfigLoader.java
```diff
@@ -0,0 +1,81 @@
...
+import org.eclipse.microprofile.config.Config;
+
+public class ConfigLoader {
+
+    private ConfigLoader() {}
+
+    /**
+     * Loads properties in this order:
+     * 1. MicroProfile Config configuration file META-INF/microprofile-config.properties in the classpath
+     * 2. Environment variables
```
coltmcnealy-lh, line 21: Is `LITTLEHORSE_CANARY_FOO` translated to `littlehorse.canary.foo`? If so, we should migrate the LH Server to use this as well. That is really cool.

coltmcnealy-lh (approved):
There's some interesting things going on in here. I like it. If we want to keep PR's small, we can merge this, but I would maybe prefer to wait until we have an end-to-end working one with the aggregator + metronome, and tutorial on how to do it in the README.

canary/src/main/java/io/littlehorse/canary/aggregator/serdes/MetricAverageDeserializer.java
```diff
@@ -0,0 +1,24 @@
+package io.littlehorse.canary.aggregator.serdes;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import io.littlehorse.canary.aggregator.internal.ProtobufDeserializationException;
+import io.littlehorse.canary.proto.MetricAverage;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.kafka.common.serialization.Deserializer;
+
+@Slf4j
+public class MetricAverageDeserializer implements Deserializer<MetricAverage> {
```
coltmcnealy-lh, line 10: Interesting to note that you decided not to use `LHSerializable` here. Seems you're using the proto object just to store data, which is actually what it is designed to do.

If I were to re-write the LH Server, I would probably use your pattern.

canary/src/main/java/io/littlehorse/canary/aggregator/topology/TaskRunLatencyTopology.java
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.kafka.streams.kstream.TimeWindows;
+
+@Slf4j
+public class TaskRunLatencyTopology {
+
+    public TaskRunLatencyTopology(final KStream<String, Metric> metricStream) {
+        metricStream
+                .filter((key, value) -> value.hasTaskRunLatency())
+                .groupByKey()
+                .windowedBy(TimeWindows.ofSizeAndGrace(Duration.ofMinutes(1), Duration.ofSeconds(30)))
```
coltmcnealy-lh, line 22: Question here—I haven't done a full deep dive into the code. How do we want to handle grace periods? I am wondering about the specific case when:
* A `TaskRun` is scheduled.
* Before the Task Worker starts the task, the server crashes and goes through a super-long state restoration
* One hour later, the `TaskRun` is executed.

Can we support this case?

canary/src/main/java/io/littlehorse/canary/aggregator/topology/TaskRunLatencyTopology.java
```diff
@@ -0,0 +1,44 @@
...
+                .map((key, value) -> KeyValue.pair(key.key(), value))
+                .peek((key, value) -> log.info(
+                        "server={}, count={}, sum={}, avg={}", key, value.getCount(), value.getSum(), value.getAvg()));
+    }
+
+    private static MetricAverage aggregate(final Metric value, final MetricAverage aggregate) {
+        final long count = aggregate.getCount() + 1;
+        final double sum = aggregate.getSum() + value.getTaskRunLatency().getLatency();
+        final double avg = sum / count;
+        return MetricAverage.newBuilder()
```
coltmcnealy-lh, line 38: Looks like what happens is the latency will be reported as "one hour", but it will show up once the TaskRun is completed (i.e. once the incident is resolved.)

That may be fine for the first iteration, but in the future we should figure out if there is another metric (i.e. the duration of pending workflows that haven't been completed yet) so we can do alerting _before_ the incident self-resolves.

canary/src/main/java/io/littlehorse/canary/config/LittleHorseConfig.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @Override
+    public Map<String, Object> toMap() {
+        return configs;
+    }
+
+    @Override
+    public String toString() {
+        return configs.toString();
```
coltmcnealy-lh, line 35: We need to be careful with sanitizing secret values. But that can be a future GH issue.

canary/src/main/java/io/littlehorse/canary/kafka/MetricsEmitter.java
```diff
@@ -0,0 +1,52 @@
...
+        return producer.send(record, (metadata, exception) -> {
+            if (exception == null) {
+                log.debug("Emitting message {} {}", metric.getMetricCase(), key);
+            } else {
+                log.error("Emitting message {} {}", metric.getMetricCase(), key, exception);
+            }
+        });
+    }
+
+    public RecordMetadata emit(final String key, final Metric metric) {
```
coltmcnealy-lh, line 40: Could you put in the Javadoc on this method that it is blocking? Additionally, we may want to be careful about _where_ we call this method. For example, probably we shoudn't call it from within the Aggregator Streams Topology.

However, if we call it from the Metronome Task Worker, that is fine (in fact, we _need_ to make it blocking for that use-case).

canary/src/main/java/io/littlehorse/canary/metronome/MetronomeTask.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    private void emitDuplicatedTaskRunMetric(final WorkerContext context) {
+        final String key = String.format("%s/%s", context.getIdempotencyKey(), context.getAttemptNumber());
+
+        final Metric metric = getMetricBuilder()
+                .setDuplicatedTaskRun(DuplicatedTaskRun.newBuilder().setUniqueTaskScheduleId(key))
+                .build();
+
+        emitter.emit(key, metric);
```
coltmcnealy-lh, line 65: Very nice job with using the `future()` and the `emit()` here to control the flushing of the producer. Let's make sure that we have configured `enable.idempotence=true` on the producer so that we don't get false positives.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/Workflow.java
```diff
@@ -239,7 +239,7 @@ public boolean doesWfSpecExist(LittleHorseBlockingStub client, Integer majorVers
      * @param client is an LHClient.
      */
     public void registerWfSpec(LittleHorseBlockingStub client) {
-        log.info("Creating wfSpec:\n {}", LHLibUtil.protoToJson(client.putWfSpec(compileWorkflow())));
+        log.debug("Creating wfSpec:\n {}", LHLibUtil.protoToJson(client.putWfSpec(compileWorkflow())));
```
coltmcnealy-lh, line 242: This should probably stay in the `INFO` level.

server/src/main/java/io/littlehorse/server/streams/BackendInternalComms.java
```diff
@@ -19,7 +19,6 @@
 import io.littlehorse.common.LHServerConfig;
 import io.littlehorse.common.Storeable;
 import io.littlehorse.common.exceptions.LHApiException;
-import io.littlehorse.common.exceptions.LHBadRequestError;
```
coltmcnealy-lh, line 22: This class should be removed from the code I think. We should have deleted it when we did the GRPC facelift (the horrible dark days when we had our own error code schema :joy: )

coltmcnealy-lh (approved):
Looks very good. Only change request is to keep the logging of the `PutWfSpec` and `PutTaskDef` at the `INFO` level. They don't happen very often but they are quite useful.

PR #641, opened 2024-01-30
fix: python grpc configurations

sauljabin:
Infinite max_pings_without_data


coltmcnealy-lh (approved):
Thanks Saúl! Good find.

PR #643, opened 2024-01-30
fixes issue with rpc searchWfRun

coltmcnealy-lh:
Several fixes:

* In our scramble to release `0.7.1`, we broke `rpc SearchWfRun` when the client sets the `status` field. Fixed this.
* Adds e2e tests for various configurations of `rpc SearchWfRun` and also a basic `rpc SearchVariable` test
* Adds e2e tests for ThreadRun Retention and WfRunRetention.
* Fixes bugs with implementation of ThreadRun Retention.


server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/WfRunModel.java
```diff
@@ -68,6 +68,7 @@ public class WfRunModel extends CoreGetable<WfRun> {
     private WfRunIdModel id;
     private WfSpecIdModel wfSpecId;
     private List<WfSpecIdModel> oldWfSpecVersions = new ArrayList<>();
+    private int greatestThreadRunNumber;
```
coltmcnealy-lh, line 71: Somehow, this was added to the proto when I initially implemented ThreadRun retention, but I forgot to set it here.

server/src/main/java/io/littlehorse/server/streams/BackendInternalComms.java
```diff
@@ -902,14 +902,14 @@ private Pair<List<ByteString>, PartitionBookmarkPb> onePartitionPaginatedTagScan
         if (bookmark != null) {
             startKey = bookmark.getLastKey();
         } else {
-            startKey = tagPrefixScan.getKeyPrefix();
+            startKey = tagPrefixScan.getKeyPrefix() + "/";
```
coltmcnealy-lh, line 905: This is where we want it to be fixed.

server/src/test/java/e2e/RetentionTest.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    @Test
+    void childThreadShouldGcAfterOneSecond() {
+        verifier.prepareRun(threadRetentionWorkflow)
+                .thenVerifyWfRun(wfRun -> {
+                    Assertions.assertThat(wfRun.getThreadRunsCount()).isEqualTo(2);
+                })
+                .thenSendExternalEventJsonContent(RETENTION_CHILD_EVENT, Map.of())
+                .thenSleep(1000)
```
eduwercamacaro, line 100: This seems like the old e2e framework. 

I don't really recommend adding this as an option to develop e2e in our context. Instead, we should wait for the server to handle the external events, using GRPC calls to query the external events.

coltmcnealy-lh:
I was waiting for the Threadrun retention policy to expire. This one is an
advanced use case. The Threadrun won’t be garbage collected until a Command
comes, which is why I have an external event.

I can explain over zoom. I don’t think there is another good way.

Colt McNealy

*Founder, LittleHorse.dev*

PR #645, opened 2024-01-30
Small fixes:

coltmcnealy-lh:
- Ensure that threadruns with retention policy of zero seconds are immediately deleted
- Fixes the LHLibUtil.wfRunIdToString() method.


server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/thread/ThreadRetentionPolicyModel.java
```diff
@@ -69,7 +69,7 @@ public boolean shouldGcThreadRun(ThreadRunModel thread) {
         }
         Date threadEndTime = thread.getEndTime();
         if (threadEndTime != null && getScheduledTerminationFor(threadEndTime) != null) {
-            return new Date().after(getScheduledTerminationFor(threadEndTime));
+            return !(new Date().before(getScheduledTerminationFor(threadEndTime)));
```
coltmcnealy-lh, line 72: When a ThreadRun has a GC Policy of `0 seconds` then we need to delete it as soon as it's created. That often happens in the same millisecond...so we need to a "greater than equals" not "greater than"

sauljabin (approved):
For this kind of bug a recommend a Unit test.

coltmcnealy-lh:
@sauljabin added unit test :smile:

PR #649, opened 2024-01-31
Uses Proper `TenantId` rather than `string tenant_id = 1`;

coltmcnealy-lh:
This is necessary to future-proof our API in case we evolve the way Tenants are stored. This would be a breaking change for any deployment using principals and tenants. However, we do not know of any users using principals in production as of today.

Therefore, it is critical to get this merged before `1.0.0`.


coltmcnealy-lh:
@eduwercamacaro I validated that this works with no principals configured (also with no OAuth configured) and with the default tenant. Could you please validate it with principals enabled? Thanks!

test-utils/src/main/java/io/littlehorse/test/LHExtension.java
```diff
@@ -70,4 +82,46 @@ public void postProcessTestInstance(Object testInstance, ExtensionContext contex
...
+                                .build())
+                        .build();
+                testContext
+                        .getAnonymousClient()
+                        .putPrincipal(PutPrincipalRequest.newBuilder()
+                                .setId(principalId.get())
+                                .setGlobalAcls(acls)
+                                .putPerTenantAcls(testContext.getConfig().getTenantId(), acls)
+                                .build());
+                // wait until the server is up
```
eduwercamacaro, line 108: wrong message. I'll change it in the next commit

test-utils/src/main/java/io/littlehorse/test/LHExtension.java
```diff
@@ -70,4 +82,46 @@ public void postProcessTestInstance(Object testInstance, ExtensionContext contex
...
+            } catch (StatusRuntimeException ex) {
+                if (!ex.getStatus().getCode().equals(Status.Code.ALREADY_EXISTS)) {
+                    throw ex;
+                }
+            }
+        }
+    }
+
+    private Optional<String> getPrincipalId() {
+        return Optional.ofNullable(System.getenv().getOrDefault("LH_CLIENT_ID", "my-principal"));
```
eduwercamacaro, line 125: There is no point on returning a Optional here. I'll change this

coltmcnealy-lh: Why not `Optional<Optional<Optional<String>>>`?

But anyways, I would prefer to see it return `PrincipalIdModel` though.

coltmcnealy-lh (commented):
Looks good. Thanks Eduwer!

eduwercamacaro (approved):

PR #650, opened 2024-01-31
Support principal ids for e2e

eduwercamacaro:


server/src/main/java/io/littlehorse/server/streams/topology/core/processors/CommandProcessor.java
```diff
@@ -140,7 +140,12 @@ private ProcessorExecutionContext buildExecutionContext(Record<String, Command>
     }
 
     public void onPartitionClaimed() {
+        if (partitionIsClaimed) {
+            throw new RuntimeException("Re-claiming partition! Yikes!");
```
coltmcnealy-lh, line 144: This looks like I wrote this code :joy: You are picking up my style!

server/src/main/java/io/littlehorse/server/streams/topology/core/processors/CommandProcessor.java
```diff
@@ -155,11 +160,6 @@ public void onPartitionClaimed() {
     private void rehydrateTenant(TenantModel tenant) {
         TenantScopedStore coreDefaultStore =
                 TenantScopedStore.newInstance(this.nativeStore, tenant.getId(), new BackgroundContext());
-        if (partitionIsClaimed) {
-            throw new RuntimeException("Re-claiming partition! Yikes!");
```
coltmcnealy-lh, line 159: Oh never mind, I think you just copied it from me haha

server/src/main/java/io/littlehorse/server/streams/topology/core/processors/CommandProcessor.java
eduwercamacaro: That's your code. just moving

coltmcnealy-lh (approved):
Looks good. But what about checking to make sure we don't doubly-hydrate the `default` tenant if it exists in the store?

eduwercamacaro:
I need to a add to prevent double task rehydration for default tenant.

PR #651, opened 2024-02-04
FIX: NPE when accessing variable outside of scope.

coltmcnealy-lh:
When a ThreadSpec attempts to mutate a real variable (but that variable is out of scope for that threadspec) the server threw an NPE and ignored the Command that triggered the advance that resulted in reading that variable.

The proper behavior is for the illegal access to throw an LHVarSubError.

This commit adds a failing e2e test for that case and fixes it. However, the e2e test has apparently discovered another bug related to moving a failed thread from HALTED to ERROR, which will be addressed in subsequent commits.

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/VariableMutationModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/VariableMutationModel.java
@@ -3,6 +3,7 @@ package io.littlehorse.common.model.getable.global.wfspec.variable;
 import com.google.protobuf.Message;
 import io.littlehorse.common.LHSerializable;
 import io.littlehorse.common.exceptions.LHVarSubError;
+import io.littlehorse.common.model.getable.core.variable.VariableModel;
 import io.littlehorse.common.model.getable.core.variable.VariableValueModel;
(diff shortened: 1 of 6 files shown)


coltmcnealy-lh:
Looks like everything should be good to go as long as I can fix the build issues + flakiness on the legacy e2e framework...

eduwercamacaro (approved):

server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/VariableMutationModel.java
```diff
@@ -90,11 +91,19 @@ public VariableValueModel getLhsValue(ThreadRunModel thread, Map<String, Variabl
 
     private VariableValueModel getVarValFromThreadInTxn(
             String varName, ThreadRunModel thread, Map<String, VariableValueModel> txnCache) throws LHVarSubError {
-        VariableValueModel lhsVar = txnCache.get(this.lhsName);
-        if (lhsVar == null) {
-            lhsVar = thread.getVariable(this.lhsName).getValue();
+        VariableValueModel result = txnCache.get(this.lhsName);
```
coltmcnealy-lh, line 94: Here's where I fixed the bug. Everything else is just logs and tests

PR #654, opened 2024-02-05
Receives tag configuration as build and push action input

MateoRojas:
Tag overriding is needed on the build and push action to be able to set the sha as the tag for our particular use case

### DIFF
--- a/.github/actions/publish-image/action.yml
+++ b/.github/actions/publish-image/action.yml
@@ -19,6 +19,9 @@ inputs:
   prefix:
     description: prefix to be appended to image tag ex brach-
     default: ""
+  tags:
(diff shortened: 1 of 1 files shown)


coltmcnealy-lh:
@MateoRojas let's chat offline about what the use-case is here (offline since I suspect it's related to one of our internal projects)

mijailr (approved):
@coltmcnealy-lh @MateoRojas this custom action will be moved eventually to it's own repository as an internal action. This change makes sense to cover scenarios where the release cycle is not linked to git tags. I'll approve this PR to unlock Mateo without asking him to fork the action. Also I'll increase the priority to move this action into it's own repository as soon as possible.

coltmcnealy-lh:
This will be moved out of the OSS repo since it's not related to LH OSS.

PR #660, opened 2024-02-08
feat(canary): export latency metrics to prometheus

sauljabin:
**Includes**

- Add prometheus exporter (Interactive Query)
- New protobuf classes
- Using semantic keys
- Add common tags for prometheus
- Metric cache: Micrometer (the framework we are using to parse our metrics to prometheus, in other words the presentation layer) generates gauges, each one represents one metric, we have to keep the pointer because that way we can update the value. We have to clean this cache in case of a rebalance, because the partitions assignments will change and some metrics will change of instance.   

**Future:**

- Clean cache
- Duplicated scheduled task run metric
- Improve code (clean code)


coltmcnealy-lh (approved):
Excited to see this in production.

canary/src/main/java/io/littlehorse/canary/kafka/KafkaTopicBootstrap.java
```diff
@@ -2,32 +2,35 @@
...
+
+        adminClient = KafkaAdminClient.create(config.toKafkaAdminConfig().toMap());
+        Shutdown.addShutdownHook("Topics Creator", adminClient);
 
         try {
-            final NewTopic canaryTopic = new NewTopic(metricsTopicName, topicPartitions, topicReplicas);
+            final NewTopic canaryTopic =
+                    new NewTopic(config.getTopicName(), config.getTopicPartitions(), config.getTopicReplicas());
 
             adminClient.createTopics(List.of(canaryTopic)).all().get();
```
eduwercamacaro, line 32: Suggestion:

Maybe it could be a good idea to add a timeout here

sauljabin: done: `adminClient.createTopics(List.of(canaryTopic)).all().get(1, TimeUnit.SECONDS);`

eduwercamacaro (approved):

PR #662, opened 2024-02-10
Add Per-Thread Failure Handlers on WAIT_FOR_THREADS Node (#657)

### LINKED ISSUE #657: Add Per-Thread Failure Handlers on `WAIT_FOR_THREADS` Node
Right now you can put a Failure Handler on a `WaitForThreadsNode`; however, the failure that is handled is thrown by the `NodeRun` itself. THis means that the resulting `FailureHandler` `ThreadRun` is always a child of the parent `ThreadRun`, and thus does not have access to any of the variables of the child.

We are adding a new API to the `WaitForThreadsNodeOutput`: 
```
    WaitForThreadsNodeOutput handleExceptionOnChild(String exceptionName, WorkflowThread handler);

    WaitForThreadsNodeOutput handleErrorOnChild(LHErrorType error, WorkflowThread handler);

    WaitForThreadsNodeOutput handleAnyFailureOnChild(WorkflowThread handler);
```

When any of the child threads that are being waited for fail, then there will be a `FailureHandler` `ThreadRun` that is a _child of the child_. The success of the `WaitingThread` is determined by the success of the `FailureHandler` (if one is set).

This does not replace the current behavior; rather, it augments the current behavior and gives the users an additional tool to handle failed child threads.

coltmcnealy-lh:
Fixes #657


examples/basic/src/main/java/io/littlehorse/examples/BasicExample.java
```diff
@@ -20,16 +21,19 @@
...
-            "example-basic",
-            wf -> {
-                WfRunVariable theName = wf.addVariable(
-                    "input-name",
-                    VariableType.STR
-                );
-                wf.execute("greet", theName);
-            }
-        );
+        return new WorkflowImpl("handle-exception-on-children", wf -> {
```
eduwercamacaro, line 24: Please revert these changes or create a new example.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
```diff
@@ -519,6 +519,21 @@ public void addTimeoutToExtEvt(NodeOutputImpl node, int timeoutSeconds) {
         spec.putNodes(node.nodeName, n.build());
     }
 
+    public void addFailureHandlerOnWaitForThreadsNode(WaitForThreadsNodeOutputImpl node, FailureHandlerDef handler) {
+        checkIfIsActive();
+        Node.Builder n = spec.getNodesOrThrow(node.nodeName).toBuilder();
+
+        if (n.getNodeCase() != NodeCase.WAIT_FOR_THREADS) {
+            throw new IllegalStateException("orzdash this should only be a WAIT_FOR_THREADS node");
```
eduwercamacaro, line 527: Suggestion:
`throw new IllegalStateException("Only WAIT_FOR_THREADS nodes can have child exception handlers.");`

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WaitForThreadsNodeOutputImpl.java
```diff
@@ -15,13 +15,45 @@ public WaitForThreadsNodeOutputImpl(String nodeName, WorkflowThreadImpl parent,
...
     }
 
-    public WaitForThreadsNodeOutputImpl withPolicy(WaitForThreadsPolicy failureStrategy) {
-        Node nodesOrThrow = threadSpec.getNodesOrThrow(nodeName);
-        WaitForThreadsNode waitForThreads = nodesOrThrow.getWaitForThreads();
-        waitForThreads = waitForThreads.toBuilder().setPolicy(failureStrategy).build();
-        nodesOrThrow =
-                nodesOrThrow.toBuilder().setWaitForThreads(waitForThreads).build();
-        threadSpec.putNodes(nodeName, nodesOrThrow);
+    public WaitForThreadsNodeOutput handleExceptionOnChild(String exceptionName, ThreadFunc handler) {
```
eduwercamacaro, line 18: In my opinion, we should not allow the exception name Null here because we set FailureHandlerDef.LHFailureType.FAILURE_TYPE_ERROR when the exception name is null.

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/WfRunModel.java
```diff
@@ -399,15 +399,13 @@ private boolean startXnHandlers(Date time) {
             }
             somethingChanged = true;
             pendingFailures.remove(i);
-            Map<String, VariableValueModel> vars;
+            Map<String, VariableValueModel> vars = new HashMap<>();
+            ;
```
eduwercamacaro, line 403: remove this

eduwercamacaro (approved):

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
coltmcnealy-lh: It should be impossible for a user to make this happen; otherwise, it means that the code I wrote doesn't behave how I thought it does. I will make a note of that.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WaitForThreadsNodeOutputImpl.java
coltmcnealy-lh: Good catch, I am working on unit tests and hopefully would have caught this. The correct answer is that it should be `FAILURE_TYPE_EXCEPTION` if `exceptionName == null`, not `FAILURE_TYPE_ERROR`.

server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/RunWfRequestModel.java
```diff
@@ -133,6 +135,14 @@ public WfRun process(ProcessorExecutionContext processorContext, LHServerConfig
             }
         }
 
+        // Validate input variables before saving anything.
+        ThreadSpecModel entrypointThread = spec.getThreadSpecs().get(spec.getEntrypointThreadName());
```
eduwercamacaro, line 139: Suggestion:

`ThreadSpecModel entrypointThread = spec.getEntrypointThread();`

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
```diff
@@ -213,6 +215,75 @@ public NodeRunModel getCurrentNodeRun() {
...
+            return;
+        }
+
+        for (ThreadVarDefModel threadVarDef : threadSpec.getVariableDefs()) {
+            VariableDefModel varDef = threadVarDef.getVarDef();
+            String varName = varDef.getName();
+            VariableValueModel val;
+
+            if (threadVarDef.getAccessLevel() == WfRunVariableAccessLevel.INHERITED_VAR) {
+                if (variables.containsKey(varName)) {
```
eduwercamacaro, line 261: This is confusing. This is the same as line 269.

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/WaitForThreadsRunModel.java
```diff
@@ -11,27 +11,39 @@
...
 
 @Getter
 @Setter
+@Slf4j
 public class WaitForThreadsRunModel extends SubNodeRun<WaitForThreadsRun> {
 
-    private List<WaitForThreadModel> threads;
-    private WaitForThreadsPolicy policy;
+    private List<WaitForThreadModel> threads = new ArrayList<>();
+    ;
```
eduwercamacaro, line 40: remove this

eduwercamacaro (approved):

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
coltmcnealy-lh: That's because this is handling inherited variables. For an inherited variable, the client who starts the `ThreadRun` shouldn't pass in a value for the inherited variable. That's because we use the shared/inherited one from the parent.

The reason why there is a `continue;` statement on line 266 is because we don't want to save a `Variable` to the rocksdb store. I will just add the validation in the `ThreadSpecModel#validateStartVariables()`

PR #663, opened 2024-02-11
Disabling wfmetrics

eduwercamacaro:


coltmcnealy-lh (approved):
Approved, we will add this back in a more performant manner next week

Issue #503, opened 2023-12-19
Clean Up NodeRun Lifecycle


diablouma:
{{NodeRunModel#advanceIfPossible()}} → Look at {{NodeRunModel#isCompleteOrRecoveredFromFailure()}}

Currently, we check for status {{EXCEPTION}} {{ERROR}} or {{HALTED}}. In the past, we only checked for {{ERROR}} (not even {{EXCEPTION}}).

A recent change (previous to PR 119, I believe it was one of the Wait For Threads PR’s) we made failures cause the {{ThreadRun}} to HALT, so the PR number 119 discovered a bug where the failure handlers for {{TASK}} node’s weren’t really working.

Now the {{if()}} statement checks for all three statuses ({{EXCEPTION}}, {{ERROR}}, {{HALTED}}), but I think that we only need to check for {{HALTED}}.

However, I am also concerned whether it is safe to check for the {{HALTED}} status, because {{HALTED}} means more things than “there was a failure handler”.

It might still be safe because in the {{ThreadRunModel#advance()}}, the ThreadRun itself still checks to ensure that its own status is {{RUNNING}} before calling the {{NodeRunModel#advanceIfPossible()}}

coltmcnealy-lh:
@eduwercamacaro You and I should take a look at this in pair programming. This seems more tech debt than feature/bug.

coltmcnealy-lh:
I tried to make a stylistic change to the `NodeRunModel#isCompleteOrRecoveredFromFailure()` to make it clearer in my first commit to #665. That change broke the e2e tests. Upon further discussion, we might actually want to refactor the `message NodeRun` protobuf. I am proposing:

1. Remove the `repeated int32 failure_handler_ids` field
2. Add `optional int32 failure_handler_id` to the `Failure` protobuf.

This would be a **BREAKING CHANGE** for currently active `WfRun`'s. However, we have confirmed with our early users that they are not using features that would be affected by this change. Additionally, we are still pre-1.0.0, so we have permission to make this change.

coltmcnealy-lh:
It appears that the bugs I am finding while adding tests on this PR #665 are actually caused by ticket #656. I will finish both together in this PR.

coltmcnealy-lh:
Closed by #665

PR #665, opened 2024-02-13
minor: cleans up `NodeRunModel#isCompleteOrRecoveredFromFailure()`

coltmcnealy-lh:
Resolves ticket #503. My investigation reveals there is no bug or malicious behavior; however, we have a cleaner way to implement the isCompleteOrRecoveredFromFailure() method.


coltmcnealy-lh:
Well, that was interesting...the fact that this change breaks some things means that something else is broken. My hypothesis is that the `was_properly_handled` field on the `Failure` is not being set.

server/src/main/java/io/littlehorse/common/model/getable/core/noderun/NodeRunModel.java
```diff
@@ -328,29 +328,29 @@ public NodeRun.Builder toProto() {
     }
 
     public boolean isInProgress() {
-        return (status != LHStatus.COMPLETED && status != LHStatus.HALTED && status != LHStatus.ERROR);
```
coltmcnealy-lh, line 331: We forgot to check for `EXCEPTION` here, and that caused some weird behavior.

server/src/main/java/io/littlehorse/common/model/getable/core/noderun/NodeRunModel.java
```diff
@@ -328,29 +328,29 @@ public NodeRun.Builder toProto() {
...
+        }
+        return false;
     }
 
     public boolean isCompletedOrRecoveredFromFailure() {
         if (status == LHStatus.COMPLETED) {
             return true;
         }
 
-        if (status == LHStatus.ERROR || status == LHStatus.HALTED || status == LHStatus.EXCEPTION) {
```
coltmcnealy-lh, line 339: This super-complex logic is no longer needed now that we fixed the `EXCEPTION` issue above.

eduwercamacaro (approved):

coltmcnealy-lh:
I'm going to YOLO merge this and we can address the small comments in follow-up PR's

Issue #606, opened 2024-01-16
Support Backoff for TaskRun Retries


coltmcnealy-lh:
Currently, `TaskRun` retries do not have any backoff. They just immediately re-schedule the `TaskRun` synchronously. This is not good because if the failure is caused by something _upstream_ of the Task Worker (i.e. flaky API), the retries will quickly be exhausted.

We should allow users to set a configurable retry backoff policy, probably starting with an exponential backoff policy.

### Proto

We should add it to the `TaskNode` proto. There should be:

```
message ExponentialBackoff {
  initial_backoff_seconds = 1;
  multiplier = 2;
  maximum_backoff_seconds = 3;
}

message TaskNode {
 // ...
  oneof retry_policy {
    ExponentialBackoff exponential_backoff = ?;
  }
}
```

This will allow us to easily add more retry policies in the future. If we have `RETRYPOLICY_NOT_SET` then we treat it as "retry immediately" which is the current behavior.

### Implementation

Currently, retries don't go to the Timer Topology. In this case, retries _must_ go through the timer topology. We need to introduce a new `Command` for this.

### Discussion
There are a few questions:
* Do we need to update the `status` field of the `TaskRun` to support the case when we are waiting for the next retry?
* When does the `TaskAttempt` get created? It may be useful to have some info on the `TaskRun` itself about when the next retry is scheduled.

### Workflow SDK's

The Workflow SDK's should make it easy to set a retry policy at the `Workflow`, `WorkflowThread`, and `NodeOutput` level.

coltmcnealy-lh:
I will add a new value to the `TaskStatus` enum: `TASK_PENDING`. This enum will be used to represent a `TaskAttempt` that is not yet `TASK_SCHEDULED` (because it is waiting for the retry policy to mature before being put on the Task Queue).

We should add as a follow-up ticket that a `TaskNodeRun` whose `TaskRun` is inbetween `TaskAttempt`s (i.e. it is `TASK_PENDING`) can be interrupted. Currently, a `TaskNodeRun` is never `interruptable()`. This means that when trying to make a `ThreadRun` move from `HALTING` to `HALTED`, we always wait for the `TaskRun` to complete or fail.

[...]

PR #666, opened 2024-02-13
Add Exponential Backoff Retry Policy to TaskNode

coltmcnealy-lh:
Implements #606


coltmcnealy-lh:
Closing this PR as it is numbered 666, which is not cool with me.

PR #667, opened 2024-02-16
601 fix usertask trigger lifecycle hooks

eduwercamacaro:


schemas/user_tasks.proto
```diff
@@ -104,6 +104,7 @@ message UserTaskRun {
 
   // The NodeRun with which the UserTaskRun is associated.
   NodeRunId node_run_id = 11;
+  int64 epoch = 12;
```
coltmcnealy-lh, line 107: Please add a comment for this since it is part of the public api.

```
// Current observed epoch of the UserTaskRun, related to the number of times it has been
// updated or re-assigned. Used internally to implement automated reassignment and reminder
// tasks.
```

coltmcnealy-lh: Additionally, the `epoch` can be an `int32` since it will only be re-assigned at most 10-20 times, not 1M+.

server/src/main/java/io/littlehorse/common/model/getable/core/usertaskrun/UserTaskRunModel.java
```diff
@@ -99,6 +100,7 @@ public UserTaskRunModel(
         this.nodeRunId = nodeRunModel.getObjectId();
         this.id = new UserTaskRunIdModel(nodeRunId.getWfRunId());
         this.scheduledTime = new Date();
+        this.epoch = new Date().getTime();
```
coltmcnealy-lh, line 103: Ah, let's please just start with `0` and increment it each time something changes.

server/src/test/java/e2e/UserTaskTest.java
```diff
@@ -102,6 +124,20 @@ public Workflow buildDeadlineReassignmentWorkflowUserWithoutGroup() {
...
+    public Workflow buildDeadlineReassignmentWithEpoch() {
+        return new WorkflowImpl("deadline-reassignment-with-epoch", entrypointThread -> {
+            entrypointThread.spawnThread(
+                    utThread -> {
+                        UserTaskOutput formOutput = utThread.assignUserTask(USER_TASK_DEF_NAME, "yoda", "my-group");
+                        utThread.releaseToGroupOnDeadline(formOutput, 20);
+                    },
+                    "user-task-thread",
+                    Map.of());
+            entrypointThread.spawnThread(sleepThread -> sleepThread.sleepSeconds(6), "sleep-thread", Map.of());
```
coltmcnealy-lh, line 137: Can we use `1` second here, and only `3` seconds on line 133? Otherwise it will make the pipeline much slower.

server/src/test/java/e2e/UserTaskTest.java
```diff
@@ -71,6 +77,22 @@ void shouldTransferOwnershipFromUserToSpecificGroupOnDeadline() {
...
 
+    @Test
+    void shouldTestDeadlineReassignmentWithEpoch() {
+        WfRunVerifier wfRunVerifier = workflowVerifier
+                .prepareRun(deadlineReassignmentWithEpoch)
+                .waitForStatus(RUNNING)
+                .waitForUserTaskRunStatus(1, 1, UserTaskRunStatus.ASSIGNED)
+                .waitForNodeRunStatus(2, 1, COMPLETED, Duration.ofSeconds(8))
+                .thenAssignUserTask(1, 1, true, "anakin", null)
+                .waitForUserTaskRunStatus(1, 1, UserTaskRunStatus.UNASSIGNED, Duration.ofSeconds(20));
```
coltmcnealy-lh, line 88: Don't think we need to wait `20` seconds here, see my comment below.

coltmcnealy-lh (approved):
Overall this is very good work. Just a few suggestions:
- use an `int32` that we increment for the `epoch` rather than a `Date`. We should only increment the `epoch` when the task is re-assigned or completed.
- Add the ability to choose between `ON_ASSIGNED` and `ON_ARRIVAL` for the `scheduleReminderTask()` method.
- See if we can make the e2e test a bit faster :smile:

server/src/test/java/e2e/UserTaskTest.java
eduwercamacaro: Test becomes flaky if I add a short sleep time here. I'm not sure if this test is worth it.

server/src/test/java/e2e/UserTaskTest.java
eduwercamacaro: see
https://github.com/littlehorse-enterprises/littlehorse/pull/667#discussion_r1492914451

server/src/test/java/e2e/UserTaskTest.java
coltmcnealy-lh: Hmm. Yeah, we can delete this test *for now* but we should open a ticket to figure out a better way to implement it.

PR #668, opened 2024-02-16
490 bound the size of the taskqueues lh server

eduwercamacaro:


server/src/main/java/io/littlehorse/common/model/ScheduledTaskModel.java
```diff
@@ -45,6 +46,7 @@ public ScheduledTaskModel(
             TaskDefIdModel taskDefId,
             List<VarNameAndValModel> variables,
             UserTaskRunModel userTaskRun,
+            WfRunIdModel wfRunId,
```
coltmcnealy-lh, line 49: Can't we just do:
```
userTaskRun.getId().getWfRunId()
```
?

server/src/main/java/io/littlehorse/common/util/LHUtil.java
```diff
@@ -186,6 +188,18 @@ public static String toLHName(String oldStr) {
         return str;
     }
 
+    public static WfRunIdModel getWfRunId(TaskRunSourceModel taskRunSource) {
```
coltmcnealy-lh, line 191: I believe we have this method in `LHLibUtil`, right?

server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
```diff
@@ -241,7 +241,7 @@ private RequestExecutionContext requestContext() {
     public KafkaStreamsServerImpl(LHServerConfig config) {
         MetadataCache metadataCache = new MetadataCache();
         this.config = config;
-        this.taskQueueManager = new TaskQueueManager(this);
+        this.taskQueueManager = new TaskQueueManager(this, 20_000);
```
coltmcnealy-lh, line 244: I just realized that 20,000 might be a bit too big. Perhaps 1,000. Because we could easily have 1,000 different taskDef's, which means 20M total tasks...yikes.

This is a similar problem to the WriteBufferManager deciding when to flush Memtables in Speedb :wink:

server/src/main/java/io/littlehorse/server/streams/taskqueue/OneTaskQueue.java
```diff
@@ -79,17 +83,16 @@ public void onRequestDisconnected(PollTaskRequestObserver disconnectedObserver)
...
         // There's two cases here:
         // 1. There are clients waiting for requests, in which case we know that
         // the pendingTaskIds queue/list must be empty.
         // 2. There are no clients waiting for requests. In this case, we just
         // add the task id to the taskid list.
-
         log.trace(
                 "Instance {}: Task scheduled for wfRun {}, queue is empty? {}",
                 hostName,
-                LHLibUtil.getWfRunId(scheduledTaskId.getSource().toProto().build()),
```
coltmcnealy-lh, line 92: Suggestion: get rid of `LHUtil.getWfRunId()`, and change `LHLibUtil()` to take in `TaskRunSourceOrBuilder`.

eduwercamacaro: Done. 
However, I believe that we will transform the model into a 'proto' just because LHUtil only expects a 'proto'.

server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/TaskClaimEvent.java
```diff
@@ -106,6 +110,7 @@ public static TaskClaimEvent fromProto(TaskClaimEventPb proto, ExecutionContext
     public void initFrom(Message p, ExecutionContext context) {
         TaskClaimEventPb proto = (TaskClaimEventPb) p;
         taskRunId = LHSerializable.fromProto(proto.getTaskRunId(), TaskRunIdModel.class, context);
+        taskDefId = LHSerializable.fromProto(proto.getTaskDefId(), TaskDefIdModel.class, context);
```
coltmcnealy-lh, line 113: Do we need this? We already have the `TaskRunId`, so the server can get the `TaskRun` and find out the `TaskDefId` already.

This seems like extra stuff we are sending over the wire for no reason

server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
```diff
@@ -241,7 +241,7 @@ private RequestExecutionContext requestContext() {
     public KafkaStreamsServerImpl(LHServerConfig config) {
         MetadataCache metadataCache = new MetadataCache();
         this.config = config;
-        this.taskQueueManager = new TaskQueueManager(this);
+        this.taskQueueManager = new TaskQueueManager(this, 1_000);
```
coltmcnealy-lh, line 244: Can we move this to `LHConstants.MAX_TASKRUNS_IN_ONE_TASKQUEUE` or something like that?

server/src/main/java/io/littlehorse/server/streams/storeinternals/ReadOnlyGetableManager.java
```diff
@@ -84,6 +84,14 @@ protected <U extends Message, T extends CoreGetable<U>> List<GetableToStore<U, T
         return store.range(start, end, cls);
     }
 
+    public <T extends Storeable<?>> LHKeyValueIterator<T> scanStoreables(String start, String end, Class<T> cls) {
+        return store.range(start, end, cls);
+    }
+
+    public <U extends Message, T extends Storeable<U>> T get(String storeKey, Class<T> cls) {
```
coltmcnealy-lh, line 91: This is somewhat of a leaky abstraction, is it needed?

server/src/main/java/io/littlehorse/server/streams/taskqueue/OneTaskQueue.java
```diff
@@ -79,17 +89,16 @@ public void onRequestDisconnected(PollTaskRequestObserver disconnectedObserver)
      *                        that was just
      *                        scheduled.
      */
-    public void onTaskScheduled(ScheduledTaskModel scheduledTaskId) {
+    public boolean onTaskScheduled(ScheduledTaskModel scheduledTaskId) {
```
coltmcnealy-lh, line 92: Could you please add a comment here saying what the return type is? I am guessing that it is whether or not we need to do a rehydration, but I don't think it would be obvious to someone looking at this code for the first time.

coltmcnealy-lh (approved):
Thanks! This is better, it's fully backwards compatible. A few requests before merging:
- The `TaskClaimEvent` does not need the `TaskDefId`. That seems wasteful
- Magic Number (task queue max size) should be a constant
- A few methods could use comments
- Let's think about leaky abstractions for having the `GetableManager()` return a `Storeable`

server/src/main/java/io/littlehorse/server/streams/storeinternals/ReadOnlyGetableManager.java
coltmcnealy-lh: I would prefer that it is a `public ScheduledTaskModel get(TaskRunId taskRunId)`. I don't think we should be exposing arbitrary `Storeable`s in the `GetableManager`, since the `GetableManager` manages `Getable`s (not `Storeable`s)

server/src/main/java/io/littlehorse/server/streams/taskqueue/OneTaskQueue.java
eduwercamacaro: Oops. This is no longer required.

PR #673, opened 2024-02-20
Bugfix/store principal for specific tenant

eduwercamacaro:
Fix:
Enable `lhctl put principal` to specify a `tenantId`

### DIFF
--- a/lhctl/cmd/principal.go
+++ b/lhctl/cmd/principal.go
@@ -87,5 +87,5 @@ var (
 func init() {
 	putCmd.AddCommand(putPrincipalCmd)
 	putPrincipalCmd.Flags().String("acl", "", "ACLs")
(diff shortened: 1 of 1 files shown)


lhctl/cmd/principal.go
```diff
@@ -87,4 +87,5 @@ var (
 func init() {
 	putCmd.AddCommand(putPrincipalCmd)
 	putPrincipalCmd.Flags().String("acl", "", "ACLs")
+	putPrincipalCmd.Flags().String("tenantId", "", "Tenant associated with the principal")
```
coltmcnealy-lh, line 90: I thought that the `Principal` had a `map<string, ServerACLs> per_tenant_acls` which allows different ACL's per tenant. Is this possible to create using `lhctl` or is that not posible?

coltmcnealy-lh (approved):

Issue #676, opened 2024-02-21
Refactor and remove duplicated component TaskExecutionMetrics.tsx


mijailr:
## Context

The component `TaskExecutionMetrics.tsx` has a dup sibling component at task level that should be removed, also the component is quite complex and has a lot of responsibilities error prone and it's not taking advantage of server side rendering.

coltmcnealy-lh:
Let's put this on hold, because the "Metrics API" is not considered stable and is highly likely to change in the next few weeks.

PR #680, opened 2024-02-21
Handle dashboard loading when there's no data

### LINKED ISSUE #410: Improve user feedback when no items
* when no items were returned from the backend we show “loading” we should show a different message. THis happens on every panel and search result component.

image

mijailr:
Show a `No data available` instead of persisting the `Loading...` message.

image

Fixes #410

This doesn't solve the issue in inner pages and will be fixed through #676


coltmcnealy-lh (approved):
Good first improvement :pray:

PR #686, opened 2024-02-26
Add sdk-js 

### LINKED ISSUE #677: Create proto client for JS
## Context

Dashboard is currently using a "manual" version of the proto client, we should generate it from the proto files as we do with `java-sdk`

mijailr:
## Context

This creates the foundations for `sdk-js` it currently supports all client requests. With this config keys:

```
LHC_API_HOST
LHC_API_PORT
LHC_API_PROTOCOL
LHC_TENANT_ID
LHC_CA_CERT
```

It also publishes the node package to npm [littlehorse-client](https://www.npmjs.com/package/littlehorse-client)

Closes #677


sauljabin (approved):
no special comments, just maybe, is there any new config? if so we have to update thishttps://github.com/littlehorse-enterprises/littlehorse/blob/master/docs/CLIENT_CONFIGURATIONS.md

local-dev/compile-proto.sh
```diff
@@ -84,4 +86,13 @@ $docker_run protoc \
 echo "Fixing imports for protobuff JS for more information see the README file"
 find "${WORK_DIR}"/dashboard/apps/web/littlehorse-public-api/ -type f -readable -writable -exec sed -i "s/import _m0/import * as _m0/g" {} \;
 
+
+# This segment is for the sdk-js should replace the dashboard version eventually
+$docker_run protoc \
+	--plugin=/usr/local/lib/node_modules/ts-proto/protoc-gen-ts_proto \
```
coltmcnealy-lh, line 92: Is it possible to combine this with the other one on line 44?

coltmcnealy-lh (approved):

coltmcnealy-lh:
I think next step is to get the dashboard to use this new library then we can have a Big Red Diff. Good work!

local-dev/compile-proto.sh
mijailr: Sadly no, since they generate different outputs

PR #694, opened 2024-03-01
490 bound the size of the taskqueues lh server

eduwercamacaro:


server/src/main/java/io/littlehorse/server/streams/storeinternals/ReadOnlyGetableManager.java
```diff
@@ -84,6 +85,23 @@ protected <U extends Message, T extends CoreGetable<U>> List<GetableToStore<U, T
...
+        return store.range(start, end, Tag.class);
+    }
+
+    /**
+     * Only use this method if you need to retrieve a Storeable
+     * @param storeKey Storeable's storekey
+     * @param cls Storeable's class
+     * @return Null if not found
+     */
+    public <U extends Message, T extends Storeable<U>> T get(String storeKey, Class<T> cls) {
```
eduwercamacaro, line 101: @coltmcnealy-lh  This is leaking the implementation detail. Any suggestion here? see `OneTaskQueue#rehydrateFromStore`

coltmcnealy-lh: Can we just rename it to `getStoreable()`? I don't like overloading the `get()` method, which is used to get a `Getable`.

coltmcnealy-lh: Actually, try this:

```
public ScheduledTaskModel getScheduledTask(TaskRunIdModel scheduledTaskId);
```

That still leaks, but it leaks a lot less.

coltmcnealy-lh (approved):

server/src/main/java/io/littlehorse/server/streams/storeinternals/ReadOnlyGetableManager.java
eduwercamacaro: That leaks the business application into the store abstraction. ;)

Issue #700, opened 2024-03-07
Add variable values to exceptions


coltmcnealy-lh:
There is already a `content` field on the `Failure` proto. However, it is not always used throughout the code, and is quite often `null`. We should also ensure that the `WorkflowThread#fail()` method has the ability to provide the content, and also add a test to verify that the `FailureHandler` `ThreadRun` can access the `content` via the variable named `INPUT`.

PR #714, opened 2024-03-13
700 add variable values to exceptions

eduwercamacaro:
solves #700


sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/LHServerConnectionManager.java
```diff
@@ -306,6 +306,7 @@ private io.littlehorse.sdk.common.proto.LHTaskException exnToTaskException(LHTas
         return io.littlehorse.sdk.common.proto.LHTaskException.newBuilder()
```
coltmcnealy-lh, line 306: @mjsax this is what I was tweeting about :smile:

server/src/test/java/e2e/FailureHandlingTest.java
```diff
@@ -185,7 +198,9 @@ public Workflow handleClientSpecificException() {
         return new WorkflowImpl("handle-client-specific-exception", thread -> {
             NodeOutput node = thread.execute("business-exception-failure");
             thread.handleException(node, handler -> {
-                handler.execute("my-handler");
+                WfRunVariable failureContent =
+                        handler.addVariable(WorkflowThread.HANDLER_INPUT_VAR, VariableType.JSON_OBJ);
```
coltmcnealy-lh, line 202: Perfect, this e2e is exactly what we need. Only thing left to do is update the documentation :smile:

coltmcnealy-lh (approved):
Great work. Looks like it was halfway implemented in the past, and now it's working. The e2e is exactly what we want.

PR #717, opened 2024-03-14
Creates new Request Context for each onNext in the PollObserver

eduwercamacaro:


examples/basic/src/main/java/io/littlehorse/examples/BasicExample.java
```diff
@@ -1,6 +1,9 @@
 package io.littlehorse.examples;
 
+import io.littlehorse.sdk.common.LHLibUtil;
```
coltmcnealy-lh, line 3: ?

coltmcnealy-lh (approved):

PR #721, opened 2024-03-15
LH Compatible with java 21

sauljabin:


.github/workflows/tests.yml
```diff
@@ -68,27 +68,33 @@ jobs:
...
+        java-version: [ "11", "17", "21" ]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
       - name: Setup Java
         uses: actions/setup-java@v3
         with:
           distribution: "corretto"
-          java-version: 17
+          java-version: ${{ matrix.java-version }}
```
coltmcnealy-lh, line 97: Do we really need to run a matrix for the server? I think we can just support only the latest version of Java SDK, especially since we will be distributing it as a docker image in production.

For the `sdk-java`, a matrix is a very good idea.

mijailr: I agree, server just the main version, sdk-java down to 17.

sauljabin: The latest LTS is java 21, so far let's have a matrix with 17 and 21, and then let's remove 17

sauljabin:
@coltmcnealy-lh @mijailr so far is working, but some tests are failing in 21
This class is failing: WfSpecModelTest

sauljabin:
we can continue with this when we have deleted and migrated the e2e tests

.github/workflows/tests.yml
coltmcnealy-lh: sdk-java should continue to support Java 11. Kafka supports all the way down to Java 8. Some of our users may be really outdated; I would hate to lose a deal because we had one less step in our pipeline

server/src/test/java/e2e/ChildThreadTest.java
```diff
@@ -0,0 +1,199 @@
...
+        verifier.prepareRun(basicChildThread)
+                .thenVerifyWfRun(wfRun -> {
+                    // The child should be created in the first Command
+                    Assertions.assertThat(wfRun.getThreadRuns(1).getStatus()).isEqualTo(LHStatus.RUNNING);
+                })
+                .thenSendExternalEventWithContent(COMPLETE_CHILD, Map.of("result", "not-an-integer"))
+                .thenVerifyNodeRun(1, 1, nodeRun -> {
+                    // I thought this should be true, but apparently it's not. We need to review this.
+                    // Assertions.assertThat(nodeRun.getStatus()).isEqualTo(LHStatus.ERROR);
+                    Assertions.assertThat(nodeRun.getFailuresCount()).isEqualTo(1);
```
coltmcnealy-lh, line 81: I think I found a problem, or at least behavior that is different than what I expected. I had a NodeRun whose SubNodeRun completed properly but the variable mutations failed. The result is:
- :smiley: The ThreadRun is in the ERROR state
- :thinking: There is a Failure with type VAR_MUTATION_ERROR on the NodeRun rather than VAR_MUTATION_ERROR
- :thinking_face: The NodeRun is in the status COMPLETED

I thought that we decided that since an Edge is part of a Node, and the VariableMutation lives on the Edge, then technically the VariableMutation is part of the NodeRun itself. So therefore the NodeRun should be in the ERROR status...what do you think?
My tests show that this is not necessarily the case. Maybe I need to review that PR I made in March which fixes this 

(Link to original slack message, private to LH employees: https://littlehorseworkspace.slack.com/archives/C060A2XCX9D/p1725819836840409)


[11:27](https://littlehorseworkspace.slack.com/archives/C060A2XCX9D/p1725820034527379)
Also, another problem is that the failure is of the type VAR_SUB_ERROR not VAR_MUTATION_ERROR. I don't know if we actually use the status VAR_MUTATION_ERROR. Maybe we can get rid of it?

test-utils/src/main/java/io/littlehorse/test/WfRunVerifier.java
```diff
@@ -72,7 +73,8 @@ public WfRunVerifier thenVerifyTaskRunResult(
         return this;
     }
 
-    public WfRunVerifier thenVerifyAllTaskRuns(int threadRunNumber, Consumer<List<TaskRun>> verifier) {
+    public WfRunVerifier thenVerifyAllTaskRuns(Consumer<List<TaskRun>> verifier) {
+        steps.add(new VerifyAllTaskRunsStep(verifier, steps.size() + 1));
```
coltmcnealy-lh, line 77: This was very broken...when we tried to add the step it would accept it and then  silently not check it.

coltmcnealy-lh (approved):
LOL, technically since this PR was originally opened by @sauljabin I can approve it :sweat_smile: 

However, let's wait till @eduwercamacaro has a look before I approve it and go with it.

.github/workflows/tests.yml
```diff
@@ -84,6 +84,9 @@ jobs:
   tests-canary:
     if: ${{ !contains(github.event.head_commit.message, '[skip main]') }}
     runs-on: ubuntu-latest
+    strategy:
```
sauljabin, line 87: not needed if you hard coded the java version, example:` java-version: 21`

.github/workflows/tests.yml
```diff
@@ -92,14 +95,17 @@ jobs:
...
 
       - name: Tests
         run: ./gradlew canary:spotlessCheck canary:test canary:build
 
   tests-server:
     if: ${{ !contains(github.event.head_commit.message, '[skip main]') }}
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java-version: [ "21" ]
```
sauljabin, line 108: same here

.github/workflows/tests.yml
```diff
@@ -108,30 +114,14 @@ jobs:
...
-          java-version: 17
+          java-version: 21
 
       - name: Validate Indentations
         run: ./gradlew server:spotlessCheck test-utils:spotlessCheck
 
       - name: Tests
         run: ./gradlew server:test server:e2e
 
-  tests-e2e:
```
sauljabin, line 119: nice

test-utils/src/main/java/io/littlehorse/test/internal/StandaloneTestBootstrapper.java
```diff
@@ -33,9 +33,9 @@ public StandaloneTestBootstrapper() {
     }
 
     public void setup() throws Exception {
-        kafka = new KafkaContainer(DockerImageName.parse("apache/kafka:3.7.0"));
+        kafka = new KafkaContainer(DockerImageName.parse("apache/kafka-native:3.8.0"));
```
sauljabin, line 36: Interesting

server/src/test/java/e2e/ChildThreadTest.java
eduwercamacaro: Hazim had a similar conversation a few months ago, in which he mentioned that the LH server handles mutation failures in a confusing way. Given the current design, I believe this is the best mechanism for handling variable mutation failures, as the node itself didn't fail. For instance, consider a wfspec that defines a TaskNode, defines several other variables, and then mutates those variables.

```
wf.execute('my-silly-validations');

wf.addVariable(...)
// More variables here

wf.mutate(...)
// More mutations here
```

Debugging will be quite confusing at first, clients may believe that the 'my-silly-validations' node has failed, which isn't the case. Therefore, in this scenario, it's preferable to have an edge failure within the node, indicating that the node should be completed. If we really want to change this behavior, I recommend splitting nodes and edges into separate thread-level objects.

server/src/test/java/e2e/RescueThreadRunTest.java
```diff
@@ -83,7 +86,14 @@ void shouldExecuteFailedNodeAgainOnRescue() {
                 .waitForNodeRunStatus(0, 1, LHStatus.ERROR)
                 .thenRescueThreadRun(0, false) // dont skip failed node: try again
                 .waitForStatus(LHStatus.COMPLETED)
-                .thenVerifyAllTaskRuns(0, taskRuns -> {
+                .thenVerifyAllTaskRuns(unsortedTaskRuns -> {
```
eduwercamacaro, line 89: In my opinion, this is not semantically correct and goes against the objectives of this test framework, which is to provide a way to describe a set of verifiable individual constraints that enable clients (server devs)  describe how the workflow should look after applying some inputs (external events, input variables, user task forms, etc.). What I see in the `thenVerifyAllTaskRuns` method is a way to break that promise, because this allows to include a big step that executes a lot of assertions in it. My suggestion here is to include separate steps with fixed node positions to verify each task run.

coltmcnealy-lh: Can we make this change in another PR?

server/src/test/java/e2e/ChildThreadTest.java
coltmcnealy-lh: There is a difference between the `SubNodeRun` and the `NodeRun`. I think the fact that the `NodeRun` has a `Failure` on it and yet its status is `COMPLETED` is bad. Especially because the `Mutation` is part of the `Node`

server/src/test/java/e2e/RescueThreadRunTest.java
eduwercamacaro: Sure, we can create a separate PR for this.

eduwercamacaro (approved):

Issue #562, opened 2023-12-26
Implement Retention Policy in External Event


sauljabin:
### Context

PENDING

### Acceptance Criteria

PENDING

**Scenario** [the name for the behavior that will be described]
**Given** [the beginning state of the scenario]
**When** [specific action that the user makes]
**Then** [the outcome of the action in **When**]
**And** [used to continue any of three previous statements]

### Components

- [ ] Dashboard
- [X] Server
- [X] Python SDK
- [X] Go SDK
- [X] Java SDK
- [X] C# SDK
- [ ] LH Control
- [ ] LH Tests Utils

coltmcnealy-lh:
I think I had actually already implemented this ticket. However, the `lhctl get externalEvent` is not working. I willl need to debug that

coltmcnealy-lh:
Resolution:
* The `PutExternalEventDefModel#process()` was **not** saving the `retentionPolicy` properly on the new `ExternalEventDef`. That is fixed.
* The recent changes to `lhctl` caused segmentation faults when the tenant id was not set. That has been fixed.
* A pull request from earlier made a breaking change to the `ExternalEvent` proto which is why the `lhctl get externalEvent` didn't work. Once I updated `lhctl` to the new version via `go install ./lhctl`, then it has worked.

coltmcnealy-lh:
Implemented by some PR I can't find right now

PR #724, opened 2024-03-17
Fixes lhctl and External Event Retention

coltmcnealy-lh:
- lhctl was throwing a segfault if tenant id wasn't set
- PutExternalEventDefRequest wasn't saving the external event retention policy

Manual tests have confirmed that external event retention is working as specified.


coltmcnealy-lh:
Fixes #562

sauljabin (approved):

PR #727, opened 2024-03-19
FIX: internal task queue concurrency

coltmcnealy-lh:
Previous implementation was incorrect in keeping track of whether there are more tasks on disk.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streams/taskqueue/OneTaskQueue.java
+++ b/server/src/main/java/io/littlehorse/server/streams/taskqueue/OneTaskQueue.java
@@ -43,7 +43,7 @@ public class OneTaskQueue {
     /*
      * If it is true, the queue should execute a task rehydration from store
      */
-    private boolean outOfCapacity;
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/littlehorse/server/streams/taskqueue/OneTaskQueue.java
```diff
@@ -127,8 +127,8 @@ public boolean onTaskScheduled(ScheduledTaskModel scheduledTask) {
                 luckyClient = hungryClients.poll();
             } else {
                 // case 2
-                outOfCapacity = !pendingTasks.offer(scheduledTask);
-                return !outOfCapacity;
+                hasMoreTasksOnDisk = !pendingTasks.offer(scheduledTask) || hasMoreTasksOnDisk;
```
coltmcnealy-lh, line 130: This line was the problem. The other changes were just renaming the variable to make more sense.

sauljabin (approved):

PR #734, opened 2024-03-22
Modify Liveness Controller

sauljabin:


sdk-python/littlehorse/worker.py
```diff
@@ -33,7 +33,7 @@
 
 REPORT_TASK_DEFAULT_RETRIES = 5
 HEARTBEAT_DEFAULT_INTERVAL = 5
-HEALTH_TIMEOUT = 60000
+GRPC_UNARY_CALL_TIMEOUT = 1
```
sauljabin, line 36: @coltmcnealy-lh this is important, is 1sec OK?

examples/health-status/README.md
```diff
@@ -0,0 +1,39 @@
...
+2. Pass that variable into the execution of the "greet" task.
+3. Shows the cluster's health status.
+
+Let's run the example in `HealthStatus.java`
+
+```
+./gradlew example-health-status:run
+```
+
+In another terminal, use `lhctl` to run the workflow:
```
coltmcnealy-lh, line 15: How does this have something to do with the health status? The example isn't required (documentation _is_ required once we have added it to our OSS repo), but if you have it, please explain in the README how the health status thing works, and what the example does. Thanks!

examples/health-status/src/main/java/io/littlehorse/examples/HealthStatus.java
```diff
@@ -0,0 +1,93 @@
...
+public class HealthStatus {
+
+
+    private static final Logger log = LoggerFactory.getLogger(HealthStatus.class);
+
+    public static Workflow getWorkflow() {
+        return new WorkflowImpl(
+            "example-health-status",
+            wf -> {
+                WfRunVariable theName = wf.addVariable("input-name", VariableType.STR).searchable();
```
coltmcnealy-lh, line 36: How is this any different from the `BasicExample`?

sdk-python/littlehorse/worker.py
coltmcnealy-lh: Way too small...at least 20 seconds I think. In certain cases, if the LH Server is in a different region from the task worker, the round-trip latency might be close to 500 milliseconds without any dropped packets. Where is this used?

sdk-python/littlehorse/worker.py
```diff
@@ -333,7 +334,7 @@ async def _report_task(self, task_result: ReportTaskRun, retries_left: int) -> N
         )
 
         try:
-            await self._stub.ReportTask(task_result)
+            await self._stub.ReportTask(task_result, timeout=GRPC_UNARY_CALL_TIMEOUT)
```
coltmcnealy-lh, line 337: Oh, I see. It's used here. Some comments:

1. Default deadline for LittleHorse grpc requests are 60 seconds
2. It should be configurable (probably not in this PR though)

examples/health-status/src/main/java/io/littlehorse/examples/HealthStatus.java
sauljabin: i will delete the examples

sdk-python/littlehorse/worker.py
sauljabin: sure

coltmcnealy-lh (approved):
Looks good, eventually we should make the timeout a configurable property.

PR #740, opened 2024-03-26
feat(dashboard): Graph and wfRun views

mijailr:
This PR includes graph and wfRun view


dashboard-new/src/app/(authenticated)/components/Search.tsx
```diff
@@ -1,88 +1,48 @@
...
-        setResponse({
-          ...data,
-          results: [...response?.results, ...data.results] as any, // it's impossible know at build time which type would contain response
-        })
-      })
-    }
-    // response is not really a dependency of this hook
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [bookmark])
+  const { isPending, data, hasNextPage, fetchNextPage } = useInfiniteQuery({
```
coltmcnealy-lh, line 21: Well this is actually really cool

dashboard-new/src/app/(authenticated)/components/Search.tsx
```diff
@@ -1,88 +1,48 @@
...
       ) : (
         <div className="min-h-[360px]">
-          {response?.type === 'WfSpec' && <WfSpecTable items={response.results} />}
-          {response?.type === 'TaskDef' && <TaskDefTable items={response.results} />}
-          {response?.type === 'UserTaskDef' && <UserTaskDefTable items={response.results} />}
-          {response?.type === 'ExternalEventDef' && <ExternalEventDefTable items={response.results} />}
+          {type === 'WfSpec' && <WfSpecTable pages={data?.pages} />}
+          {type === 'TaskDef' && <TaskDefTable pages={data?.pages} />}
+          {type === 'UserTaskDef' && <UserTaskDefTable pages={data?.pages} />}
+          {type === 'ExternalEventDef' && <ExternalEventDefTable pages={data?.pages} />}
```
coltmcnealy-lh, line 42: This looks like it could be reusable for TaskRun and WfRun search.

dashboard-new/src/app/(authenticated)/components/searchAction.ts
```diff
@@ -48,15 +49,15 @@ const bookmarkFrom = (bookmark?: string): Buffer | undefined => {
   return Buffer.from(bookmark, 'base64')
 }
 
-interface SearchResult {
+export interface SearchResult {
   type: SearchType
   bookmark?: string
   results: any
```
coltmcnealy-lh, line 55: Are there generics in Javascript? If so, this would be a good use-case for it. Alternatively, you could just pass the proto classes around.

dashboard-new/src/app/(authenticated)/wfSpec/[...props]/components/WfRuns.tsx
```diff
@@ -0,0 +1,85 @@
...
+    queryKey: ['wfRun', status, tenantId, limit, startTime],
+    initialPageParam: undefined,
+    getNextPageParam: (lastPage: WfRunIdList) => lastPage.bookmark?.toString('base64'),
+    queryFn: async ({ pageParam }) => {
+      console.log({ startTime })
+      return await searchWfRun({
+        wfSpecName: id!.name,
+        wfSpecMajorVersion: id!.majorVersion,
+        wfSpecRevision: id!.revision,
+        variableFilters: [],
```
Comment (coltmcnealy-lh) [inline dashboard-new/src/app/(authenticated)/wfSpec/[...props]/components/WfRuns.tsx:45]: You can use this to search by variables, but it only works for variables in the Entrypoint ThreadRun.

Note that there is also `rpc SearchVariable` which is much more performant, but it requires the variables to be indexed. However it allows variables in any threadRun (so long as the variable has an index)

coltmcnealy-lh:
This looks good to me. Well done.

README would be nice (:

dashboard-new/src/app/(authenticated)/components/searchAction.ts
mijailr: Yes, that's actually what I'm doing. Although not for this interface because I'm using that just as the parent interface, then below this I'm mapping which `SearchType` belongs to each `result` type. That is because I'm using a generic search function which resolves the type.

mijailr:
> This looks good to me. Well done.
> 
> README would be nice (:

README will be on next PR which enables the docker image and local run.

PR #742, opened 2024-03-27
Create principals with admin privileges and no tenant

eduwercamacaro:
This PR enables LH server to create principal without specifying any tenant. This is useful when clients need to create principal that are able to do any operation in any tenant. 
The principal requester must have admin privileges:

- `ACL_ALL_RESOURCES`

- `ALL_ACTIONS`

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutPrincipalRequestModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutPrincipalRequestModel.java
@@ -98,8 +98,9 @@ public class PutPrincipalRequestModel extends MetadataSubCommand<PutPrincipalReq
 
             toSave.setCreatedAt(oldPrincipal.getCreatedAt());
         }
-
-        if (perTenantAcls.isEmpty()) {
+        boolean canWriteAdminPrincipals = requester.isAdmin();
+        if (perTenantAcls.isEmpty()
+                && !(canWriteAdminPrincipals && !globalAcls.getAcls().isEmpty())) {
             throw new LHApiException(Status.INVALID_ARGUMENT, "Must provide list of tenants");
         }
         ensureThatIsAllowedToWriteInRequestedTenants(requester);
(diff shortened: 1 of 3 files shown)


test-utils/src/main/java/io/littlehorse/test/LHExtension.java
```diff
@@ -107,7 +107,6 @@ private void maybeCreateTenantAndPrincipal(TestContext testContext) {
                     .putPrincipal(PutPrincipalRequest.newBuilder()
                             .setId(principalId)
                             .setGlobalAcls(acls)
-                            .putPerTenantAcls(testContext.getConfig().getTenantId(), acls)
```
eduwercamacaro, line 110: No need to do this anymore. This makes the default principal a Tyler for end-to-end tests.

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutPrincipalRequestModel.java
```diff
@@ -98,8 +98,9 @@ public Principal process(MetadataCommandExecution context) {
 
             toSave.setCreatedAt(oldPrincipal.getCreatedAt());
         }
-
-        if (perTenantAcls.isEmpty()) {
+        boolean canWriteAdminPrincipals = requester.isAdmin();
+        if (perTenantAcls.isEmpty()
+                && !(canWriteAdminPrincipals && !globalAcls.getAcls().isEmpty())) {
```
coltmcnealy-lh, line 103: I am confused—what is the purpose of this line? And why do we want to throw the error?

eduwercamacaro: Basically `!(canWriteAdminPrincipals && !globalAcls.getAcls().isEmpty())` this condition verifies that if there is no per tenant acls at least it contains global acls, and it verifies that the requester is an admin principal

eduwercamacaro: Ok, we will allow store principal without any privilege.

coltmcnealy-lh (approved):
Thanks!

PR #744, opened 2024-03-28
743 server listeners that dont use any authorizers must use anonymous principals by default

eduwercamacaro:


server/src/main/java/io/littlehorse/server/listener/ServerListener.java
```diff
@@ -40,6 +41,7 @@ public ServerListener(
                 .intercept(new RequestAuthorizer(
                         service, executionContextKey, metadataCache, coreStoreProvider, config.getConfig()))
                 .intercept(config.getAuthorizer())
+                .intercept(new RequestSanitizer())
```
coltmcnealy-lh, line 44: Do we need to think more about the interceptors ordering here? I think the order we want them to go is:

* First, strip the header
* Second, figure out the oauth client id
* Third: translate oauth client id into principal

I think this works actually...it's just weird how they do the ordering backwards.

eduwercamacaro: Yes, it is the reverse order, so in the chain, first interceptor is sanitizer

eduwercamacaro:
@coltmcnealy-lh There is no sdk that allows client id header, so we don't need to modify our sdks.

PR #749, opened 2024-04-03
Add validation for principal ID containing slashes

yesidevelop:
This commit adds a check to ensure that the principal ID provided does not contain slashes. If the ID contains a slash ("/"), it throws a LHApiException with an INVALID_ARGUMENT status, indicating that slashes are not allowed in the principal ID.

This validation helps maintain data integrity and prevents potential issues with principal IDs.

Issue 456

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutPrincipalRequestModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutPrincipalRequestModel.java
@@ -85,10 +85,15 @@ public class PutPrincipalRequestModel extends MetadataSubCommand<PutPrincipalReq
         PrincipalModel toSave = new PrincipalModel();
         toSave.setId(new PrincipalIdModel(id));
 
-        // Check if the ID contains a slash
-        if (id.contains("/")) {
-            throw new LHApiException(Status.INVALID_ARGUMENT, "Principal ID cannot contain slashes.");
+        char[] disallowedCharacters = {'/', '\\'};
+        // Check if the ID contains any disallowed characters
+        for (char disallowedChar : disallowedCharacters) {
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutPrincipalRequestModel.java
```diff
@@ -84,6 +84,11 @@ public Principal process(MetadataCommandExecution context) {
                 context.service().getPrincipal(context.authorization().principalId());
         PrincipalModel toSave = new PrincipalModel();
         toSave.setId(new PrincipalIdModel(id));
+
+        // Check if the ID contains a slash
+        if (id.contains("/")) {
```
coltmcnealy-lh, line 89: Can we also add `\` as well? We _might_ want to use that character for nested principals, which is a feature floating around in my head.

coltmcnealy-lh (approved):

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutPrincipalRequestModel.java
yesidevelop: Updated the code. 
In Linux, when we run command (with single backslash `littlehorse\1`) like 
`lhctl put principal littlehorse\1 --acl "acl_principal:write,acl_workflow:read,acl_tenant:write" --tenantId A` It sends principal id as `littlehorse1` automatically by removing the backslash (So it passes validation check). 

This command can be used instead to receive validation error from Linux command line (with double backslash `littlehorse\\1`)

`lhctl put principal littlehorse\\1 --acl "acl_principal:write,acl_workflow:read,acl_tenant:write" --tenantId A`

coltmcnealy-lh:
Ready to merge!

Issue #692, opened 2024-02-27
Search available tenants


mijailr:
### Context

Something like this might work

```proto
rpc SearchTenant(SearchTenantRequest) returns (TenantIdList) {}

message SearchTenant {
  int32 limit = 1;
  optional bytes bookmark = 2;
}

message TenantIdList {
  repeated TenantId tenants = 1;
  optional bytes bookmark = 2;
}
```

### Acceptance Criteria

Should return the list of tenants if the principal has `global_acls` permissions.

coltmcnealy-lh:
This should require the acl `READ` over `TENANT` in the `global_acls` field of the `Principal` making the call.

coltmcnealy-lh:
After thinking further, I think we may need to extend the way the `@Authorize` decorator works. Maybe @eduwercamacaro can weigh in on this tomorrow.

jhosep-marin:
In the end, the team decided to keep this as follows:

`rpc SearchTenant(SearchTenantRequest) returns (TenantIdList) {}`

PR #750, opened 2024-04-04
692 search available tenants

jhosep-marin:
Search Available Tenants (#692)

- Adds Search Tenant gRPC endpoint.
- Enables gRPC endpoint in lhctl.
- Updates Proto schema with respective definitions used in new endpoint.
- Adds exception that gets thrown whenever a principal tries to assign tenant resources as perTenantACLs.
- Adds test methods that provide Tenant ACLs.


server/src/main/java/io/littlehorse/common/model/getable/objectId/TenantIdModel.java
```diff
@@ -44,7 +44,7 @@ public String toString() {
 
     @Override
     public void initFromString(String storeKey) {
-        this.id = storeKey;
+        this.id = storeKey.split("/")[1];
```
coltmcnealy-lh, line 47: Why is this necessary? This looks like it could permit many bugs wherein I pass in a `WfSpecId` and get out a `TenantId` for example.

server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
```diff
@@ -39,166 +29,24 @@
...
-import io.littlehorse.sdk.common.proto.ListExternalEventsRequest;
-import io.littlehorse.sdk.common.proto.ListNodeRunsRequest;
-import io.littlehorse.sdk.common.proto.ListTaskMetricsRequest;
-import io.littlehorse.sdk.common.proto.ListTaskMetricsResponse;
-import io.littlehorse.sdk.common.proto.ListTaskRunsRequest;
-import io.littlehorse.sdk.common.proto.ListUserTaskRunRequest;
-import io.littlehorse.sdk.common.proto.ListVariablesRequest;
-import io.littlehorse.sdk.common.proto.ListWfMetricsRequest;
-import io.littlehorse.sdk.common.proto.ListWfMetricsResponse;
+import io.littlehorse.sdk.common.proto.*;
```
coltmcnealy-lh, line 39: Hmm, @eduwercamacaro do we have a style guide for these changes?

coltmcnealy-lh (commented):
I will need to test some of the other search functionality here...this PR seems much more difficult than I thought it would be.

As a rule, we are going to have non-implementers do testing of PR's before they get merged

server/src/main/java/io/littlehorse/common/model/getable/objectId/TenantIdModel.java
eduwercamacaro: When we are iterating the result of a range scan, we pass the full store key to the `ObjectIdModel#fromString` method. This is very common with the other ObjectId objects see `TaskRunIdModel.initFromString` , and I believe we should do the same for this `TenantIdModel`. 
Could you please explain to me how this can allow many bugs that are not present in the previous version of this class?

server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
eduwercamacaro: spotless only verify linting problems. In order to add a validation for this can of changes we need a tool like Checkstyle. see ([AvoidStarImport](https://checkstyle.sourceforge.io/checks/imports/avoidstarimport.html#AvoidStarImport))

coltmcnealy-lh: Do you have an opinion about these changes?

coltmcnealy-lh:
TODO's:

1. Prevent having a `/` in a `TenantId` in order to prevent injection attacks. For example, the following code causes a crash:
```
lhctl put tenant 123/123
lhctl search tenant
```
2. Rename `objectIdExtractorPattern` -> `tenantScopedObjectIdExtractorPattern`
3. `TenantIdModel#initFromString` should accept a `objectId`. It should not do a split()

server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
eduwercamacaro: we should not use wildcard imports

jhosep-marin:
@coltmcnealy-lh I think the latest requested changes are in place. Please, have a look one more time, and let me know if there is anything else we need to do here.

coltmcnealy-lh (approved):

PR #752, opened 2024-04-08
A task can contain placeholders on its name. 

diablouma:
You can have placeholders in the taskDefName in the form of:

`a-task-name-${PLACEHOLDER_1}-${PLACEHOLDER-2}`

Each placeholder should be replaced by its corresponding value coming from the valuesForPlaceHolders map.
     * ` PLACEHOLDER_1: VALUE_1`
     *  `PLACEHOLDER_2: VALUE_2`
 So after the values are replaced, you will have a taskDefName like:` a-task-name-VALUE_1-VALUE_2`


coltmcnealy-lh (approved):
Looks good but Edu should check it

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/TaskDefBuilder.java
```diff
@@ -5,15 +5,17 @@
...
 
 public class TaskDefBuilder {
 
     public Object executable;
     public String taskDefName;
     public LHTaskSignature signature;
 
-    public TaskDefBuilder(Object executable, String taskDefName) throws TaskSchemaMismatchError {
-        signature = new LHTaskSignature(taskDefName, executable);
+    public TaskDefBuilder(Object executable, String taskDefName, Map<String, String> valuesForPlaceHolders)
```
coltmcnealy-lh, line 16: @eduwercamacaro do we expose this class to our clients? And should we? I think it is internal, right?

coltmcnealy-lh:
I have tested this PR locally on my laptop and it all works. However, the tests aren't passing. Once the pipeline is green, feel free to merge.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/TaskDefBuilder.java
eduwercamacaro: It is internal, and we should not expose this to our clients. However, I suggest reducing the access for this class's constructor and making it final for now.

diablouma: @eduwercamacaro you mean making the class TaskDefBuilder to be final?

coltmcnealy-lh: Yes. However, I think that should be done in a separate Issue / PR because it is not related to Nelson's improvement. We also need to do that for many other classes which we "leak" to our users, so it deserves a GH Issue so that we can address the entire problem rather than just only one class.

@diablouma feel free to merge this!

PR #754, opened 2024-04-09
599 throw exception on cancelled user task runs

eduwercamacaro:
This PR adds the ability to both cancel a user task after a deadline on assignment or arrival and handle user task cancellation as a business exception.

**User Task Canceled exception:**
Now, clients can define a workflow spec that catch a EXCEPTION when a user task run gets canceled via rpc or scheduled action. In case there is no exception name defined for the user task node, it will be treated as a technical ERROR, for the WfSpec needs to define an exception for the user task cancellation.

**Cancel a UserTaskRun:**
Current versions allows cancel a user task run via rpc, but some use cases require cancel an specific user task run after it was created or assigned, for that this PR includes two new methods:

```
/**
     * Cancels a User Task Run if it exceeds a specified deadline.
     * @param userTask is a reference to the UserTaskNode that will be canceled after the deadline
     * @param delaySeconds is the delay time after which the User Task Run should be canceled
     */
    void cancelUserTaskRunAfter(UserTaskOutput userTask, Serializable delaySeconds);

    /**
     * Cancels a User Task Run if it exceeds a specified deadline after it is assigned
     * @param userTask is a reference to the UserTaskNode that will be canceled after the deadline
     * @param delaySeconds is the delay time after which the User Task Run should be canceled
     */
    void cancelUserTaskRunAfterAssignment(UserTaskOutput userTask, Serializable delaySeconds);
```

Note: Changes are properly propagated for every sdk clients (Java, Python and Go)

**Test Utils:**
Now it is possible to cancel a user task via RPC with the following method:
`.thenCancelUserTaskRun(0, 1)`


schemas/wf_spec.proto
```diff
@@ -207,6 +207,9 @@ message UserTaskNode {
 
   // Allow WfRun-specific notes for this User Task.
   optional VariableAssignment notes = 6;
+
+  // Specifies the name of the exception thrown when the User Task is canceled
+  optional VariableAssignment on_cancel_exception_name = 7;
```
coltmcnealy-lh, line 212: :heart: 

One request—this is purely cosmetic. Could we rename it to `on_cancellation_exception_name`? That's perhaps slightly more semantic.

sdk-go/wflib/wf_lib_internal_test.go
```diff
@@ -85,6 +85,21 @@ func TestUserTaskWithNotes(t *testing.T) {
...
+	}, "my-workflow")
+
+	putWf, _ := wf.Compile()
+	entrypoint := putWf.ThreadSpecs[putWf.EntrypointThreadName]
+	node := entrypoint.Nodes["1-sample-user-task-USER_TASK"]
+
+	utNode := node.GetUserTask()
+	assert.NotNil(t, utNode)
+
+	assert.Equal(t, "no-response", utNode.GetOnCancelExceptionName().GetLiteralValue().GetStr())
```
coltmcnealy-lh, line 100: :heart:

sdk-go/wflib/wf_lib_public.go
```diff
@@ -281,3 +289,14 @@ func (u *UserTaskOutput) WithNotes(notes interface{}) *UserTaskOutput {
 	userTaskNode.Notes = notesVar
 	return u
 }
+
+func (u *UserTaskOutput) WithOnCancelException(exceptionName interface{}) *UserTaskOutput {
```
coltmcnealy-lh, line 293: I see you, writing Java in Go :smile: I like this.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/UserTaskOutput.java
```diff
@@ -6,4 +6,8 @@ public interface UserTaskOutput extends NodeOutput {
     public UserTaskOutput withNotes(WfRunVariable notes);
 
     public UserTaskOutput withNotes(LHFormatString notes);
+
+    UserTaskOutput withOnCancelException(String exceptionName);
```
coltmcnealy-lh, line 10: Nit pick: `withOnCancellationException`. That's slightly more semantic, even though it's longer to type.

coltmcnealy-lh (approved):
This PR is a masterpiece. Well done!

PR #758, opened 2024-04-13
Add persistence to the metronome

sauljabin:
WIP

1. execute a run wf
2. save the id into rocks db
3. in a thread pool  to query the rocks db instance
4. execute get wf for each if saved into rocks db
5. if an attempt fails increase the counter
6. if attempt success delete the id from rocks db and emit a metric beat with **completed**
7. if the attempts get exhausted delete the id from rocks db and emit a metric beat with **error**

Pros:

1. the metronome knows the get wfs pending to be executed after a restart
2. the metronome reduce the possibility of loosing data, because it is saving it on disk
3. it is not necessary to consume form kafka, reducing traffic

Cons

1. thread management for accessing to rocks db
2. we need a statefulset and a pvc


canary/build.gradle
```diff
@@ -37,6 +37,9 @@ dependencies {
     // Prometheus
     implementation 'io.micrometer:micrometer-registry-prometheus:1.12.2'
 
+    // RocksDB
+    implementation 'org.rocksdb:rocksdbjni:9.0.0'
```
coltmcnealy-lh, line 41: Interesting, Kafka Streams is still on the 7.x.x I think. Tell Matthias to hurry up and upgrade!

canary/src/main/java/io/littlehorse/canary/metronome/MetronomeGetWfExecutor.java
```diff
@@ -0,0 +1,11 @@
+package io.littlehorse.canary.metronome;
+
+import io.littlehorse.canary.metronome.internal.LocalRepository;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class MetronomeGetWfExecutor {
+    public MetronomeGetWfExecutor(final LocalRepository repository) {
+        log.info("GetWf Metronome Started");
+    }
```
coltmcnealy-lh, line 10: Assuming that this is the "Work In Progress" part? And this will check everything in the RocksDB, and keep doing a `GetWfRun` until it is completed?

coltmcnealy-lh:
Looks good, this was very clever and I would not have come up with this idea on my own. This is definitely headed in the right direction.

This was a very difficult challenge. Good job!

coltmcnealy-lh:
I have two feature requests:

> until the attempts get exhausted

Would we be able to keep retrying for 3+ hours? I would imagine that if there is a really big cluster and there's a state restoration, we could have a workflow that takes 3+ hours or longer to complete because it is blocked by restoration.

Now, on that note, we might want to send intermediate beats saying "uh oh, this workflow is taking forever" so that we can be alerted of it and take action. Otherwise, we will only find out after the problem is resolved.

sauljabin:
> I have two feature requests:
> 
> > until the attempts get exhausted
> 
> Would we be able to keep retrying for 3+ hours? I would imagine that if there is a really big cluster and there's a state restoration, we could have a workflow that takes 3+ hours or longer to complete because it is blocked by restoration.
> 
> Now, on that note, we might want to send intermediate beats saying "uh oh, this workflow is taking forever" so that we can be alerted of it and take action. Otherwise, we will only find out after the problem is resolved.

When rebalancing **lhClient.runWf()** would throws exception? right?

canary/src/main/java/io/littlehorse/canary/metronome/MetronomeGetWfExecutor.java
sauljabin: yes, WIP

sauljabin:
> I have two feature requests:
> 
> > until the attempts get exhausted
> 
> Would we be able to keep retrying for 3+ hours? I would imagine that if there is a really big cluster and there's a state restoration, we could have a workflow that takes 3+ hours or longer to complete because it is blocked by restoration.
> 
> Now, on that note, we might want to send intermediate beats saying "uh oh, this workflow is taking forever" so that we can be alerted of it and take action. Otherwise, we will only find out after the problem is resolved.

I could add another metric for "uh oh, this workflow is taking forever"

canary/src/main/java/io/littlehorse/canary/metronome/MetronomeGetWfRunExecutor.java
```diff
@@ -0,0 +1,110 @@
...
+    private void scheduledRun() {
+        log.trace("Executing GetWfRun metronome");
+        final Instant searchCriteria = Instant.now().minus(Duration.ofMinutes(1));
+        final Map<String, Attempt> attempts = repository.getAttemptsBefore(searchCriteria);
+        attempts.forEach(this::executeRun);
+    }
+
+    private void executeRun(final String id, final Attempt attempt) {
+        if (attempt.getAttempt() >= retries) {
+            repository.delete(id);
```
coltmcnealy-lh, line 67: Can we make some form of notification or some beat saying "workflow totally failed"?

PR #760, opened 2024-04-15
use proper partition key for parent wf run

eduwercamacaro:


coltmcnealy-lh (approved):
Much better

server/src/main/java/io/littlehorse/server/streams/BackendInternalComms.java
```diff
@@ -696,7 +696,7 @@ private InternalScanResponse objectIdPrefixScanOnThisHost(InternalScan req) {
         }
         String bookmarkKey = null;
         boolean brokenBecauseOutOfData = true;
-
+        System.out.println("partition key " + req.partitionKey);
```
mijailr, line 699: Is this intended?

mijailr (approved):

PR #762, opened 2024-04-17
Changes commit.interval for COMMIT_INTERVAL_MS_CONFIG constant

MateoRojas:


coltmcnealy-lh (approved):
Thank you and good catch!

server/src/main/java/io/littlehorse/common/LHServerConfig.java
```diff
@@ -751,7 +752,7 @@ public Properties getCoreStreamsConfig() {
         //
         // That's not to mention that we will be writing fewer times to RocksDB. Huge win.
         int commitInterval = Integer.valueOf(getOrSetDefault(LHServerConfig.CORE_STREAMS_COMMIT_INTERVAL_KEY, "2000"));
-        props.put("commit.interval", commitInterval);
+        props.put(StreamsConfig.COMMIT_INTERVAL_MS_CONFIG, commitInterval);
```
MateoRojas, line 755: I intentionally used constants from the Kafka client library to avoid any typos. Constants are great to avoid mistakes. One drawback is that it might not be as easy to understand what config they are referring to.

coltmcnealy-lh: Yeah we should do this everywhere

Issue #732, opened 2024-03-22
Per-Thread Exception Handlers on WaitForThreadsNode in Go


coltmcnealy-lh:
It has been implemented in Java and Go. We need to copy PR #730  into GoLang.

PR #766, opened 2024-04-22
732 golang per thread exn handlers

coltmcnealy-lh:
Implements #732


sdk-go/wflib/wf_lib_internal.go
```diff
@@ -558,6 +558,94 @@ func (n *NodeOutput) jsonPathImpl(path string) NodeOutput {
...
+		n.thread.throwError(errors.New("can only call handleExceptionOnChild on WaitForThreads Node"))
+	}
+
+	var threadName string
+	if exceptionName != nil {
+		threadName = "exn-handler-" + n.nodeName + "-" + *exceptionName
+	} else {
+		threadName = "exn-handler-" + n.nodeName
+	}
+	threadName = n.thread.wf.addSubThread(threadName, handler)
```
mijailr, line 574: This is a perfect case to follow the rule of three, a refactor might be a good thing to do since all three follow the same pattern.

mijailr (approved):
Looks good, just a minor refactor request.

sdk-go/wflib/wf_lib_internal.go
coltmcnealy-lh: Tried it...it's actually pretty complex to get it right for all three cases. I think it's more readable this way.

PR #772, opened 2024-04-26
feat(nodeRun): paginate list nodeRun

### LINKED ISSUE #747: Pagination for 'rpc ListNodeRun'
### Context

The current implementation doesn't provides pagination which could cause issues if there are more than 200 NodeRuns. 

### Acceptance Criteria

**Scenario** a call to ListNodeRuns with `limit`
**Given** a rpc call to ListNodeRuns 
**When** `limit` is defined to `10`
**Then** it should return 10 NodeRuns and the corresponding `bookmark` for the next page.

### Out of Scope

_No response_

### Components

Server

mijailr:
This adds:
- `limit` and `bookmark` to listNodeRuns
- `threadRunNumber` to filter nodeRuns on a specific thread

Closes #747


lhctl/cmd/node_run.go
```diff
@@ -62,22 +62,43 @@ var getNodeRunCmd = &cobra.Command{
...
+
+		if len(args) == 2 {
+			threadNumber, err := strconv.Atoi(args[1])
+			if err != nil {
+				log.Fatal("Couldn't convert threadRunNumber to int.")
+			}
+			trn = threadNumber
+		}
 
+		threadRunNumber := int32(trn)
```
mijailr, line 96: I'm not sure about this, I'll test if this is defaulting to zero.

lhctl/cmd/node_run.go
```diff
@@ -62,22 +62,43 @@ var getNodeRunCmd = &cobra.Command{
...
+	Args: func(cmd *cobra.Command, args []string) error {
+		if err := cobra.MinimumNArgs(1)(cmd, args); err != nil {
+			return err
+		}
+
+		if err := cobra.MaximumNArgs(2)(cmd, args); err != nil {
+			return err
+		}
+
+		return nil
```
mijailr, line 76: This handles the argument count, so there's no need to handle in the `Run` function.

dashboard/apps/web/littlehorse-public-api/acls.ts
```diff
@@ -698,7 +698,7 @@ export const PutPrincipalRequest = {
     if (message.globalAcls !== undefined) {
       ServerACLs.encode(message.globalAcls, writer.uint32(26).fork()).ldelim();
     }
-    if (message.overwrite === true) {
+    if (message.overwrite !== false) {
```
mijailr, line 701: I'm not sure why `compile-proto` is changing this

coltmcnealy-lh: Yeah, it might be worth looking.

lhctl/cmd/node_run.go
coltmcnealy-lh: Much better. Thanks

lhctl/cmd/node_run.go
coltmcnealy-lh: In Go, it defaults to `0`. However, the behavior should be that it passes `nil` if you don't specify `--thread-run-number`, thereby allowing you to list all NodeRun's for a particular WfRun.

PR #782, opened 2024-05-07
fix(proto): remove proto build for dashboard

mijailr:
This was replaced with `sdk-js` so there's no need to compile proto for the dashboard specifically.


jhosep-marin (approved):
It looks good and now compile-proto is working as expected

coltmcnealy-lh (approved):
Thanks!

PR #784, opened 2024-05-07
sdk-java: Rewrite server connection manager

eduwercamacaro:


sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/LHTaskExecutor.java
```diff
@@ -0,0 +1,238 @@
...
+            specificStub.reportTask(result, new ReportTaskObserver(result, TOTAL_RETRIES, specificStub));
+            log.debug("Successfully contacted LHServer on reportTask for wfRun {}", wfRunId);
+        } catch (Exception exn) {
+            log.warn("Failed to report task for wfRun {}: {}", wfRunId, exn.getMessage());
+            scheduleRetry(result, specificStub, TOTAL_RETRIES);
+        }
+    }
+
+    private void scheduleRetry(ReportTaskRun reportedTaskRun, LittleHorseStub stub, int retriesLeft) {
+        pool.schedule(
```
eduwercamacaro, line 108: see:
https://github.com/littlehorse-enterprises/littlehorse/pull/784/files#diff-ca313014916ffeb554136f07d5c209c4c0d0f7d8e31975887c413af280637c51L204

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/LHServerConnectionManager.java
```diff
@@ -1,238 +1,44 @@
...
-
-    private static final long HEARTBEAT_INTERVAL_MS = 5000L;
+public class LHServerConnectionManager {
 
+    private final RebalanceThread rebalanceThread;
+    private final TaskDef taskDef;
     private final LHLivenessController livenessController;
-
-    private static final int TOTAL_RETRIES = 5;
+    private static final long HEARTBEAT_INTERVAL_MS = 5000L;
```
coltmcnealy-lh, line 16: While we are already rewriting everything—can we move this up to like 25 or 30 seconds?

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/LHTaskExecutor.java
```diff
@@ -0,0 +1,238 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Manages a queue and executor for scheduling and executing tasks in memory.
+ * Future versions will share the same queue for many workers.
+ * It handles retries with a specified delay.
```
coltmcnealy-lh, line 34: Could you please update this comment to be specific about what you mean by "retries"?

What I **think** this means is retries on the `rpc ReportTask`. Why do we need that?

Let's think about what we have going on here:
- We have a "bootstrap stub" which has access to all the servers.
- We have a "specific stub" which has access to one specific LH Server.
- When we do a `rpc ReportTask`, we use the Specific Stub so that it goes directly to the server that will process it. This saves us some network hops and is more performant.
- If the specific server is down, we retry with the Bootstrap Stub which has access to all the servers. This is useful for increasing availability during a server rolling restart.

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/LHTaskExecutor.java
```diff
@@ -0,0 +1,238 @@
...
+    private final ScheduledExecutorService pool;
+    private final Semaphore semaphore;
+    private static final int TOTAL_RETRIES = 5;
+
+    public LHTaskExecutor(int concurrency) {
+        this.pool = Executors.newScheduledThreadPool(concurrency);
+        this.semaphore = new Semaphore(concurrency);
+        Thread closeExecutorHook = new Thread(() -> {
+            try {
+                this.close(4, TimeUnit.SECONDS);
```
coltmcnealy-lh, line 52: Perhaps we may want to make this longer—like 15 seconds—in case a task method takes a long time to execute. In the future it should become a configuration. Maybe add a GH Issue for that?

eduwercamacaro: Yes, I can create a GH issue for this.

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/LHTaskExecutor.java
```diff
@@ -0,0 +1,238 @@
...
+     * Schedule a new task to be executed
+     */
+    public void submitTaskForExecution(
+            ScheduledTask scheduledTask,
+            LittleHorseStub specificStub,
+            List<VariableMapping> mappings,
+            Object executable,
+            Method taskMethod) {
+        try {
+            this.semaphore.acquire();
```
coltmcnealy-lh, line 70: This looks like we have already gotten the ScheduledTask _before_ we wait for the Semaphore. That would make us vulnerable to a lot of `TIMEOUT`s when we have a slow LHTaskMethod and a lot of tasks in the queue.

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/LHTaskExecutor.java
```diff
@@ -0,0 +1,238 @@
...
+                scheduledTask, LHLibUtil.fromProtoTs(scheduledTask.getCreatedAt()), mappings, executable, taskMethod);
+        this.semaphore.release();
+        String wfRunId = LHLibUtil.getWfRunId(scheduledTask.getSource()).getId();
+        try {
+            log.debug("Going to report task for wfRun {}", wfRunId);
+            specificStub.reportTask(result, new ReportTaskObserver(result, TOTAL_RETRIES, specificStub));
+            log.debug("Successfully contacted LHServer on reportTask for wfRun {}", wfRunId);
+        } catch (Exception exn) {
+            log.warn("Failed to report task for wfRun {}: {}", wfRunId, exn.getMessage());
+            scheduleRetry(result, specificStub, TOTAL_RETRIES);
```
coltmcnealy-lh, line 103: We should not schedule retries with the specific stub but rather with the `bootstrap` stub. The GRPC clients can do retries automatically themselves, but the problem is if we are doing a rolling bounce of the servers, the Specific Server will be down for about 45 seconds. That means the Specific Stub won't work for 45 seconds. We should use the Bootstrap Stub which balances across all servers.

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/LHTaskExecutor.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        @Override
+        public void onNext(Empty value) {
+            // nothing to do.. maybe metrics?
+        }
+
+        @Override
+        public void onError(Throwable t) {
+            scheduleRetry(reportedTaskRun, stub, retriesLeft);
```
coltmcnealy-lh, line 232: This place should also use the bootstrap stub

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/PollConnection.java
```diff
@@ -0,0 +1,116 @@
...
+        } else {
+            // This may cause issues when there's multiple threads, eg closing an
+            // already closed stream observer.
+            pollClient.onCompleted();
+        }
+    }
+
+    private void askForMoreWork() {
+        log.debug("Asking for more work on {}:{}", host.getHost(), host.getPort());
+        pollClient.onNext(PollTaskRequest.newBuilder()
```
coltmcnealy-lh, line 96: The call to `semaphore.acquire()` should be here and not above

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/LHTaskExecutor.java
eduwercamacaro: > We have a "bootstrap stub" which has access to all the servers.
That's not always the case, right? It depends on infrastructure decisions.

coltmcnealy-lh (approved):
New changes look good!

PR #796, opened 2024-05-13
Introducing PollThread

eduwercamacaro:


sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
```diff
@@ -403,6 +403,6 @@ protected String[] getEnvKeyPrefixes() {
      * @return the number of worker threads to run.
      */
     public int getWorkerThreads() {
-        return Integer.valueOf(getOrSetDefault(NUM_WORKER_THREADS_KEY, "8"));
+        return Integer.valueOf(getOrSetDefault(NUM_WORKER_THREADS_KEY, "2"));
```
coltmcnealy-lh, line 406: Can you do one of two things:

1. Update the other SDK's and update the docs with the new default value, OR
2. Leave this at `8`

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/PollThread.java
```diff
@@ -10,85 +10,99 @@
...
-    public void acquire() {
+    @Override
+    public void run() {
         try {
-            this.semaphore.acquire();
-        } catch (InterruptedException exn) {
-            throw new RuntimeException(exn);
+            while (running) {
+                semaphore.acquire();
+                pollClient.onNext(PollTaskRequest.newBuilder()
```
coltmcnealy-lh, line 73: Good use of semaphore here :smile:

PR #797, opened 2024-05-13
787 - Enhancements for the User Task Runs

diablouma:
This PR includes:
 - Who is it currently assigned to? (user_group and/or user_id)_
 - The "Audit Log" which is the events field of the UserTaskRun
 - Assignment history and/or cancellation history
 - If it has a specific EXCEPTION that it throws upon cancellation
 - The results after it is completed
 - When the UserTaskRun was created (i.e. arrival time)
 - When the UserTaskRun was completed or cancelled (if it isn't still pending)


dashboard/src/app/(authenticated)/(diagram)/components/Modals/UserTaskRun.tsx
```diff
@@ -0,0 +1,161 @@
...
+import { useModal } from '@/app/(authenticated)/(diagram)/hooks/useModal'
+import { Dialog, DialogPanel, DialogTitle } from '@headlessui/react'
+import { DialogBody } from 'next/dist/client/components/react-dev-overlay/internal/components/Dialog'
+import { XMarkIcon } from '@heroicons/react/24/outline'
+import { getVariable, getVariableValue } from '@/app/utils'
+import { UserTaskRunStatus } from 'littlehorse-client/dist/proto/user_tasks'
+
+export const UserTaskRun: FC<UserTaskModal> = ({ data, nodeRun, userTaskNode }) => {
+  const { showModal, setShowModal } = useModal()
+  const refDiv = useRef(null)
```
mijailr, line 12: I wonder why is this necessary if TaskRun doesn't need it?

dashboard/src/app/(authenticated)/(diagram)/components/NodeTypes/UserTask/UserTask.tsx
```diff
@@ -26,11 +29,10 @@ const Node: FC<NodeProps> = ({ data, selected }) => {
             </Link>
           </div>
           <div className="mb-2 flex gap-2 text-nowrap">
-            {userTask.userGroup && (
-              <div className="flex items-center justify-center">Group: {getVariable(userTask.userGroup)}</div>
-            )}
-            {userTask.userId && (
-              <div className="flex items-center justify-center">User: {getVariable(userTask.userId)}</div>
+            {nodeRun ? (
+              <UserTaskRunDetails userTaskNode={userTask} nodeRun={nodeRun} />
```
mijailr, line 33: This component should handle `nodeRun` with `undefined` value. Please handle the case on a parent component to simplify this one.

dashboard/src/app/(authenticated)/(diagram)/components/NodeTypes/UserTask/UserTaskDefDetails.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { getVariable } from '@/app/utils'
+
+export const UserTaskDefDetails: FC<{ userTask: UserTaskNode }> = ({ userTask }) => {
+  return (
+    <>
+      {userTask.userGroup && (
+        <div className="flex items-center justify-center">Group: {getVariable(userTask.userGroup)}</div>
+      )}
+
+      {userTask.userId && <div className="flex items-center justify-center">User: {getVariable(userTask.userId)}</div>}
```
mijailr, line 12: This content is similar to `UserTaskRunDetails` can it be a shared component instead?

mijailr (approved):
Looks good, just some optional comments.

coltmcnealy-lh:
We could use a new ticket that differentiates what we show on the `WfSpec` from the `WfRun`

dashboard/src/app/(authenticated)/(diagram)/components/Modals/UserTaskRun.tsx
diablouma: I was needing certain properties like nodeRun.endTime and the cancellationException which are not part of the UserTask object itself.

dashboard/src/app/(authenticated)/(diagram)/components/NodeTypes/UserTask/UserTaskDefDetails.tsx
diablouma: extracted :)

dashboard/src/app/(authenticated)/(diagram)/components/NodeTypes/UserTask/UserTask.tsx
diablouma: Mijail are you saying that I should have a component that receives the nodeRun and handle both cases internally instead of handling them at the UserTask.tsx?

PR #813, opened 2024-05-17
updates site for 0.9.0

coltmcnealy-lh:


coltmcnealy-lh:
Appears that the `lh-dashboard` tests don't get triggered on changes to the `docs`. But it's a required check. @mijailr what do you propose here?

eduwercamacaro (approved):
LGTM

PR #820, opened 2024-05-20
Make Canary Deployable

coltmcnealy-lh:
- Should not initialize Metronome stuff if Metronome not enabled
- Kafka configuration were ignoring anything in the SslConfig.java class
- Kafka configuration were ignoring anything in the CommonClientConfig.java class
  - Unfortunately, I had to remove those classes and just pass anything with a '.kafka' prefix iinto all Kafka constructors


canary/src/main/java/io/littlehorse/canary/config/KafkaConfig.java
```diff
@@ -2,14 +2,14 @@
...
 
-public class KafkaConsumerConfig implements Config {
+public class KafkaConfig implements Config {
     private final Map<String, Object> configs;
 
-    public KafkaConsumerConfig(final Map<String, Object> configs) {
+    public KafkaConfig(final Map<String, Object> configs) {
         this.configs = configs.entrySet().stream()
-                .filter(entry -> ConsumerConfig.configNames().contains(entry.getKey()))
+                .filter(entry -> entry.getKey().startsWith("kafka."))
```
coltmcnealy-lh, line 11: I realized that there is a better way to do this. Rather than deleting the other files and using the `kafka.` prefix, we can make the `.filter()` also check for:
- `ConsumerConfig`
- `CommonClientConfig`
- `SslConfig`

It was ignoring `CommonClientConfig` and `SslConfig`. I think that smaller solution is much better.

eduwercamacaro (approved):

canary/src/main/java/io/littlehorse/canary/config/KafkaConfig.java
coltmcnealy-lh: Hmm, this might be something we optimize for later.

mijailr:
Should this deployable generate an artifact? Docker image perhaps?

PR #821, opened 2024-05-20
788 search user task run

diablouma:
On the UserTaskDef page, we should be able to search for UserTaskRun's of that type. We should be able to filter by:

* The status
* Assigned userId
* Assigned userGroup
* Created time
* Clicking on the result should take you to the relevant WfRun and also highlight the appropriate NodeRun on that WfRun.

### DIFF
--- a/dashboard/src/app/(authenticated)/userTaskDef/[...props]/components/UserTaskDef.tsx
+++ b/dashboard/src/app/(authenticated)/userTaskDef/[...props]/components/UserTaskDef.tsx
@@ -24,7 +24,6 @@ export const UserTaskDef: FC<Props> = ({ spec }) => {
     UserTaskRunStatus.CANCELLED,
     UserTaskRunStatus.DONE,
     UserTaskRunStatus.UNASSIGNED,
-    UserTaskRunStatus.UNRECOGNIZED,
   ]
   const [selectedStatus, setSelectedStatus] = useState(UserTaskRunStatus.UNASSIGNED)
   const { tenantId } = useWhoAmI()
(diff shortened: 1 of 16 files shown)


dashboard/src/app/(authenticated)/userTaskDef/[...props]/components/UserTaskDef.tsx
```diff
@@ -1,18 +1,214 @@
...
+  ]
+  const [selectedStatus, setSelectedStatus] = useState(UserTaskRunStatus.UNASSIGNED)
+  const [userId, setUserId] = useState('')
+  const [userGroup, setUserGroup] = useState('')
+  const [createdAfter, setCreatedAfter] = useState('')
+  const [createdBefore, setCreatedBefore] = useState('')
+  const { tenantId } = useWhoAmI()
+  const [limit, setLimit] = useState<number>(SEARCH_DEFAULT_LIMIT)
+  const [userIdToSearchFor] = useDebounce(userId, 1000)
+  const [userGroupToSearchFor] = useDebounce(userGroup, 1000)
```
Comment (mijailr) [inline dashboard/src/app/(authenticated)/userTaskDef/[...props]/components/UserTaskDef.tsx:40]: Why is useDebounce used here, what's the effect that it prevents if it's just a second?

Also `1000` is the delay, can it be a constant instead?

mijailr (approved):
Looks good, just a comment about useDebounce.

dashboard/src/app/(authenticated)/userTaskDef/[...props]/components/UserTaskDef.tsx
Comment (diablouma) [reply dashboard/src/app/(authenticated)/userTaskDef/[...props]/components/UserTaskDef.tsx:40]: it prevents to make server calls after every letter that the user types on the search boxes, with the debounce it will wait 1 second after the last typed character to call the server.
Yep I can extract that to a constant :)

PR #823, opened 2024-05-21
feat(dashboard): Starts accepting LHC_API_PROTOCOL configuration

MateoRojas:
`LHC_API_PROTOCOL` is required to be able to use a secure communication between the dashboard and the lh-server without having to enable authentication


mijailr (approved):

mijailr:
By default it assumes that when authentication is enabled it will be `SSL` otherwise it will be `PLAINTEXT` is that still accurate? 

For context, that assertion was inherited from old dashboard.

coltmcnealy-lh (approved):

mijailr (approved):

HazimAr (approved):

PR #824, opened 2024-05-21
fix/dashboard-ui-cleanup

HazimAr:


dashboard/src/app/(authenticated)/components/Principal.tsx
```diff
@@ -9,44 +9,46 @@ function classNames(...classes: Array<string | boolean>) {
...
   return (
-    <Menu as="div" className="relative inline-block text-left">
-      <div>
-        <Menu.Button className="inline-flex w-full justify-center gap-x-1.5 px-3 py-2 text-sm font-semibold text-gray-900 shadow-sm">
-          <div className="relative inline-flex h-8 w-8 items-center justify-center overflow-hidden rounded-full bg-gray-100 dark:bg-gray-600">
-            <span className="font-medium text-gray-600 dark:text-gray-300">{user?.name?.at(0)}</span>
-          </div>
-        </Menu.Button>
-      </div>
+    user?.name != 'anonymous' && (
```
mijailr, line 12: This is interesting, although `anonymous` is a valid `Principal` name so instead of checking the username, check if there's a session or if authentication is enabled.

coltmcnealy-lh (approved):

Issue #826, opened 2024-05-21
Java task worker does not reconnect to server after unexpected server restart


Snarr:
### Context

When a Java task worker is connected to a server and the server unexpectedly restarts, the worker recognizes that the gRPC connection closed but fails to reconnect to the server. This makes the worker unusable until the worker itself restarts. In this case I was able to produce the error when running the `example-basic` workflow example that creates a `wfSpec` with the same name.

```
14:02:26 ERROR [LH] PollThread - Unexpected error from server
io.grpc.StatusRuntimeException: CANCELLED: RST_STREAM closed stream. HTTP/2 error code: CANCEL
        at io.grpc.Status.asRuntimeException(Status.java:539) ~[grpc-api-1.56.1.jar:1.56.1]
        at io.grpc.stub.ClientCalls$StreamObserverToCallListenerAdapter.onClose(ClientCalls.java:491) [grpc-stub-1.56.1.jar:1.56.1]
        at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:567) [grpc-core-1.56.1.jar:1.56.1]
        at io.grpc.internal.ClientCallImpl.access$300(ClientCallImpl.java:71) [grpc-core-1.56.1.jar:1.56.1]
        at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInternal(ClientCallImpl.java:735) [grpc-core-1.56.1.jar:1.56.1]
        at io.grpc.internal.ClientCallImpl$ClientStreamListenerImpl$1StreamClosed.runInContext(ClientCallImpl.java:716) [grpc-core-1.56.1.jar:1.56.1]
        at io.grpc.internal.ContextRunnable.run(ContextRunnable.java:37) [grpc-core-1.56.1.jar:1.56.1]
        at io.grpc.internal.SerializingExecutor.run(SerializingExecutor.java:133) [grpc-core-1.56.1.jar:1.56.1]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) [?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) [?:?]
        at java.lang.Thread.run(Thread.java:842) [?:?]
14:02:26 ERROR [LH] PollThread - Unexpected error from server
[...]

PR #827, opened 2024-05-21
fix(java-sdk): Detect if poll thread stops running and re-establish connection

Snarr:
When PollThreads stop running for any reason, a private boolean in the PollThread class `stillRunning` is set to false. Until now, this variable wasn't being used in the RebalanceThread, so when PollThreads stopped running they would never start up again.

This pull request attempts to remedy this by checking the state of the PollThread in the RebalanceThread. If the PollThread `isRunning` variable is true, it is kept. If the PollThread `isRunning` variable is false, it is replaced with a new PollThread.

Refs: #826

### DIFF
--- a/sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/RebalanceThread.java
+++ b/sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/RebalanceThread.java
@@ -104,11 +104,12 @@ final class RebalanceThread extends Thread {
                         removed.close();
                     }
                 }
-                LittleHorseGrpc.LittleHorseStub stub = config.getAsyncStub(runningConnection.getHost(),
-                        runningConnection.getPort());
+                LittleHorseGrpc.LittleHorseStub stub =
+                        config.getAsyncStub(runningConnection.getHost(), runningConnection.getPort());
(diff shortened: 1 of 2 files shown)


sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/RebalanceThread.java
```diff
@@ -104,6 +104,23 @@ public void onNext(RegisterTaskWorkerResponse response) {
                         removed.close();
                     }
                 }
+                LittleHorseGrpc.LittleHorseStub stub =
+                        config.getAsyncStub(runningConnection.getHost(), runningConnection.getPort());
+                // This loop replaces each PollThread that stops running with a fresh PollThread
+                for (int i = 0; i < runningConnections.get(runningConnection).size(); i++) {
+                    PollThread existingPollThread =
+                            runningConnections.get(runningConnection).get(i);
```
eduwercamacaro, line 112: [Suggestion]
Keep the number of gets as low as possible because this is a `ConcurrentHashMap` and might block other threads.

eduwercamacaro (approved):

eduwercamacaro (approved):

PR #829, opened 2024-05-24
feat(dashboard): taskRun Table

HazimAr:


HazimAr:
I will rewrite this with the new decided frontend design system. I just wanted to merge what I already wrote.

HazimAr:
fixed some minor bugs, ready for review now.

HazimAr:
Currently when clicking on the link for the taskRun it will not highlight the node as you need to make another request to get the nodeName 😭. I will worry about this when rewriting the table to use the new frontend decisions we made.

diablouma:
> Currently when clicking on the link for the taskRun it will not highlight the node as you need to make another request to get the nodeName 😭. I will worry about this when rewriting the table to use the new frontend decisions we made.

@HazimAr if we are going to split the task in 2, please could you create a new issue for that particular enhancement? so we don't lose track. I think you should take that work as soon as this ticket is done.

dashboard/src/app/(authenticated)/(diagram)/components/Diagram.tsx
```diff
@@ -49,7 +49,7 @@ export const Diagram: FC<Props> = ({ spec, wfRun }) => {
 
   const searchParams: ReadonlyURLSearchParams = useSearchParams()
   const threadRunNumberFromRedirection: number = Number(searchParams.get('threadRunNumber'))
-  const nodeRunNameToBeHighlighted: string = searchParams.get('nodeRunName')!
+  const nodeNameToBeHighlighted: string = searchParams.get('nodeName')!
```
diablouma, line 52: I'd prefer to keep the "Run" word here, because node and nodeRun are not the same thing :)

dashboard/src/app/(authenticated)/taskDef/[name]/actions/searchTaskRun.ts
```diff
@@ -0,0 +1,53 @@
...
+      const nodeRun = await client.getNodeRun(taskRun.id!)
+
+      return {
+        taskRun,
+        nodeRun,
+      }
+    })
+  }
+
+  const userTaskRunWithDetails: runDetails[] = await Promise.all(hydrateWithTaskRunDetails())
```
Comment (diablouma) [inline dashboard/src/app/(authenticated)/taskDef/[name]/actions/searchTaskRun.ts:46]: it should be only taskRun, not userTaskRun

dashboard/src/app/(authenticated)/taskDef/[name]/components/TaskDef.tsx
```diff
@@ -1,18 +1,152 @@
...
+              </tr>
+            </thead>
+            <tbody>
+              {data?.pages.map((page, i) => (
+                <Fragment key={i}>
+                  {page.resultsWithDetails.length > 0 ? (
+                    page.resultsWithDetails.map(({ taskRun, nodeRun }) => {
+                      return (
+                        <tr key={taskRun.id?.taskGuid} className="border-b border-neutral-200">
+                          <td className="px-6 py-4">😡
```
Comment (mijailr) [inline dashboard/src/app/(authenticated)/taskDef/[name]/components/TaskDef.tsx:112]: Is the emoji intended?

dashboard/src/app/(authenticated)/taskDef/[name]/components/TaskDef.tsx
```diff
@@ -1,18 +1,152 @@
...
+                            <Link
+                              className="py-2 text-blue-500 hover:underline"
+                              target="_blank"
+                              //? Will have to create a new request to get the nodeName to select the node in the diagram 😡 (will do this in the rewrite)
```
Comment (diablouma) [inline dashboard/src/app/(authenticated)/taskDef/[name]/components/TaskDef.tsx:116]: left a comment on top of your comment on the PR. Please I think we can remove this comments :)

dashboard/src/app/(authenticated)/(diagram)/components/Diagram.tsx
mijailr: True, good catch

dashboard/src/app/(authenticated)/taskDef/[name]/components/TaskDef.tsx
```diff
@@ -1,18 +1,152 @@
...
+                    </tr>
+                  )}
+                </Fragment>
+              ))}
+            </tbody>
+          </table>
+        </div>
+      )}
+
+      {/* <div className="mt-6">
```
Comment (diablouma) [inline dashboard/src/app/(authenticated)/taskDef/[name]/components/TaskDef.tsx:142]: If not used,  think we can remove this code

HazimAr:
> > Currently when clicking on the link for the taskRun it will not highlight the node as you need to make another request to get the nodeName 😭. I will worry about this when rewriting the table to use the new frontend decisions we made.
> 
> @HazimAr if we are going to split the task in 2, please could you create a new issue for that particular enhancement? so we don't lose track. I think you should take that work as soon as this ticket is done.

will do

diablouma (approved):

PR #830, opened 2024-05-24
build(deps): bump version

HazimAr:


coltmcnealy-lh (approved):
When merging, just make sure to note that this only affects the dashboard

Issue #818, opened 2024-05-18
Add `rpc GetTenant`


coltmcnealy-lh:
We should add the following `rpc` call:

```
rpc GetTenant(TenantId) returns (Tenant) {}
```

It should require the `READ` permission over the `ACL_TENANT` resource.

coltmcnealy-lh:
Implemented by #819

PR #832, opened 2024-05-27
feat(server): allow principal deletion #818

coltmcnealy-lh:
- Adds 'rpc DeletePrincipal'
- Fixes bug in the OAuthServerAuthorizer which incorrectly uses userName() on tokens from machine clients


schemas/internal/command.proto
```diff
@@ -39,6 +39,7 @@ message Command {
         CancelUserTaskRunRequest cancel_user_task = 19;
         BulkUpdateJob bulk_job = 20;
         TaskAttemptRetryReady task_attempt_retry_ready = 21;
+        DeletePrincipalRequest delete_principal = 22;
```
eduwercamacaro, line 42: `DeletePrincipalRequest` is an external request. why it needs to be in this proto file?

server/src/main/java/io/littlehorse/server/streams/topology/core/MetadataCommandExecution.java
```diff
@@ -81,4 +87,17 @@ private AuthorizationContext authContextFor(TenantIdModel tenantId, PrincipalIdM
         // fine-grained authorization
         return new AuthorizationContextImpl(principalId, tenantId, List.of(), false);
     }
+
+    public Collection<PrincipalIdModel> adminPrincipalIds() {
+        Attribute tagAdminAttribute = new Attribute("isAdmin", Boolean.TRUE.toString());
+        Set<PrincipalIdModel> result = this.metadataManager()
+                .tagScan(PrincipalModel.getTypeEnum(PrincipalModel.class), List.of(tagAdminAttribute))
```
eduwercamacaro, line 94: `.tagScan` should be deprecated now. you can either modify the tagScan method in order to close the iterator or use other mechanism to execute the range scan

examples/basic/src/main/java/io/littlehorse/examples/BasicExample.java
```diff
@@ -52,8 +55,10 @@ public static LHTaskWorker getTaskWorker(LHConfig config) {
 
     public static void main(String[] args) throws IOException {
         // Let's prepare the configurations
-        Properties props = getConfigProps();
-        LHConfig config = new LHConfig(props);
+        // Properties props = getConfigProps();
+        // LHConfig config = new LHConfig(props);
+
+        LHConfig config = new LHConfig("/tmp/worker-my-tenant.config");
```
eduwercamacaro, line 61: I don't think this is intended.

Snarr: Was this changed for testing purposes?

Snarr (approved):

schemas/internal/command.proto
coltmcnealy-lh: Because it's a `SubCommand`. This is the protobuf for a `Command`.

We have other "external requests" which are also `SubCommand`s, like `RunWfRequest` and `DeleteWfRunRequest`.

However, you're right...it's a _METADATA_ command, not a run command, so it shouldn't be here. That was a brain fart.

examples/basic/src/main/java/io/littlehorse/examples/BasicExample.java
coltmcnealy-lh: Yes it was, oops

server/src/main/java/io/littlehorse/server/streams/topology/core/MetadataCommandExecution.java
coltmcnealy-lh: I see, I copied this and pasted it from the other place. Will fix

eduwercamacaro (approved):

PR #843, opened 2024-05-29
test(sdk-java): PollThread unit tests

eduwercamacaro:


sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/ScheduledTaskExecutor.java
Snarr, line 1: Why are we relocating this logic to this file? Just curious. Is it just to split up the complex PollThread class?

Snarr (approved):

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/ScheduledTaskExecutor.java
eduwercamacaro: Exactly. `PollThread` is too complex this PR just split the logic inside `PollThread`.

PR #845, opened 2024-05-29
feat(dashboard): taskRun actual values

HazimAr:
image

### DIFF
--- a/dashboard/src/app/(authenticated)/(diagram)/components/NodeTypes/Task/TaskDetails.tsx
+++ b/dashboard/src/app/(authenticated)/(diagram)/components/NodeTypes/Task/TaskDetails.tsx
@@ -1,8 +1,8 @@
-import { getVariable } from '@/app/utils'
-import { ExternalLinkIcon, EyeIcon } from 'lucide-react'
+import { getVariable, getVariableValue } from '@/app/utils'
 import { useQuery } from '@tanstack/react-query'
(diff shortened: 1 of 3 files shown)


dashboard/src/app/(authenticated)/(diagram)/components/NodeTypes/Task/TaskDetails.tsx
```diff
@@ -53,9 +53,13 @@ export const TaskDetails: FC<{ task?: TaskNode; nodeRun?: NodeRun }> = ({ task,
           <ol className="list-inside list-decimal">
             {task.variables.map((variable, i) => (
               <li className="mb-1 flex gap-1" key={`variable.${i}`}>
-                <div className="bg-gray-200 px-2 font-mono text-fuchsia-500">arg{i}</div>
+                <div className="bg-gray-200 px-2 font-mono text-fuchsia-500">
+                  {nodeRun ? data?.inputVariables[i].varName : `arg${i}`}
```
mijailr, line 57: What if `varName` is null/undefined?

mijailr: Might be better also to use the ternary with `data` since `data` will only be available if `nodeRun` is available

```tsx
{ data ? data... : ... }
```

mijailr (approved):

PR #846, opened 2024-05-30
feat(dashboard): varName of TaskDef in WfSpec

HazimAr:
Before:
image

After:
image


HazimAr:
This is not the correct implementation of getting the taskDef name

HazimAr:
image
Works now.


The only issue is that it makes a call for every single task, there is no straightforward way to fetch the `TaskDef` only when the tooltip is open. We can either keep this PR open or merge this and open an issue to refactor for performance

HazimAr:
> > The only issue is that it makes a call for every TaskDef needed, there is no straightforward way to fetch the TaskDef only when the tooltip is open. We can either keep this PR open or merge this and open an issue to refactor for performance.
> 
> You can prevent this by using the `selected` boolean available at the node, so you make the call only when `selected === true`

Good thinking, I have implemented that.

mijailr (approved):

Issue #836, opened 2024-05-28
Add `rpc SearchPrincipal`


coltmcnealy-lh:
Should follow the standard API patterns of taking in a `SearchPrincipalRequest` and returning a `PrincipalIdList`.

Should be able to search by:
- `isAdmin`
- `tenant`
- `createdAt`

All search values should be optional.

coltmcnealy-lh:
@eduwercamacaro how should we handle permissions here? If I have the following acl's:

```
{
 per_tenant_acls: {
   my-tenant: {
     actions: ALL_ACTIONS,
     resource: ALL_RESOURCES
   }
 }
}
```

should that `Principal` be able to execute the `rpc SearchPrincipal`? There are three options:

1. Return `PERMISSION_DENIED`
2. Return all `Principal`s with `per_tenant_acls` in the `my-tenant` Tenant
3. Return all `Principals` with `per_tenant_acls` in the `my-tenant` Tenant _and also_ any admin Principals

From the perspective of LittleHorse Cloud, I think option 2 is the best.

coltmcnealy-lh:
@Snarr what do you think?

eduwercamacaro:
I think we should go with option 2

Snarr:
After talking it over with Eduwer, I agree option 2 sounds the best.

PR #855, opened 2024-06-03
Add `rpc SearchPrincipal`

Snarr:
This PR adds the `rpc SearchPrincipal` as defined in Issue #836


lhctl/cmd/principal.go
```diff
@@ -148,6 +196,12 @@ func init() {
 	putPrincipalCmd.Flags().Bool("overwrite", false, "Overwrites principal information")
 	putPrincipalCmd.Flags().String("tenantId", "", "Tenant associated with the principal")
 
+	searchCmd.AddCommand(searchPrincipalCmd)
+	searchPrincipalCmd.Flags().String("tenantId", "", "List Principals associated with this Tenant ID")
+	searchPrincipalCmd.Flags().Bool("isAdmin", false, "List only Principals that are admins")
+	searchPrincipalCmd.Flags().Int("earliestMinutesAgo", -1, "Search only for Principals that were created no more than this number of minutes ago")
+	searchPrincipalCmd.Flags().Int("latestMinutesAgo", -1, "Search only for Principals that were created at least this number of minutes ago")
+
```
eduwercamacaro, line 204:
```suggestion
searchPrincipalCmd.MarkFlagsOneRequired("isAdmin", "tenantId")
```

eduwercamacaro: I believe this can replace line 91.
see [Command.MarkFlagsOneRequired](https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsOneRequired)

schemas/service.proto
```diff
@@ -778,6 +781,33 @@ message TenantIdList {
       optional bytes bookmark = 2;
 }
 
+message SearchPrincipalRequest {
```
eduwercamacaro, line 784: Please add documentation at `SearchPrincipalRequest` level

server/src/main/java/io/littlehorse/server/streams/lhinternalscan/publicrequests/SearchPrincipalRequestModel.java
```diff
@@ -0,0 +1,132 @@
...
+
+    @Override
+    public void initFrom(Message proto, ExecutionContext context) throws LHSerdeError {
+        SearchPrincipalRequest p = (SearchPrincipalRequest) proto;
+
+        if (p.hasBookmark()) {
+            try {
+                this.bookmark = BookmarkPb.parseFrom(p.getBookmark());
+            } catch (Exception exn) {
+                log.error("Failed to load bookmark: {}", exn.getMessage(), exn);
```
eduwercamacaro, line 93: I know this is a pattern for our other search classes, but I think we should throw `LHApiException` in this case because users can send an invalid bookmark and still receive the wrong results page. I would like to hear @coltmcnealy-lh 's opinion.

eduwercamacaro (approved):

server/src/main/java/io/littlehorse/server/streams/lhinternalscan/publicrequests/SearchPrincipalRequestModel.java
coltmcnealy-lh: I agree, it should be `LHApiException` with `Status.INVALID_ARGUMENT`

schemas/service.proto
coltmcnealy-lh: Thank you, this shows up on our website :smile:

server/src/main/java/io/littlehorse/server/streams/lhinternalscan/publicrequests/SearchPrincipalRequestModel.java
Snarr: Should we apply this change for all SearchRequestModels?

eduwercamacaro: Yes, we should. but for now, just this class is fine. We can create a separate PR for the other search classes.

PR #856, opened 2024-06-03
fix(dashboard): handle `LHTaskError` properly

HazimAr:
Before:
image

After: 
image


coltmcnealy-lh:
Hmm, @Snarr and @eduwercamacaro we might want to make a ticket to improve the stacktrace on the server side. It shouldn't contain the `InvocationTargetException` but rather only the actual "caused by".

HazimAr:
If error || exception:
image

Else:
image

dashboard/src/app/(authenticated)/(diagram)/components/Modals/TaskRun.tsx
```diff
@@ -1,4 +1,5 @@
 import { getVariableValue } from '@/app/utils'
+import { cn } from '@/lib/utils'
```
mijailr, line 2: Can you assign a better name to this?

dashboard/src/app/(authenticated)/(diagram)/components/Modals/TaskRun.tsx
```diff
@@ -56,11 +59,21 @@ export const TaskRun: FC<Modal> = ({ data }) => {
                   <div className="">{attempt.endTime}</div>
                 </div>
               </div>
-              {attempt.output && (
-                <div className="mt-2 flex flex-col rounded bg-zinc-500 p-1 text-white">
-                  <h3 className="font-bold">Output</h3>
-                  <pre className="overflow-x-auto">{getVariableValue(attempt.output)}</pre>
+
+              {attempt.error || attempt.exception ? (
```
mijailr, line 63: No need to do a ternary here since they are guaranteed to be `oneOf`.

I would prefer the form `attempt.error && ...`

mijailr (changes requested):

dashboard/src/app/(authenticated)/(diagram)/components/Modals/TaskRun.tsx
HazimAr: We can rename the function, but it's commonly used like this, it's a naming convention shad/cn has, and all their examples include it with the name 'cn' as to reference 'shad/cn'

What do you prepose as this is what everyone in shadcn community uses

dashboard/src/app/(authenticated)/(diagram)/components/Modals/TaskRun.tsx
HazimAr: Yea didn't know it was oneOf, just wanted to account for it in case it wasn't. I'll update later

dashboard/src/app/(authenticated)/(diagram)/components/Modals/TaskRun.tsx
mijailr: I see, in that case at least the file name must be different, to reflect what's the scope of that function. `lib/utils` is too generic, this looks like a `classNameHelper` to me.

HazimAr: utils is a file that usually starts off with just cn, but over time will have other function helpers like `classNameHelper` you were referring to this would include stuff like `getVariableValue` but the reason it is in that folder structure is because its is a global helper function, the other global helper utils would reside there. Ultimately, it's up to you, if you would still like to change it, what would you propose?

mijailr (approved):

PR #862, opened 2024-06-05
fix(server): report on standby lag

coltmcnealy-lh:
This PR fixes the issue where the `/status` endpoint wasn't returning an empty list for the standby states.

However, I think there is some other issue. While constantly running 5 workflow's per second (so the changelog topics should have stuff being produced to them), I see the following numbers _never get updated_ which means probably the `onBatchRestored()` isn't getting called...?

```
    "partition": 42,
    "lag": 16960,
    "changelogEndOffset": 34322
```

The `lag` never changes, nor does the `changelogEndOffset`, except on a rebalance.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/monitoring/health/ServerHealthState.java
+++ b/server/src/main/java/io/littlehorse/server/monitoring/health/ServerHealthState.java
@@ -118,7 +118,7 @@ public class ServerHealthState {
                 .lagInfoForPartition(
                         metadata.topicPartitions().stream().findFirst().get().partition())
                 .orElse(null);
-        if (lagInfo == null) {
+        if (lagInfo != null) {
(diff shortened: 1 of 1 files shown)


coltmcnealy-lh:
@eduwercamacaro is this perhaps related to the bug that Bruno found?

eduwercamacaro:
Oh no.. let me check that. I don't think it is related to [KAFKA-16836](https://issues.apache.org/jira/browse/KAFKA-16836)

eduwercamacaro (approved):

coltmcnealy-lh:
Looks like the problem I observed in the PR description was caused by changing `LHS_CLUSTER_PARTITIONS` without deleting/recreating the kafka topics. So we are fine @eduwercamacaro

PR #864, opened 2024-06-06
feat(dashboard): search for `wfRuns` by variable `name` & `value`

HazimAr:
image
image
image
image

### DIFF
--- a/dashboard/src/app/(authenticated)/(diagram)/wfSpec/[...props]/components/SearchVariableDialog.tsx
+++ b/dashboard/src/app/(authenticated)/(diagram)/wfSpec/[...props]/components/SearchVariableDialog.tsx
@@ -1,0 +1,120 @@
+import { SearchFooter } from '@/app/(authenticated)/components/SearchFooter'
+import { Button } from '@/components/ui/button'
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogHeader,
+  DialogTitle,
+  DialogTrigger,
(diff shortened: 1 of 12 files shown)


coltmcnealy-lh:
Does this support json search?

HazimAr:
> Does this support json search?

No it doesn't, it only supports `str` search right now. I just wanted to get a bones version of the search that is easy to build on. Going to try to finish all the tickets on dashboard then support other var types.

HazimAr:
Just found a bug with this, wont be able to fix until after my deployment.

The select shows all variables, it should only show variables that are `indexed`

HazimAr:
Bug fixed now

dashboard/src/app/(authenticated)/(diagram)/wfSpec/[...props]/components/SearchVariableDialog.tsx
```diff
@@ -0,0 +1,147 @@
...
+      return { str: value }
+    case 'int':
+      return { int: parseInt(value) }
+    case 'bool':
+      return { bool: value.toLowerCase() === 'true' }
+    case 'double':
+      return { double: parseFloat(value) }
+    default:
+      return { str: value }
+  }
```
Comment (mijailr) [inline dashboard/src/app/(authenticated)/(diagram)/wfSpec/[...props]/components/SearchVariableDialog.tsx:146]: This should converted by the form instead, but this implementation is sufficient we should use a different approach if we plan to support `JSON_OBJ` and `JSON_ARR`

mijailr (approved):

PR #870, opened 2024-06-07
feat(dashboard): edge modal to display details (mutations, errors, etc.)

### LINKED ISSUE #852: Show variable mutations at edge or node
### Context

We're currently rendering the thread variable mutations globally but we should render it as part of the edge somehow.

A possible solution to this is to make the edges clickable and render a tooltip with the mutations occurring on that edge.

> The problem though is that a user might not know to click the edge. Is there any way you can put a hint on the edge when it has a mutation so that a user knows that it is clickable? - Colt

> Also the edges are hard to click because they are thin, so it's easy to miss with the mouse. - Mijail

Another solution would be to extend the `NodeDetails` component or similar to append the list of mutation occurring on it's outgoingEdges. 

> Okay, I guess then if you see it when clicking on the node that would be fine. Just make sure it's clear who owns which mutations

### Acceptance Criteria

We need a little bit of thinking to find a good UX for this.

### Out of Scope

_No response_

### Components

Dashboard

HazimAr:
image
image

Closes #852

### DIFF
--- a/dashboard/src/app/(authenticated)/(diagram)/components/EdgeTypes/Default.tsx
+++ b/dashboard/src/app/(authenticated)/(diagram)/components/EdgeTypes/Default.tsx
@@ -28,23 +28,30 @@ const Default: FC<SmoothStepEdgeProps<EdgeProto>> = ({
     borderRadius: pathOptions?.borderRadius,
     offset: pathOptions?.offset,
   })
+  const { setModal, setShowModal } = useModal()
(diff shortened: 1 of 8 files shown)


HazimAr:
After speaking with @mijailr.We need to meet about this pull request again as the DX to build this is horrendous and we should talk about how much dev time we should allocate to this problem to improve the UX

mijailr (approved):

PR #871, opened 2024-06-09
fix(tenant): make rpc PutTenant idempotent

coltmcnealy-lh:
- All other metadata requests in the LH Server are idempotent and return OK if duplicated.
- This includes TaskDef, Principal, WfSpec, ExternalEventDef, and UserTaskDef.


Snarr (approved):
Looking good

eduwercamacaro (approved):

PR #872, opened 2024-06-11
fix(server): retry wait for command before a rebalance

eduwercamacaro:


server/src/main/java/io/littlehorse/server/streams/util/POSTStreamObserver.java
```diff
@@ -28,11 +36,20 @@ public POSTStreamObserver(
...
     }
 
     @Override
     public void onError(Throwable t) {
-        ctx.onError(t);
+        final boolean isRetryable = t instanceof StatusRuntimeException grpcRuntimeException
+                && grpcRuntimeException.getStatus().getCode().equals(Status.UNAVAILABLE.getCode());
+        final boolean retryOneMoreTime = timeoutAt.compareTo(new Date()) > 0;
+        if (isRetryable && retryOneMoreTime) {
+            internalComms.waitForCommand(command, this);
```
coltmcnealy-lh, line 49: Hmm. This seems like a busy retry. Fine for now I think but in the future we'll have to implement a backoff mechanism.

coltmcnealy-lh (approved):
Approved but let's add a ticket to implement a backoff mechanism.

server/src/main/java/io/littlehorse/server/streams/util/POSTStreamObserver.java
eduwercamacaro: https://github.com/littlehorse-enterprises/littlehorse/issues/873

Issue #860, opened 2024-06-05
Support mTLS `Principal`s


coltmcnealy-lh:
LittleHorse federates identity in two ways:

1. Using a Certificate Authority (mTLS)
2. Using an OAuth Identiy Provider.

We currently support both as a listener configuration (see [`LHS_LISTENERS_AUTHENTICATION_MAP`](https://littlehorse.dev/docs/operations/server-configuration#lhs_listeners_authentication_map)).

For listeners of type `OAUTH`, we can infer the `Principal` ID for any authenticated request with the following rules:
- If the token is not valid from the perspective of the Identity Provider, the `UNAUTHENTICATED` error is returned.
- If it is using the Authorization Code flow (human user), the `Principal` id is taken from the User Id from the token.
- If using the Client Credentials flow (machine user), the `Principal` id is simply the OAuth Client ID.
- If a `Principal` exists with the specified ID, then the request is treated as that `Principal`. Otherwise, it is treated as `anonymous`.

However, we currently treat every request that hits a port with authentication type `MTLS` as the `anonymous` principal (no need to throw `UNAUTHENTICATED` because the MTLS handshake will fail if the client certificate is invalid).

On an `MTLS` listener, we should look at the Subject Name on the certificate and use that to determine the `Principal` ID. If no such `Principal` exists, we can treat the request as `anonymous`.

Some resources:
* https://github.com/grpc/grpc-java/issues/6235
* https://github.com/grpc/grpc-java/blob/master/examples/example-tls/README.md

PR #874, opened 2024-06-12
feat(server): Add support for mTLS Principals

Snarr:
This PR introduces code for reading the subject name of an mTLS request certificate to determine the Principal ID. 

Notes:
- Uses CN (common name) value from the subject name of the certificate
- Adds CertificateUtil class and static method for getting subject name value from Certificate File. Implementation of getting this subject name value can change independent of MTLSServerAuthorizer logic.

Refs: #860


local-dev/configs/mtls.config
```diff
@@ -16,6 +16,7 @@ LHS_INTERNAL_ADVERTISED_HOST=localhost
 
 LHS_LISTENERS=MTLS:2023
 LHS_LISTENERS_PROTOCOL_MAP=MTLS:MTLS
+LHS_LISTENERS_AUTHENTICATION_MAP=MTLS:MTLS
```
coltmcnealy-lh, line 19: Can you add somewhere in a local-dev readme a tutorial about how to create a certificate that has a certain principal/subject name on it so that other people can test?

server/src/main/java/io/littlehorse/server/auth/MTLSServerAuthorizer.java
```diff
@@ -0,0 +1,29 @@
...
+            this.commonName = CertificateUtil.getCommonNameFromCertificate(mtlsConfig.getCaCertificate());
+        } catch (Exception ex) {
+            throw new IllegalArgumentException("Certificate is not valid", ex);
+        }
+    }
+
+    @Override
+    public <ReqT, RespT> Listener<ReqT> interceptCall(
+            ServerCall<ReqT, RespT> call, Metadata headers, ServerCallHandler<ReqT, RespT> next) {
+        headers.put(CLIENT_ID, this.commonName);
```
coltmcnealy-lh, line 25: If the `commonName` is `null`, should we put `anonymous`? Or is that handled later in the stream @eduwercamacaro ?

coltmcnealy-lh (approved):
Looks mostly good to me but we need a third pair of eyes on it, and we need someone who wasn't involved in the implementation to test it

PR #879, opened 2024-06-14
chore(sdk-js): add global imports

mijailr:
This allows to import protobuf types from the sdk-js entrypoint.

```diff
- import { NodeRun } from 'littlehorse-client/dist/proto/node_run'
- import { WfRun } from 'littlehorse-client/dist/proto/wf_run'
- import { WfSpec } from 'littlehorse-client/dist/proto/wf_spec'
+ import { NodeRun, WfRun, WfSpec } from 'littlehorse-client/proto'
```

It also introduces `buf.build` which will allow us to run compile-proto.sh from different architectures and setups.


coltmcnealy-lh (approved):
Looks good. And it would be awesome to move the proto compilation into the pipeline one day.

diablouma (approved):

PR #883, opened 2024-06-15
feat(server): rpc rescue threadRun

coltmcnealy-lh:
Allows rescuing a failed ThreadRun. Currently, only entrypiont ThreadRun's may be rescued.


coltmcnealy-lh:
I'm marking this as "Ready for Review." The current limitations are:
- We only support rescuing the "Parent" ThreadRun.
- We don't support reviving halted Child ThreadRuns after rescuing the parent.

Both issues will be addressed in a follow-up PR. I am separating them in an attempt to keep the code easier to understand.

You'll also notice I'm not doing `throw new LHApiException` but rather passing around a grpc `Optional<Status>` flag. The reason for this is because at one point we will have the ability to Bulk Rescue ThreadRun's by looking for WfRun's that failed due to a certain TaskRun failing. It won't make sense to throw an LHApiException in that case because there won't be a waiting client to accept it.

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/WfRunModel.java
```diff
@@ -499,8 +499,46 @@ public void stop(ThreadRunModel thread, ThreadHaltReasonModel threadHaltReason)
         this.advance(new Date()); // Seems like a good idea, why not?
     }
 
+    public Pair<Boolean, Status> canRescueThreadRun(int threadRunNumber, ProcessorExecutionContext ctx) {
+        this.executionContext = ctx;
```
eduwercamacaro, line 503: I don't think this is necessary.

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
```diff
@@ -359,6 +360,51 @@ private void initializeInterrupt(ExternalEventModel trigger, InterruptDefModel i
...
+        try {
+            NodeRunModel currentNR = getCurrentNodeRun();
+            NodeModel toActivate;
+            if (skipCurrentNode) {
+                toActivate = currentNR.evaluateOutgoingEdgesAndMaybeMutateVariables(ctx);
+            } else {
+                toActivate = currentNR.getNode();
+            }
+            setStatus(LHStatus.RUNNING);
+            activateNode(toActivate);
```
eduwercamacaro, line 393: it seems that now we can have more than 1 NodeRun for an specific Node. is this correct

eduwercamacaro (approved):

server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
```diff
@@ -71,105 +72,8 @@
...
-import io.littlehorse.sdk.common.proto.ListExternalEventsRequest;
-import io.littlehorse.sdk.common.proto.ListNodeRunsRequest;
-import io.littlehorse.sdk.common.proto.ListTaskMetricsRequest;
-import io.littlehorse.sdk.common.proto.ListTaskMetricsResponse;
-import io.littlehorse.sdk.common.proto.ListTaskRunsRequest;
-import io.littlehorse.sdk.common.proto.ListUserTaskRunRequest;
-import io.littlehorse.sdk.common.proto.ListVariablesRequest;
-import io.littlehorse.sdk.common.proto.ListWfMetricsRequest;
-import io.littlehorse.sdk.common.proto.ListWfMetricsResponse;
+import io.littlehorse.sdk.common.proto.*;
```
eduwercamacaro, line 75: I don't think this is what we want...

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/WfRunModel.java
coltmcnealy-lh: Noted, will fix in the follow-up PR

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
coltmcnealy-lh: This is correct. But it has been true for a while, since we can have loops in your ThreadSpec which cause the ThreadRun to arrive at a given Node multiple times.

The alternative is to discard the old NodeRun and run it again, but I think that is sloppy since it would be making something mutable that was previously immutable.

server/src/main/java/io/littlehorse/server/KafkaStreamsServerImpl.java
coltmcnealy-lh: :porg: That is not what I wanted to do. Maybe `gradle spotlessApply` did that?

coltmcnealy-lh: image
That's what happened :joy:

server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/RescueThreadRunRequestModel.java
```diff
@@ -0,0 +1,76 @@
...
+            throw new LHApiException(Status.NOT_FOUND, "Couldn't find WfRun %s".formatted(wfRunId));
+        }
+
+        try {
+            wfRun.rescueThreadRun(threadRunNumber, skipCurrentNode, ctx);
+        } catch (MissingThreadRunException exn) {
+            throw exn.toLHApiException();
+        } catch (UnRescuableThreadRunException exn) {
+            throw exn.toLHApiException();
+        }
```
eduwercamacaro, line 67: You need to catch `RuntimeException` too
```suggestion
        } catch (Exception exn) {
            throw new LHApiException(Status.INTERNAL, exn.getMessage());
        }
```

PR #884, opened 2024-06-16
chore(e2e): run tests in distributed mode

coltmcnealy-lh:
- Use two servers instead of one on the StandaloneBootstrapper
- Increase the number of partitions to 12
- Use two core stream threads per instance
- Use 4 JUNIT test threads

Note that this causes a bunch of failures on localhost, implying that we have some race conditions and weaknesses in our server. Yikes!


coltmcnealy-lh:
By refactoring the `tests.yml` file, we have cut the time for `Validate Build` on `tests-server` from 2 minutes and 34 seconds to 1 minute and 35 seconds.

We can also improve the time taken for the e2e once we get it running with parallel threads on a larger test runner.

.github/workflows/tests.yml
```diff
@@ -89,12 +91,10 @@ jobs:
...
           distribution: "corretto"
           java-version: 17
-      - name: Validate Indentations
-        run: ./gradlew spotlessCheck
       - name: Validate Build
-        run: ./gradlew build -x test
+        run: ./gradlew server:spotlessCheck test-utils:spotlessCheck server:build -x test
       - name: Tests
-        run: ./gradlew server:test server:e2e server:build
+        run: ./gradlew server:test server:e2e
```
coltmcnealy-lh, line 97: These changes do not reduce our pipeline coverage; however, they reduce the build time by 59 seconds.

coltmcnealy-lh:
Alright, it's working now, and I fixed some rather unpleasant things that were happening in the server (including a tenant security vulnerability with awaitWorkflowEvent) but I think this should be an improvement overall.

test-utils/src/main/java/io/littlehorse/test/LHExtension.java
```diff
@@ -75,26 +75,34 @@ public void postProcessTestInstance(Object testInstance, ExtensionContext contex
...
-                throw ex;
-            }
-        }
+        Awaitility.await()
+                .atMost(Duration.ofSeconds(25))
+                .ignoreExceptionsMatching(exn -> RuntimeException.class.isAssignableFrom(exn.getClass()))
+                .until(() -> {
+                    testContext
+                            .getLhClient()
+                            .withDeadlineAfter(2, TimeUnit.SECONDS)
```
coltmcnealy-lh, line 101: We have a weakness in our server where if we send a request before the server is fully rebalanced, it will hang until the default grpc deadline. @eduwercamacaro we should look at that this week. If I remove the `deadlineAfter` then the `StandaloneTestBootstrapper` wont start on my laptop (it works fine on the pipeline though because the smaller resources causes different sequencing).

test-utils/src/main/java/io/littlehorse/test/internal/ExternalTestBootstrapper.java
```diff
@@ -18,10 +17,7 @@ public ExternalTestBootstrapper() {
             throw new IllegalStateException(String.format("Configuration file %s doesn't exist", LH_CONFIG_FILE));
         }
         workerConfig = new LHConfig(configPath.toString());
-        lhClient = workerConfig
-                .getBlockingStub()
-                .withCallCredentials(new StandaloneTestBootstrapper.MockCallCredentials(
-                        new TenantIdModel(workerConfig.getTenantId())));
+        lhClient = workerConfig.getBlockingStub();
```
coltmcnealy-lh, line 20: Setting the call credentials is the responsibility of the `LHConfig`, so we don't need to do it here.

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/PollThread.java
```diff
@@ -37,6 +37,7 @@ public void run() {
                 }
             }
         } catch (InterruptedException ex) {
+            System.out.println("PEDROOOOOOOO");
```
Snarr, line 40: 💀

server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/VariableMutationModel.java
```diff
@@ -155,7 +155,7 @@ public void execute(ThreadRunModel thread, Map<String, VariableValueModel> txnCa
         } catch (LHVarSubError exn) {
             throw exn;
         } catch (Exception exn) {
-            log.error(exn.getMessage(), exn);
+            log.trace(exn.getMessage(), exn);
```
Snarr, line 158: Keep this?

coltmcnealy-lh: This is a user-level error that will show up in the `errorMessage` on the `NodeRun`. It's not an actual exception but rather the User's WfSpec is invalid. We don't want that polluting our stacktraces—they should debug it using the dashboard or `lhctl` rather than looking at our server logs (especially because in LH Cloud they won't see the logs).

eduwercamacaro (approved):

PR #889, opened 2024-06-18
feat: shows [bytes] placeholder for variables that have such content

diablouma:
* Variables that have bytes as value where causing the dashboard to break, now we are showing a placeholder for them.


dashboard/src/app/utils/variables.ts
```diff
@@ -11,9 +11,19 @@ export const getVariable = (variable?: VariableAssignment) => {
 }
 
 export const getVariableValue = (variable?: VariableValue) => {
+  console.log("variable:", variable)
   if (!variable) return
+
   const key = Object.keys(variable)[0] as keyof VariableValue
-  return variable[key]
+  console.log("key:", key, )
```
mijailr, line 18: Remove log

diablouma: done

mijailr (approved):

Issue #801, opened 2024-05-14
Garbage Collect `TaskWorkerGroup`s


coltmcnealy-lh:
Consider the following things happening in this order:

1. Create `TaskDef` `foo`
2. A bunch of Task Workers for `foo` start working, and then we have a `TaskWorkerGroup`
3. We stop the `foo` Task Workers and wait a few hours
4. Checkpoint `4`
5. We delete the `foo` `TaskDef`
6. Checkpoint `6`

At checkpoint `4`, the `TaskWorkerGroup` object (via `rpc GetTaskWorkerGroup` will still show that there are polling workers. We should have some field in it that keeps track of the last time a worker has polled, so that when we `rpc GetTaskWorkerGroup` we can tell that there is no activity.

At checkpoint `6`, we currently will leave the `TaskWorkerGroup` in our data store forever and not clean it up. We should delete it.

PR #894, opened 2024-06-20
fix(server): Delete TaskWorkerGroups after deleting TaskDef

Snarr:
Add new internal DeleteTaskWorkerGroupRequest that is forwarded following a DeleteTaskDefRequest. PR is a work in progress.

Ref: #801


server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/DeleteTaskWorkerGroupRequestModel.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public boolean hasResponse() {
+        return false;
+    }
+
+    @Override
+    public Message process(ProcessorExecutionContext executionContext, LHServerConfig config) {
+        System.out.println("Hello World i'm running");
```
coltmcnealy-lh, line 32: Delete please :smile:

coltmcnealy-lh (approved):
Looks good

PR #896, opened 2024-06-22
Fix/flakey test wait for threads

coltmcnealy-lh:


coltmcnealy-lh:
@mijailr any idea why there are proto changes? Ideally the `local-dev/compile-proto.sh` script should still do the same as the pipeline.

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
coltmcnealy-lh, line 1: The only real changes should be in this file. Everything else seems like a misconfiguration of our proto compilers

eduwercamacaro (approved):

PR #897, opened 2024-06-23
[WIP] Feature/finish wfspec migration

coltmcnealy-lh:
Just putting this PR here in case anyone is curious. I'm going to be hacking on this when I am stuck on a plane with no wifi.


coltmcnealy-lh:
It turns out that iterating with a rangescan in the `CommandProccessor` is a Very Big No-No. I think we'll have to consider implementing this feature from within the `RequestExecutionContext` rather than inside the core processor context. Going to close this PR. It's also incredibly stale.

PR #902, opened 2024-06-26
fix(dashboard): handle tenantId for taskRuns

mijailr:
Fixes issue reported by @MateoRojas while clicking `"Inspect TaskRun"` on a tenant.


coltmcnealy-lh (approved):
It's a little scary if we have to provide this _everywhere_...we will probably forget at some point.

mijailr:
> It's a little scary if we have to provide this _everywhere_...we will probably forget at some point.

True, we could forget. I'll think how to abstract that.

MateoRojas (approved):

PR #907, opened 2024-07-02
feat(sdk-java): Support future stubs for async API requests

eduwercamacaro:
Some clients require to send async requests to the server this is helpful for high throughput use cases, for instance, executing `rpc RunWf()` high number of requests per second.


sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
```diff
@@ -250,17 +243,15 @@ public LittleHorseBlockingStub getBlockingStub(String host, int port) {
...
+     * Gets a Future gRPC stub for the LH Public API on a specified host and port.
      * @param host is the host that the LH Server lives on.
      * @param port is the port that the LH Server lives on.
-     * @param tenantId is the current authenticated tenant.
-     * @return a blocking gRPC stub for that host/port combo.
+     * @return a future gRPC stub for that host/port combo.
      */
-    public LittleHorseBlockingStub getBlockingStub(String host, int port, TenantId tenantId) {
-        return getBaseBlockingStub(host, port).withCallCredentials(new TenantMetadataProvider(tenantId.getId()));
+    public LittleHorseBlockingStub getFutureStub(String host, int port) {
```
coltmcnealy-lh, line 251: Should we also support `getFutureStub(String tenantId)`?

coltmcnealy-lh (approved):

PR #908, opened 2024-07-02
fix(server): fixes rebalance implementation during instance shutdowns

eduwercamacaro:
This caused unclean shutdowns in the kafka streams instance. Follow up PR will be created after this to implement a proper mechanism to handle uncaught exceptions in kafka streams.


Snarr (approved):
So what happens if the server rebalance logic fails? Can that happen?

server/src/main/java/io/littlehorse/server/monitoring/metrics/InstanceState.java
```diff
@@ -76,7 +76,13 @@ public void onChange(KafkaStreams.State newState, KafkaStreams.State oldState) {
 
         if (newState == KafkaStreams.State.RUNNING) {
             activeTasks.set(currentActiveTaskIds);
-            internalComms.handleRebalance(activeTasks.get());
+            try {
+                internalComms.handleRebalance(activeTasks.get());
+            } catch (Exception toIgnore) {
```
coltmcnealy-lh, line 81: Should we at least `log.warn()` it ?

coltmcnealy-lh (approved):
:tada:

PR #914, opened 2024-07-08
ci(lhctl): release lhctl binaries and generate release notes

mijailr:
This PR adds lhctl binaries to the release page and generates release notes for every new release.

Supersedes #913


HazimAr (approved):
LGTM!

PR #917, opened 2024-07-12
ci(goreleaser): use go releaser

mijailr:
This uses a different approach to generate the go binaries using [goreleaser](https://github.com/goreleaser/goreleaser)

image


HazimAr (approved):
will do some more testing on windows for this releaser

PR #926, opened 2024-07-16
fix(canary): sets workflow retention policy to 1 day

coltmcnealy-lh:


Snarr (approved):
Good to have a retention policy!

PR #932, opened 2024-07-17
fix(server): don't busy-retry on wait-for-command

coltmcnealy-lh:
- Uses network thread pool for internal GRPC clients to reduce number of total threads used by server
- Uses network thread pool to clean up old command waiters
- Fixes stack overflow error on the wait for command retry logic and avoids busy waiting by using network thread pool to schedule retries later.


jhosep-marin (approved):
Code wise these changes look good to me :+1:

Snarr (approved):
TIL about ScheduledExecutorServices 👍 Looks good

coltmcnealy-lh:
Eduwer will have to review this when he's back, because it's going to be veeeerrrrrry controversial :joy: 

But overall this should be a good change, it will reduce the number of threads used and will also improve reliability during rebalances.

coltmcnealy-lh:
We had seen some `StackOverflowError`s during soak testing which led to unclean shutdown, state store restoration, etc... this should fix that.

PR #933, opened 2024-07-26
fix(dashboard): return nodeRun when `isChecked`

HazimAr:
`ExternalEventSearch` request needs to only `GetNodeRun` when the property `isClaimed` is true otherwise `GetNodeRun` returns a `NOT FOUND` error and breaks the search.


Snarr (approved):
Good fix 👍

PR #937, opened 2024-07-29
fix(ci): trigger tests on prs open by forks

HazimAr:


Snarr (approved):
I think this looks good

PR #939, opened 2024-08-03
[WIP] Add Docker Compose server examples

Snarr:
Adds Docker Compose examples for the following LittleHorse Orchestrator use cases:
- 3 server cluster
- Authentication with OAuth
- Authentication with mTLS
- Using Keycloak iDP
And more.


examples/docker-compose/three-servers/docker-compose.yml
```diff
@@ -0,0 +1,86 @@
+services:
+  kafka:
+    container_name: lh-kafka
+    image: bitnami/kafka:3.5
```
coltmcnealy-lh, line 4: Let's just use the `apache/kafka:3.8.0` one since it's now an official image, and it also means you don't need to configure any environment variables.

Snarr: Until now I didn't realize the full extent of what an Apache Kafka image release meant. Cool that they are releasing their own images now with built in env configurations!

coltmcnealy-lh:
We will want to put this in the `docs/docs/operations` folder so that it shows up on our website rather than just in the `examples` space. I'm thinking we should have inside the "operations" tab on our website a "docker compose" tab, which has a few examples.

This can be the `Basic` one. We should also have some others:
* OAuth with Keycloak
* Using Confluent Cloud
* TLS
* mTLS

Each file should also have tutorials for connecting with `lhctl` and also a Task Worker in Java/Go/Python.

coltmcnealy-lh (approved):

PR #944, opened 2024-08-12
Feature/server secrets

eduwercamacaro:
This PR allows clients to associate a VariableDef to a NodeOutput, set a custom variable name and an LHType, and also mark variables as masked, which masks the value associated with that variable from external client queries.

See `e2e.VariablesTest`

Go and Python sdk support will be included in a separate PR


server/src/main/java/io/littlehorse/common/model/getable/core/taskrun/TaskAttemptModel.java
```diff
@@ -45,8 +47,13 @@ public Class<TaskAttempt> getProtoBaseClass() {
     @Override
     public void initFrom(Message proto, ExecutionContext context) {
         TaskAttempt p = (TaskAttempt) proto;
+        maskedValue = p.getMaskedValue();
         if (p.hasOutput()) {
-            output = VariableValueModel.fromProto(p.getOutput(), context);
+            if (maskedValue && context instanceof RequestExecutionContext) {
+                output = new VariableValueModel(LHConstants.STRING_MASK);
```
coltmcnealy-lh, line 53: The only concern I have here is that non- `STR` variables will be converted into `STR`.

coltmcnealy-lh (approved):
Very well done!

server/src/main/java/io/littlehorse/common/model/getable/core/taskrun/TaskAttemptModel.java
eduwercamacaro: It makes sense to me because mask is a String literal.

PR #946, opened 2024-08-13
feat(server): allow rpc save user task

coltmcnealy-lh:
Jacob will add `lhctl` support.

I will add e2e tests.


jhosep-marin:
So far is looking good to me :+1:

coltmcnealy-lh:
We will have Jacob's docs + `lhctl` in a separate PR.

server/src/main/java/io/littlehorse/common/model/getable/core/usertaskrun/UserTaskRunModel.java
```diff
@@ -297,6 +299,17 @@ public void cancel() {
         failureToThrowKenobi = new FailureModel("User task cancelled", failureName);
     }
 
+    public void processProgressSavedEvent(SaveUserTaskRunProgressRequestModel req, ProcessorExecutionContext ctx)
+            throws LHApiException {
+        if (isTerminated()) {
+            throw new LHApiException(Status.FAILED_PRECONDITION, "UserTaskRun is in status " + status);
+        }
+
+        this.results = req.getResults();
```
eduwercamacaro, line 308: So, this will overwrite the entire current progress.

eduwercamacaro (approved):

server/src/main/java/io/littlehorse/common/model/getable/core/usertaskrun/UserTaskRunModel.java
coltmcnealy-lh: Yes, that is true. We should document that. It's not a `PATCH` but more of a `POST` I think

Snarr (approved):

PR #952, opened 2024-08-16
feat(lhctl): Add `saveUserTaskRunProgressCmd`

Snarr:
This PR adds the `saveUserTaskRunProgressCmd` to the `lhctl` client. It also adds Python and Go documentation for using the RPC call.


lhctl/cmd/user_task_run.go
```diff
@@ -151,6 +151,83 @@ var getUserTaskRunCmd = &cobra.Command{
...
+		}
+
+		assignmentPolicy, err := promptFor(
+			"Enter the number corresponding to your choice: ",
+			model.VariableType_INT,
+		)
+		if err != nil {
+			log.Fatal(err)
+		}
+		saveUserTaskRunProgress.Policy = model.SaveUserTaskRunProgressRequest_SaveUserTaskRunAssignmentPolicy(assignmentPolicy.GetInt())
```
Snarr, line 221: Most of this code comes from `executeUserTaskRunCmd`, so it is now duplicated.

If we want to spend more time on this, I can refactor and move duplicate code to functions. I don't think this is worth our time right now but let me know your thoughts.

lhctl/cmd/user_task_run.go
```diff
@@ -151,6 +151,83 @@ var getUserTaskRunCmd = &cobra.Command{
 	},
 }
 
+var saveUserTaskRunProgressCmd = &cobra.Command{
+	Use:   "saveUserTaskRun",
```
coltmcnealy-lh, line 155: To make this fit with the rest of `lhctl`'s design patterns, the Verb and the Object Noun should be separate. This means it should be:

```
lhctl save userTaskRun <wfRunId> <userTaskGuid>
```

To do that, you'll need to add a `save.go` and a `saveCmd`.

coltmcnealy-lh (changes requested):
Needs to re-work it a little bit to fit the rest of our design patterns.

lhctl/cmd/user_task_run.go
coltmcnealy-lh: That's not a huge issue, but @eduwercamacaro may think differently. The important thing is the verbiage should be `lhctl save userTaskRun` rather than `lhctl saveUserTaskRun`

lhctl/cmd/user_task_run.go
Snarr: I recognized this design pattern when developing this but chose not to follow it because I think is bad UX having one single command inside of a verb command.

It adds too many layers of `--help` menus that are unnecessary. 2 layers vs 3.

Long term, it will make adding more `save` commands slightly easier but in the short term it will cost developers more time to discover and use this command.

coltmcnealy-lh (approved):

PR #955, opened 2024-08-19
Remove bump scripts

Snarr:
This PR removes the "Bump Version" Python script and other documents referencing it from our repo in favor of our new Release workflow that manages releases.


mijailr (changes requested):
This can't be merge yet. I'm working for a replacement to this script.

local-dev/README.md
```diff
@@ -67,14 +67,6 @@ However, you can start two Brokers in your terminal as follows:
 ./local-dev/do-server.sh server-2
 ```
```
coltmcnealy-lh, line 69: ## Release a New Version

To release a new version, you should push a properly-formed Git Tag, and our pipeline will take care of it.

```
git tag -am "Release X.Y.Z" vX.Y.Z
git push --follow-tag
```
Note that we follow [Semantic Versioning](https://semver.org) and that `X` is the major version, `Y` is the minor version, and `Z` is the patch.

While SemVer frowns upon the `v` prefix, we need it because of how Go module releases work. Our pipeline removes the `v` from all generated artifacts (eg. `sdk-java` on maven, `sdk-python` on pypi, our docker images, etc).

mijailr:
This can continue since #962 was merged already

PR #956, opened 2024-08-20
feat(sdk-python): Masked values support

eduwercamacaro:
Supports masked variables in python sdk


sdk-python/examples/basic/example_basic.py
```diff
@@ -28,7 +29,9 @@ def my_entrypoint(wf: WorkflowThread) -> None:
     return Workflow("example-basic", my_entrypoint)
 
 
-async def greeting(name: str, ctx: WorkerContext) -> str:
+async def greeting(
+    name: Annotated[str, LHType(name="another-name", masked=True)], ctx: WorkerContext
```
Snarr, line 33: I think all of our SDK examples across languages should mimic one another. Por ejemplo, if this example is going to have a masked value in Python I think the basic example in Java should also have a masked value.

coltmcnealy-lh (approved):
Looks good!

PR #962, opened 2024-08-23
build(ci): add automated release management

mijailr:
This removes the need to run `./local-dev/bump.sh` to prepare for new version releases. By default `0.0.0-development` or just `0.0.0` is statically present in some files but it's replaced by the corresponding version during the workflow run.

The release workflow will be triggered on any commit tagged with  prefix `v` and the semantic version like `v0.0.0-development`.

The pipeline process generates:

- docker images with the matching semantic version
    - lh-server:0.0.0-development
    - lh-dashboard:0.0.0-development
    - etc
- SDKs with semantic version
    - sdk-java
    - sdk-python
    - sdk-js
- sdk-go module with git tag
- lhctl version 
- lh-server ServerVersionResponse{majorVersion, minorVersion, patchVersion, preReleaseIdentifier}

Example output `lhctl version`
```
$ lhctl version
# lhctl version: 0.0.0-development (Git SHA none)
# Server version: 0.0.0-development
```


eduwercamacaro (approved):

gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=0.10.1
+version=0.0.0-development
```
coltmcnealy-lh, line 1: When Kafka is on version `3.8.1` (the next version is 3.9), they do: `3.9.0-SNAPSHOT`. Wondering if we can do that?

For our case, right now it would be `0.11.0-SNAPSHOT` since we're about to release `0.11.0`. But then we'd have to update it either manually or in the release pipeline, which might be too much work.

coltmcnealy-lh (approved):
Good work!

PR #973, opened 2024-08-26
feat(trivy): trivy integrated to scan the docker images

yesidevelop:


mijailr:
Not required since `v1` alias is already pointing to the version enabling trivy.

PR #957, opened 2024-08-21
refactor(sdk-go): Independent Go Modules for lhctl and sdk-go

eduwercamacaro:
The objective of this PR is to distribute `lhctl` and `sdk-go` across various Go modules, enabling clients to select the appropriate Go module for import. Also, this PR includes a major refactor for Go SDK packages:

- The `lhproto` package contains all of the protobuf structures related to the public API of Littlehorse Server.

- `littlehorse` package contains all of the functions that create a workflow spec and task worker

PR #984, opened 2024-08-28
build(sdk-go): use root module

### LINKED ISSUE #814: Properly Tag `sdk-go` Releases
In GoLang, if you want to have properly tagged releases, you need to prefix the git tags with a `v`.

We currently don't do that since we are following Semantic Versioning. The result is that, when I do the following:

```
go get github.com/littlehorse-enterprises/littlehorse/sdk-go@0.9.0
```

my `go.mod` has the following entry:

```
require github.com/littlehorse-enterprises/littlehorse v0.0.0-20240516033734-7b120ed34cc9
```

It should look like:

```
require github.com/littlehorse-enterprises/littlehorse v0.9.0
```

### Potential Fixes
The only way I know of to fix this is that we would have to add a new git `tag` for each release, prefixed with the `v`.

### Things to Break

We will insist on Semantic Versioning for the rest of the project, which means no `v` prefix for our `sdk-java`, `sdk-python`, `sdk-js`, `lh-standalone`, `lh-dashboard`, and `lh-server` artifacts. 

I just propose that we push a second tag with the `v` prefix in our `local-dev/bump.sh` script, and that we configure our release pipeline to ignore anything starting with a `v`. Everything else stays the same.

mijailr:
This move back the `go.mod` file to the root to make it available to install with

```shell
go get github.com/littlehorse-enterprises/littlehorse:vN.N.N
```

Closes #814
Related to #962 and #957


eduwercamacaro (approved):
LGTM

PR #994, opened 2024-09-06
csharp flag added to protobuf

KarlaCarvajal:
- Add csharp flag to generate sdk-dotnet models


eduwercamacaro:
I think you should push the generated classes too. This might change in the future though. @mijailr  what do you think?

KarlaCarvajal:
> I think you should push the generated classes too. This might change in the future though. @mijailr what do you think?

I was going to upload the generated files when the dotnet-sdk were fix, in a new branch

mijailr:
Yeah it's fine. We want to remove the generated files from the repo but that will be done later.

sauljabin (approved):

KarlaCarvajal:
There was no need to generate WF classes from the proto script, in csharp it is generated within a cproj declaration

PR #1002, opened 2024-09-13
Refactor Dotnet SDK

KarlaCarvajal:
- Refactor Dotnet SDK code to avoid using much complexity using builders and interfaces
- Added some dotnet files to gitignore
- Fixed the dotnet basic example

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,7 @@ dist/
 
 .DS_Store
 *.work
+
+# dotnet
+*.sln
+global.json
--- a/sdk-dotnet/LittleHorse.Sdk/CONTRIBUTING.md
+++ b/sdk-dotnet/LittleHorse.Sdk/CONTRIBUTING.md
@@ -1,20 +1,29 @@
 # LittleHorse Contributing .NET SDK
 
-You need to package the project to generate the nuget file `LittleHorse.{version}.nupkg`
+Create a folder where to place local nuget packages
 
-```dotnet pack```
+```mkdir $HOME/nuget-packages```
 
-How to set a local nuget directory packages to call its dependencies, you should specify where your .nupkg is created
+You need to package the project to generate the nuget file `LittleHorse.Sdk.{{version}}.nupkg` in a common folder
 
-```dotnet nuget add source {{your_local_path}}/littlehorse/sdk-dotnet/LittleHorse.Sdk/bin/Debug --name LocalNugetPackages```
+```
+dotnet build
+dotnet pack --output $HOME/nuget-packages
+```
 
-next: call the dependency client csproj file specifying the local version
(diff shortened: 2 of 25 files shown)


.gitignore
```diff
@@ -41,3 +41,7 @@ dist/
 
 .DS_Store
 *.work
+
+# dotnet
+*.sln
```
eduwercamacaro, line 46: Is there any particular reason to ignore sln files?

sdk-dotnet/LittleHorse.Sdk/CONTRIBUTING.md
```diff
@@ -0,0 +1,21 @@
+# LittleHorse Contributing .NET SDK
+
+You need to package the project to generate the nuget file `LittleHorse.{version}.nupkg`
+
+```dotnet pack```
+
+How to set a local nuget directory packages to call its dependencies, you should specify where your .nupkg is created
+
+```dotnet nuget add source {{your_local_path}}/littlehorse/sdk-dotnet/LittleHorse.Sdk/bin/Debug --name LocalNugetPackages```
```
eduwercamacaro, line 9: Is it possible to use relative paths?
```suggestion
```dotnet nuget add source ./bin/Debug --name LocalNugetPackages```
```

KarlaCarvajal: I belive yes, lets see

.gitignore
eduwercamacaro: But I think a `solution` is an artifact that groups multiple projects together. In this case we have multiple projects the actual sdk and the other examples. 
A solution file might be important to add because we could build and execute examples at `sdk-dotnet` root path.

sdk-dotnet/LittleHorse.Sdk/CONTRIBUTING.md
KarlaCarvajal: done

.gitignore
KarlaCarvajal: but we do not have a solution with different projects, we have a folder with different dotnet projects, until now I have not seen the usage of the sln file, vscode adds it in each proj but not need in command line

eduwercamacaro: What I'm saying is that you can do something like this:

```bash
dotnet new sln --name Littlehorse 
dotnet sln add LittleHorse.Sdk
dotnet sln add Examples/BasicExample
dotnet build
```

this is a way to build the binaries for the sdk and examples.

KarlaCarvajal: we do not need because you can have examples where you want event client wont be part of the solution, we just have a SDK project published to public

eduwercamacaro: I think we will need this at some point, but it is ok. We can add it in the future if need it.

eduwercamacaro (approved):

PR #1006, opened 2024-09-16
chore(lh-server): Updates javalin version

eduwercamacaro:
Solves a vulnerability problem related to javalin dependency


coltmcnealy-lh (approved):
`lh-operator` probably has this problem too.

mijailr:
BTW we're planning to do an experiment to remove javalin entirely. I'll let you know when it's done.

PR #1010, opened 2024-09-18
feat(sdk-dotnet): Add logger in LH configuration

KarlaCarvajal:
- Add logerFactory so it can create a logger from Classes
- Add unit tests
- Change basic example to pass logger configs in LHConfig


sauljabin (approved):
clean 👍🏽

sdk-dotnet/LittleHorse.Sdk/Internal/LHLoggerFactoryProvider.cs
```diff
@@ -0,0 +1,22 @@
...
+
+        public static void Initialize(ILoggerFactory? loggerFactory = null)
+        {
+            _loggerFactory = loggerFactory ?? NullLoggerFactory.Instance;
+        }
+
+        public static ILogger<T> GetLogger<T>()
+        {
+            if (_loggerFactory != null) return _loggerFactory.CreateLogger<T>();
+            throw new LHLoggerFactoryNullException("_loggerFactory does not have a valid value and it is trying to create a logger instance.");
```
eduwercamacaro, line 19: I think you can use [InvalidOperationException](https://learn.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?view=net-8.0) instead of `LHLoggerFactoryNullException`

eduwercamacaro: [Suggestion]
Use if/else structure. IMO, it is cleaner.

KarlaCarvajal: sure, let me see

eduwercamacaro (approved):

PR #1031, opened 2024-09-25
fix(dashboard): better external event UX

HazimAr:
- Timeouts are now shown
- Timeouts are formatted 165 seconds -> "2m 45s"
- `Inspect TaskRun` -> `Inspect ExternalEvent`
- `Inspect ExternalEvent` only shown once `ExternalEvent` has been called

# Timeouts:
image
image

### DIFF
--- a/dashboard/src/app/(authenticated)/externalEventDef/[name]/components/ExternalEventDef.tsx
+++ b/dashboard/src/app/(authenticated)/externalEventDef/[name]/components/ExternalEventDef.tsx
@@ -19,6 +19,7 @@ import { Details } from './Details'
 type Props = {
   spec: ExternalEventDefProto
 }
+
 export const ExternalEventDef: FC<Props> = ({ spec }) => {
   const [createdAfter, setCreatedAfter] = useState('')
   const [createdBefore, setCreatedBefore] = useState('')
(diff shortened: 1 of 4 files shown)


dashboard/src/app/(authenticated)/externalEventDef/[name]/components/ExternalEventDef.tsx
```diff
@@ -27,7 +28,7 @@ export const ExternalEventDef: FC<Props> = ({ spec }) => {
   const { tenantId } = useWhoAmI()
 
   const { isPending, data, hasNextPage, fetchNextPage } = useInfiniteQuery({
-    queryKey: ['taskRun', tenantId, 10, createdAfter, createdBefore, isClaimed],
+    queryKey: ['externalEvent', tenantId, 10, createdAfter, createdBefore, isClaimed],
```
Comment (mijailr) [inline dashboard/src/app/(authenticated)/externalEventDef/[name]/components/ExternalEventDef.tsx:31]: What that 10 means?

mijailr (changes requested):
Why is it renaming userTask?

dashboard/src/app/(authenticated)/externalEventDef/[name]/components/ExternalEventDef.tsx
Comment (HazimAr) [reply dashboard/src/app/(authenticated)/externalEventDef/[name]/components/ExternalEventDef.tsx:31]: Not sure, ill remove it, it was copied from somewhere else. Maybe its needed in that other example so I'm not going to remove it in the other instance.

HazimAr:
> Why is it renaming userTask?

Good catch, looks like I accidentally moved the folder.

dashboard/src/app/utils/date.ts
```diff
@@ -14,3 +14,28 @@ export const utcToLocalDateTime = (utcISODateTime: string): string =>
   new Date(Date.parse(utcISODateTime)).toLocaleString(undefined, { hour12: false, timeZoneName: 'short' })
 
 export const localDateTimeToUTCIsoString = (localDateTime: string): string => new Date(localDateTime).toISOString()
+
+export const formatTime = (seconds: number): string => {
```
mijailr, line 18: Can you add unit tests for this?

mijailr (changes requested):

mijailr (approved):

Issue #901, opened 2024-06-26
Users able to reveal Tenants they are not assigned to


Snarr:
### Context

A user with the `ALL_RESOURCES` ACL can search for and reveal the existence of tenants even when they are not assigned to them.

### Steps To Reproduce

1. Create a Principal that has the `ACL_ALL_RESOURCES` ACL permission scoped to a specific tenant (example: test).
2. Create another tenant within the same `lhCluster` that the new Principal does not have access to.
3. Run a `SearchTenant` request as the Principal defined in Step 1.
4. BUG: Server returns both tenants, including the one the Principal is assigned to AND the one the Principal is not assigned to

### Expected Behavior

A non-admin user with the `ACL_ALL_RESOURCES` perTenantAcl that runs a SearchTenant request should receive a PERMISSION_DENIED error. Cluster-scoped resources are not available to Principals using perTenantAcls, to protect the privacy of other tenants within the same cluster.

### Components

Server

coltmcnealy-lh:
Jacob finished it

PR #1032, opened 2024-09-25
fix(server): Require global permissions for cluster scoped-resources

Snarr:
- Deny requests where the principal has PerTenantACLs over specified tenant but is requesting access to cluster-scoped resources
- Rejects `PutPrincipal` requests that declare PerTenantACLs with access to cluster-scoped resources

Solves https://github.com/littlehorse-enterprises/littlehorse/issues/901


Snarr:
Thoughts on adding these two tests:

```java
@Test
public void shouldNotAllowPrincipalWithAllResourcePerTenantACLToAccessTenantResource() {}

@Test
public void shouldNotAllowPrincipalWithAllResourcePerTenantACLToAccessPrincipalResource() {}
```

Basically to verify that per-tenant ACL "ALL_RESOURCES" no longer gives access to Principals or Tenants?

server/src/main/java/io/littlehorse/server/auth/RequestAuthorizer.java
```diff
@@ -121,9 +120,22 @@ private void loadMetadataFromServiceClass(final Class<? extends BindableService>
...
             AuthMetadata authMetadata = methodMetadata.get(methodName);
+
+            Collection<ServerACLModel> acls;
+
+            if (!authContext.perTenantAcls().isEmpty()) {
+                if (requiresClusterScopedResource(authMetadata)) {
+                    throw new PermissionDeniedException(
+                            "Insufficient ACLs. Access to Principals and Tenants can only be granted with Global ACLs");
+                }
+                acls = authContext.perTenantAcls();
```
eduwercamacaro, line 134: With this implementation, what is the reason to have different methods for `perTenantAcls` and global `acls`?

server/src/test/java/io/littlehorse/common/model/metadatacommand/PrincipalAdministrationTest.java
```diff
@@ -209,21 +219,6 @@ public void supportPrincipalDeletion() {
                 .isNull();
     }
 
-    @Test
-    public void shouldRestrictPrincipalCreationToCurrentTenant() {
```
eduwercamacaro, line 213: This test prevents the creation of a principal that has `perTenantAcls` differents from the requester principal's `perTenantAcls` Could you please explain why this test should be removed?

server/src/main/java/io/littlehorse/server/auth/RequestAuthorizer.java
Snarr: Without both methods, `RequestAuthorizer` has no way of knowing if the ACLs present are from PerTenant or Global ACLs. So it cannot differentiate between `ACL_ALL_RESOURCES` in a global context or in a per-tenant context (where it blocks principal and tenant requests).

There are ways to refactor this perhaps, where the logic for deciding between PerTenant or Global is abstracted away, and maybe an Enum is passed on indicating the context.

server/src/test/java/io/littlehorse/common/model/metadatacommand/PrincipalAdministrationTest.java
Snarr: If a requester has perTenantAcls for a tenant, and it is making requests with that tenants client_id, then the request should be rejected entirely. It should not be able to access cluster-scoped resources (Principals or Tenants) at all.

Maybe I am misunderstanding the purpose of this test, but it seemed to not make sense in this new permissions context.

server/src/main/java/io/littlehorse/server/auth/RequestAuthorizer.java
eduwercamacaro: As far I understand, the Collection on [this line](https://github.com/littlehorse-enterprises/littlehorse/pull/1032/files/24eefaf32e4b84d2ad4743dba097dbd3f72b7cd4#diff-9937bc62b704d7fa071c63d6a7633f625fd209542ff341ed9aa045111fa4df80R127)  stores both global and perTenantAcls, so it seems that you need that only for this validation ([line #130](https://github.com/littlehorse-enterprises/littlehorse/pull/1032/files/24eefaf32e4b84d2ad4743dba097dbd3f72b7cd4#diff-9937bc62b704d7fa071c63d6a7633f625fd209542ff341ed9aa045111fa4df80R130)). is that correct?

server/src/test/java/io/littlehorse/common/model/metadatacommand/PrincipalAdministrationTest.java
eduwercamacaro: Then this test should change, we need to validate that behavior that you just described.

server/src/main/java/io/littlehorse/server/auth/RequestAuthorizer.java
Snarr: Yes, I believe. My logic is as follows:

1. If per-tenant ACLs exist on authContext AND the service being accessed requires cluster-scoped resources, reject the request
2. Otherwise, if per-tenant ACLs exist, use the per-tenant ACLs
3. If per-tenant ACLs don't exist in authContext, default to using global ACLs

server/src/test/java/io/littlehorse/common/model/metadatacommand/PrincipalAdministrationTest.java
Snarr: Precisely, I believe the 2 tests I proposed here https://github.com/littlehorse-enterprises/littlehorse/pull/1032#issuecomment-2377371915 should target that scenario.

server/src/main/java/io/littlehorse/server/auth/RequestAuthorizer.java
Snarr: To satisfy this logic ^ we need two pieces of information:
1. Know whether or not the Principal has perTenantAcls for the tenant its making the request in
2. Know whether or not the service being accessed requires cluster-scoped resouces

eduwercamacaro: I think you can reject the request before the AuthContext is instantiated.
It sounds like we should not create an AuthContext If the requester has perTenantAcls AND the service being accessed requires cluster-scoped resources.

eduwercamacaro: My proposal is basically move step 1 to the point where we create AuthContext

server/src/test/java/io/littlehorse/common/model/metadatacommand/PrincipalAdministrationTest.java
eduwercamacaro: I see, yes. This test case should be replaced with those two tests.

server/src/main/java/io/littlehorse/server/auth/RequestAuthorizer.java
Snarr: Ok, I think the ACLVerifier class is the place where we are checking the service class's @Authorize annotation metadata. And the annotation metadata is necessary to satisfy Required Information #2 "Know whether or not the service being accessed requires cluster-scoped resources".

So I propose we add a new public method to the ACLVerifier that can return information about the service being accessed, like whether or not it requires cluster scoped resources.

eduwercamacaro: Agree. I think you can add that method in the acl verifier that basically will help you to decide whether the request should be rejected or not before creating the AuthContext.

coltmcnealy-lh (approved):

server/src/main/java/io/littlehorse/common/AuthorizationContextImpl.java
```diff
@@ -12,17 +12,20 @@ public final class AuthorizationContextImpl implements AuthorizationContext {
     private final TenantIdModel authorizedTenant;
 
     private final PrincipalIdModel authorizedPrincipalId;
-    private final List<ServerACLModel> acls;
+    private final List<ServerACLModel> globalAcls;
+    private final List<ServerACLModel> perTenantAcls;
```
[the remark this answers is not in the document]
Snarr: Correct, the AuthorizationContext refers to the context under the current Principal and Tenant pair, so storing the ACLs of other tenants is not necessary.

server/src/test/java/io/littlehorse/server/auth/RequestAuthorizerTest.java
```diff
@@ -177,6 +179,36 @@ public void supportRequiredAclValidation() {
...
+            when(mockCall.getMethodDescriptor()).thenReturn(mockMethod);
+            when(mockMethod.getBareMethodName()).thenReturn("PutPrincipal");
+            when(mockMetadata.get(ServerAuthorizer.CLIENT_ID)).thenReturn("tenant-admin-principal");
+            startCall();
+            ArgumentCaptor<Status> statusCaptor = ArgumentCaptor.forClass(Status.class);
+            Mockito.verify(mockCall).close(statusCaptor.capture(), eq(mockMetadata));
+            assertThat(statusCaptor
+                    .getValue()
+                    .equals(Status.PERMISSION_DENIED.withDescription(
+                            "Missing permissions [WRITE_METADATA] over resources [ACL_PRINCIPAL]")));
```
Snarr, line 209: Please take a look at these 2 tests in RequestAuthorizer

server/src/test/java/io/littlehorse/common/model/metadatacommand/PrincipalAdministrationTest.java
```diff
@@ -270,12 +271,76 @@ void shouldNotAllowPrincipalWithoutGlobalACLsToHaveAnACLThatPointsToTenantResour
...
+        ArgumentCaptor<Exception> exceptionArgumentCaptor = ArgumentCaptor.forClass(Exception.class);
+
+        verify(server).sendErrorToClient(eq(command.getCommandId()), exceptionArgumentCaptor.capture());
+
+        Exception thrown = exceptionArgumentCaptor.getValue();
+        assertThat(thrown)
+                .isNotNull()
+                .isInstanceOf(LHApiException.class)
+                .hasMessage("PERMISSION_DENIED: Missing permission WRITE_METADATA over resource ACL_TENANT.");
     }
```
Snarr, line 341: And these two in PrincipalAdministration

server/src/test/java/io/littlehorse/common/model/metadatacommand/PrincipalAdministrationTest.java
```diff
@@ -270,12 +271,76 @@ void shouldNotAllowPrincipalWithoutGlobalACLsToHaveAnACLThatPointsToTenantResour
...
+        Exception thrown = exceptionArgumentCaptor.getValue();
+        assertThat(thrown)
+                .isNotNull()
+                .isInstanceOf(LHApiException.class)
+                .hasMessage("PERMISSION_DENIED: Missing permission WRITE_METADATA over resource ACL_PRINCIPAL.");
+    }
+
+    @Test
+    public void shouldNotAllowPrincipalWithAllResourcePerTenantACLToAccessTenantResource() {
+        StoredGetable storedRequester =
```
KarlaCarvajal, line 320: Avoid using production code in tests, later it is needed to mantain both places and maybe we can be lost on what would be a real problem, something that can be done is setup an integration test to see the real behavior under certain conditions.

PR #1042, opened 2024-10-03
feat(sdk-java): add method to return all required workfloweventdef names

coltmcnealy-lh:


sauljabin (approved):

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowImpl.java
```diff
@@ -55,6 +57,10 @@ public void addExternalEventDefName(String eedName) {
         requiredEedNames.add(eedName);
     }
 
+    public void addWorkflowEventDefName(String name) {
```
eduwercamacaro, line 60: This should be private

coltmcnealy-lh: It cannot be. That is because the `WorkflowThreadImpl` has to call it. Also, note that this is the `Impl` not the `Workflow`.

See `addTaskDefName`

eduwercamacaro: I meant package private

coltmcnealy-lh: Package private is possible. Doing that.

eduwercamacaro (approved):

PR #1043, opened 2024-10-03
feat(dotnet-sdk): read lh config options from file

KarlaCarvajal:
- Add logic to read lh config options from file
- Add unit tests


sdk-dotnet/LittleHorse.Sdk.Tests/Resources/littlehorse.config
```diff
@@ -0,0 +1,5 @@
+LHW_TASK_WORKER_VERSION=test
```
eduwercamacaro, line 1: I don't think we want to include these temporary files to the test resources. @sauljabin what do you think about this?

sdk-dotnet/LittleHorse.Sdk/Utils/FileManager.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    public class FileManager
+    {
+        internal static Dictionary<string, string> GetDictFromPath(string filePath)
+        {
+            try
+            {
+                Dictionary<string, string> properties = new Dictionary<string, string>();
+                foreach (var tuples in File.ReadLines(filePath)
+                             .Select(line => line.Trim().Split('='))
```
eduwercamacaro, line 11: I think this only cover the happy path. I think we should ignore lines that are not in K=V format

sdk-dotnet/LittleHorse.Sdk/Internal/LHOptionsBinder.cs
```diff
@@ -52,5 +51,60 @@ internal static LHOptions GetOptionsFromEnvironmentVariables()
 
             return options;
         }
+
+        internal static LHOptions GetOptionsFromFile(string filePath)
+        {
+            var options = new LHOptions();
+
+            Dictionary<string, string> lHOptionsInFile = FileManager.GetDictFromPath(filePath);
```
eduwercamacaro, line 59: I see that `FileManager` class is included into a `utils` package, but I don't think this is a utility class. I think `GetDictFromPath` should be a private method inside LHOptions

KarlaCarvajal: In this case is a general implementation that can be reused for any case where it is neede to get a dict from a file path that is the reason of beign a utility, it does not have LHOptions logic

sdk-dotnet/LittleHorse.Sdk/Utils/FileManager.cs
KarlaCarvajal: What is the extra line?, This code first is selecting lines that contains = then subselecting key and values without spaces and finally grouping all by keys

sdk-dotnet/LittleHorse.Sdk/Internal/LHOptionsBinder.cs
eduwercamacaro: LHOptions is the only class that can read a property file. Where do we want to reuse this logic?

sdk-dotnet/LittleHorse.Sdk.Tests/Resources/littlehorse.config
sauljabin: not need it, check this out: https://github.com/littlehorse-enterprises/littlehorse/blob/master/canary/src/test/java/io/littlehorse/canary/config/ConfigLoaderTest.java#L66

Yo can create temp files when executing tests, so my recommendation do not commit this files

sdk-dotnet/LittleHorse.Sdk/Internal/LHOptionsBinder.cs
sauljabin: Is there any reason because we have LHOptions?

sdk-dotnet/LittleHorse.Sdk/LHConfig.cs
```diff
@@ -28,6 +28,14 @@ public LHConfig(ILoggerFactory? loggerFactory = null)
             _options = LHOptionsBinder.GetOptionsFromEnvironmentVariables();
             _createdChannels = new Dictionary<string, LittleHorseClient>();
         }
+        
+        public LHConfig(string configOptionsFilePath, ILoggerFactory? loggerFactory = null)
+        {
+            LHLoggerFactoryProvider.Initialize(loggerFactory);
+            _logger = LHLoggerFactoryProvider.GetLogger<LHConfig>();
+            _options = LHOptionsBinder.GetOptionsFromFile(configOptionsFilePath);
```
sauljabin, line 36: maybe LHConfig is the same that LHOptions

sauljabin: I would say to have a private method inside the LHConfig

sdk-dotnet/LittleHorse.Sdk/Utils/FileManager.cs
sauljabin: I would do:

a class Properties, with a private method parseEntry(string), it receives a string and it split it or ignore it in case it starts with a special character (ex #), it returns a pair entry, it throws an exception if the format is not right. then i would have a put(key, value)  (public), it trims and sets the value into a dict, also a load(path) method it reads all the lines, inside the loop it calls parseEntry, and then it calls the method put.

sdk-dotnet/LittleHorse.Sdk/Internal/LHOptionsBinder.cs
KarlaCarvajal: it is a class where all the options for the LH configuration are placed, I think if we join options in LHConfig the file will be big

sdk-dotnet/LittleHorse.Sdk/LHConfig.cs
KarlaCarvajal: I have 12 configuration options for the client, and they are being used only in the LHConfig file, I can move and remove some other LHConfig properties to make it simpler

sdk-dotnet/LittleHorse.Sdk/Utils/FileManager.cs
coltmcnealy-lh: Is there no Standard Library implementation of parsing a properties file? Go, Java, and Python all have that in the standard library so you don't need to do it yourself, nor do you need a real third-party dependency.

KarlaCarvajal: @coltmcnealy-lh No, dotnet has one for .xml, .json and .ini files :(( https://learn.microsoft.com/en-us/dotnet/core/extensions/configuration-providers

eduwercamacaro: I think littlehorse.config uses a `ini` format.

KarlaCarvajal: yes, it worked with the ini configuration :D

sdk-dotnet/LittleHorse.Sdk/LHInputVariables.cs
```diff
@@ -0,0 +1,114 @@
...
+                LHW_NUM_WORKER_THREADS = IntegerConverter.FromString(numberWorkerThreads);
+            var taskWorkerVersion = Environment.GetEnvironmentVariable("LHW_TASK_WORKER_VERSION");
+            if (!string.IsNullOrEmpty(taskWorkerVersion))
+                LHW_TASK_WORKER_VERSION = taskWorkerVersion;
+        }
+        
+        internal LHInputVariables(string filePath)
+        {
+            IConfigurationBuilder builder = new ConfigurationBuilder()
+                .AddIniFile(filePath, false, true);
```
eduwercamacaro, line 73: I believe third argument should be false because we don't want to change these config values if the file changes.

eduwercamacaro (approved):

sdk-dotnet/LittleHorse.Sdk/LHInputVariables.cs
KarlaCarvajal: done

PR #1054, opened 2024-10-07
perf(server): don't iterate over all events on each externalevent node

coltmcnealy-lh:
- Adds a Tag on wfRunId,externalEventDefId,isClaimed to the ExternalEvent
- Uses that tag when finding the oldest unclaimed ExternalEvent instead of iterating over all associated events
- Makes ReadOnlyGetableManager#iterateOverPrefix() private

Previously, any time we called `ExternalEventRunModel#advance()` (which is basically every time we callled `ThreadRunModel#advance()` when there was an `EXTERNAL_EVENT` node), we iterated through every single `ExternalEvent` associated with that `WfRun`.

What if a user has 1,000+ `ExternalEvent`s in a single `WfRun`? That could get really messy. This PR now makes it only scan one `ExternalEvent`.


server/src/main/java/io/littlehorse/common/model/getable/core/externalevent/ExternalEventModel.java
```diff
@@ -134,6 +149,9 @@ public List<IndexedField> getIndexValues(String key, Optional<TagStorageType> ta
             case "isClaimed" -> {
                 return List.of(new IndexedField(key, this.isClaimed(), tagStorageType.get()));
             }
+            case "wfRunId" -> {
+                return List.of(new IndexedField(key, this.getId().getWfRunId().toString(), tagStorageType.get()));
```
coltmcnealy-lh, line 153: @Snarr and @eduwercamacaro before 1.0 we NEED to figure out how to escape tags.

PR #1073, opened 2024-10-28
[DRAFT] feat(server): Initialize resources at startup

Snarr:


schemas/littlehorse/service.proto
```diff
@@ -1423,8 +1423,13 @@ message GetLatestWfSpecRequest {
   optional int32 major_version = 2;
 }
 
-// The version of the LH Server according to Semantic Versioning
+// Get the current Server Version
 message ServerVersionResponse {
```
coltmcnealy-lh, line 1427: Why is this change necessary? It's a breaking change and would break old `lhctl` versions who run `lhctl version` on a new server.

schemas/internal/init.proto
```diff
@@ -0,0 +1,30 @@
...
+  // If the anonymous Principal was created, this field will include
+  // its initial state
+  Principal init_anonymous_principal = 3;
+  
+  // If the default Tenant was created, this field will include
+  // its initial state
+  Tenant init_default_tenant = 4;
+  
+  // Just for fun?
+  string pedro = 5;
```
coltmcnealy-lh, line 29: `string obi_wan = 5;`

schemas/internal/init.proto
```diff
@@ -0,0 +1,30 @@
...
+option java_multiple_files = true;
+option java_package = "io.littlehorse.common.proto";
+
+import "google/protobuf/timestamp.proto";
+import "service.proto";
+import "acls.proto";
+
+// An object containing important server configurations that will be stored
+// in RocksDB and referenced after server restarts or rebalances.
+message InitConfig {
```
coltmcnealy-lh, line 13: Super pedantic comment. I think the `Config` part might be misleading. This isn't a configuration, but more of a log or "journal entry" of what happened.

Maybe we call it `InitializationHistory`?

schemas/internal/storage.proto
```diff
@@ -22,6 +22,7 @@ enum StoreableType {
     TAG = 5;
     PARTITION_METRICS = 6;
     METRIC_WINDOW = 7;
+    INIT_CONFIG = 8;
```
coltmcnealy-lh, line 25: Same comment here, if we rename it, probably `INITIALIZATION_HISTORY`.

server/src/main/java/io/littlehorse/common/model/getable/core/init/ServerVersionModel.java
```diff
@@ -0,0 +1,44 @@
...
+import io.littlehorse.server.streams.topology.core.ExecutionContext;
+
+public class ServerVersionModel extends LHSerializable<ServerVersion> {
+
+    private int major_version;
+    private int minor_version;
+    private int patch_version;
+
+    public ServerVersionModel() {}
+    ;
```
coltmcnealy-lh, line 16: :eyes:

server/src/main/java/io/littlehorse/common/model/getable/core/init/ServerVersionModel.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public ServerVersionModel(int major_version, int minor_version, int patch_version) {
```
coltmcnealy-lh, line 18: This is python variable naming. Can you please use `camelCase` since this is Java?

server/src/main/java/io/littlehorse/server/streams/topology/core/WfService.java
```diff
@@ -111,29 +111,20 @@ public PrincipalModel getPrincipal(PrincipalIdModel id) {
...
-        if (storedResult == null && id.getId().equals(LHConstants.ANONYMOUS_PRINCIPAL)) {
-            // This means that all of the following are true:
-            // - The `anonymous` Principal has not yet been modified by the customer.
-            // - We just implicitly create it.
-            return createAnonymousPrincipal();
+        PrincipalModel principalModel = metadataManager.get(id);
+
+        if (principalModel == null && id.getId().equals(LHConstants.ANONYMOUS_PRINCIPAL)) {
+            // If Anonymous Principal missing from store (should never happen...)
+            // Implicitly create empty Anonymous Principal
```
coltmcnealy-lh, line 118: "Should Never Happen" is a code smell. Can you throw an `IllegalStateException`? Maybe @eduwercamacaro would have a better idea?

However, I disagree about "should never happen." What if we have 2 servers on version 0.11.2 in our cluster, and we start a rolling restart to get to 1.0.

If we first restart the server that _does not_ have the Metadata Partition assigned to it, then when it restarts, we'll have:
* 1.0 version server that doesn't own the Metadata Partition
* 0.11.2 version server that does own the Metadata Partition
* No `InitializationHistory` or `InitConfig` record created
* No `anonymous` Principal.

This tells me that we probably _do_ want to "create anonymous from thin air" in that case. But, as we said, it's a security vulnerability....what Kafka does is do _two_ rolling restarts. The first one is more lenient so "it works," then the second one changes a configuration so it would throw `IllegalStateException`.

server/src/main/java/io/littlehorse/server/streams/topology/core/processors/MetadataProcessor.java
```diff
@@ -42,6 +52,48 @@ public MetadataProcessor(LHServerConfig config, LHServer server, MetadataCache m
...
+
+        ClusterScopedStore clusterStore =
+                ClusterScopedStore.newInstance(ctx.getStateStore(ServerTopology.METADATA_STORE), context);
+
+        InitConfigModel oldInitConfigModel =
+                clusterStore.get(InitConfigModel.SERVER_INITIALIZED_KEY, InitConfigModel.class);
+
+        // If server has not been initialized..
+        if (oldInitConfigModel == null) {
+            log.info("Initializing Server...");
```
coltmcnealy-lh, line 70: Nit: This is "initializing Cluster" not "server". Why? It happens once per cluster, so not all Servers should do this.

coltmcnealy-lh (commented):
A few minor problems but this is overall a very good start!

schemas/internal/init.proto
```diff
@@ -0,0 +1,30 @@
...
+option java_multiple_files = true;
+option java_package = "io.littlehorse.common.proto";
+
+import "google/protobuf/timestamp.proto";
+import "service.proto";
+import "acls.proto";
+
+// An object containing important server configurations that will be stored
+// in a KeyValueStore and referenced after server restarts or rebalances.
+message InitializationLog {
```
KarlaCarvajal, line 13: This message InizializationLog could be called InitialStateConfigs

server/src/main/java/io/littlehorse/common/LHConstants.java
```diff
@@ -52,6 +55,11 @@ public class LHConstants {
     public static final String DEFAULT_TENANT = "default";
     public static final String ANONYMOUS_PRINCIPAL = "anonymous";
 
+    public static final ServerACL ADMIN_ACL = ServerACL.newBuilder()
```
KarlaCarvajal, line 58: It looks like ADMIN_ACL it is not a constant, ACLs will change depending on the Principals and they should be rehused for the different cases

eduwercamacaro (approved):

docs/docs/04-concepts/13-principals-and-tenants.md
```diff
@@ -34,19 +34,75 @@ The rules for authenticating a call are as follows:
...
+```
+
+### The `anonymous` Principal
+
+An authenticated request made with an unknown `PrincipalId` is authorized with the `anonymous` `Principal`. 
+
+#### Initial Configuration
+
+:::danger
+The `anonymous` `Principal` should **ALWAYS** be demoted from this initial configuration before your LittleHorse Cluster is used in production.
```
coltmcnealy-lh, line 74: I wouldn't say `**ALWAYS**`. Some people (including one of our first users) use LittleHorse on trusted networks without authentication and therefore use only the `anonymous` principal.

Maybe
> On untrusted networks, we strongly recommend demoting the privileges of the `anonymous` Principal in order to enforce the principle of least privilege.

coltmcnealy-lh:
Fixed #1082

PR #1112, opened 2024-11-07
fix(dashboard): better json view for `TaskRun`

### LINKED ISSUE #1066: Dashboard Support JSON for description ex: (Inspect TaskRun)
### Context

image

### Steps To Reproduce

Inspect any `TaskRun` in a `WfRun`

### Expected Behavior

We should provide a better UX for viewing the output of a task like a `JSON` object

### Screenshots

_No response_

### Additional Context

_No response_

### Components

_No response_

bryson-g:
resolves #1066

**Before:**

image

**After:**

image


HazimAr:
Can you please provide screenshots for other examples that show this JSON stringify, maybe something like `WfRunVariable` or `ExternalEvent` output?

Because you changed it across all variable values I would like to see screenshots for those examples as well.

HazimAr (commented):
Can you please provide screenshots for other examples that show this JSON stringify, maybe something like `WfRunVariable` or `ExternalEvent` output?

Because you changed it across all variable values I would like to see screenshots for those examples as well.

dashboard/src/app/utils/variables.ts
```diff
@@ -16,6 +16,13 @@ export const getVariableValue = (variable?: VariableValue) => {
 
   if (variable.bytes) {
     return '[bytes]'
+  } else if (variable.str) {
+    try {
+      const json = JSON.parse(variable.str)
+      return JSON.stringify(json, null, 4)
+    } catch {
+      return variable[key]
+    }
```
mijailr, line 25: What about `variable.jsonObj` and `variable.jsonArr`?

mijailr (changes requested):

mijailr:
> Can you please provide screenshots for other examples that show this JSON stringify, maybe something like `WfRunVariable` or `ExternalEvent` output?
> 
> 
> 
> Because you changed it across all variable values I would like to see screenshots for those examples as well.

That's a good point, this same function is being used across the dashboard. So please verify every use and check if the new behavior fit on every case.

mijailr (dismissed):
Giving it a second look, I believe this change shouldn't be implemented in this utility function but intentionally on those views where a special rendering is required.

HazimAr:
> Giving it a second look, I believe this change shouldn't be implemented in this utility function but intentionally on those views where a special rendering is required.

Yeah I think we should see what it looks like on all the JSON parsing as it might be better as a default. I'm sure there are other sections of the dashboard that have JSON that needs to be formatted. 🤷‍♂️

dashboard/src/app/utils/variables.ts
HazimAr: Good catch

HazimAr (changes requested):

bryson-g:
> Giving it a second look, I believe this change shouldn't be implemented in this utility function but intentionally on those views where a special rendering is required.

Mijaíl 👋, I'm curious about your thought processes behind this. Is there any instance where you think we shouldn't display formatted JSON to the user? Do you think formatted JSON would bug some functionality in some way?

I understand this variable util function has a lot of references, so it is hard to predict exactly how it interacts with everything, but let me know what you're thinking.

mijailr:
Since our last talk about this, is this ready for review with the requested changes?

dashboard/src/app/(authenticated)/(diagram)/components/Modals/NodeRun/TaskDefDetail.tsx
```diff
@@ -144,7 +144,9 @@ export function AttemptErrorExceptionOutput({ attempt }: { attempt: TaskAttempt
       <pre className="overflow-auto">
         {attempt.error && attempt.error.message}
         {attempt.exception && attempt.exception.message}
-        {attempt.output && getVariableValue(attempt.output)}
+        {attempt.output && typeof getVariableValue(attempt.output) === 'string'
+          ? formatJsonOrReturnOriginalValue(getVariableValue(attempt.output) as string)
+          : String(getVariableValue(attempt.output))}
```
HazimAr, line 149: Good job on turning this into a string so `false` will show up on the UI.

HazimAr (approved):

HazimAr:
@mijailr I will wait for your approval and than feel free to merge after.

Issue #582, opened 2024-01-08
Mark Variable as `.internal()`


coltmcnealy-lh:
### Context

I recently implemented a major overhaul to `WfSpec` versioning in which we added the `majorVersion` and `revision` fields. A new `majorVersion` happened when there was a breaking change, defined as a change in the set of `searchable()` or `required()` variables.

### Description

What if there is a `Variable` that is not searchable (for performance reasons), and not an input variable, but it _is_ an important part of the Contract of the `WfSpec` (for example when using `WfRun`'s to store business entitites, this variable might be a useful "column" but not indexed or required)?

I think we should make any changes to the sets of `Variable`s in a `WfRun` result in a new `majorVersion`, unless the variable is marked as `internal()`.

All three SDK's should have a `WfRunVariable#internal()` method

### Technical Context

Certain variables are automatically created by the SDK, and should also be marked as `internal()`. For example:

- internal variables created by the `WorkflowThread#spawnThread()` method
- The `INPUT` variables used for receiving `ExternalEvent`'s in an Interrupt Handler, and `Failure`'s in a Failure Handler

coltmcnealy-lh:
@eduwercamacaro agrees; also we should should update our documentation. This should be done before `1.0.0` as it does break our contract a little bit (but not a huge amount)

coltmcnealy-lh:
Well, it looks like this requires some discussion to get the correct API semantics, but no one wants to do the implementation 😂

coltmcnealy-lh:
Okay, I looked at the protobuf structure. #605 introduced the `enum WfRunVariableAccessLevel` as follows:

```
// Determines the Access Level for a Variable in a ThreadSpec/WfSpec.
enum WfRunVariableAccessLevel {
  // A `PUBLIC_VAR` can be accessed (read + mutated) by child `WfRun`'s.
  PUBLIC_VAR = 0;

  // A `PRIVATE_VAR` cannot be accessed by a child `WfRun`.
  PRIVATE_VAR = 1;

  // An `INHERITED_VAR` is inherited from the parent `WfRun`. Only valid in a `WfSpec` that
  // has a parent. Also can only be declared in the Entrypoint Thread.
  INHERITED_VAR = 2;
}
```

I was originally going to add a `bool internal = 6;` to the `message ThreadVarDef`. However, that structure matches perfectly with the motivation of this ticket. We need to do the following:

[...]

PR #1118, opened 2024-11-10
feat(server)!: changes variable access level api

coltmcnealy-lh:
- Fixes ticket #582
- Default access level is now PRIVATE
- Only PUBLIC_VAR and required input variables are considered in WfSpec versioning
- Only PUBLIC_VAR can be frozen


sauljabin (approved):
No especial comments from my side

PR #1123, opened 2024-11-13
feat(dashboard): debloat wfspec view

### LINKED ISSUE #941: Group wfSpec versions in the dashboard
### Context

We're currently listing wfSpec by version which causes to have multiple pages of the same wfSpec. We can handle this in the UI by increasing the default pagination size to a bigger number and then group the versions as a single item. 

image
 

### Acceptance Criteria

**Scenario** More than 10 versions
**Given** a wfSpec with more than 10 versions
**When** the page limit is 10
**Then** it should render just one item

### Out of Scope

_No response_

### Components

Dashboard

bryson-g:
resolves #941

## Before
image

## After
https://github.com/user-attachments/assets/bc278637-57c1-4833-a4f6-9663b90098fe

## Explanation
- Simplified `WfSpec` page to show each unique `WfSpec` name once with its latest version
- Clicking a `WfSpec` redirects to its latest version page
- Users can access older versions through the existing version dropdown on the `WfSpec` page

image

## Considerations
The version dropdown may become difficult to navigate as versions accumulate. Potential improvements:
- Add search functionality
- Organize by major version categories
- Consider using a command palette interface (similar to `shadcn's` `Command` component)
https://ui.shadcn.com/docs/components/command
image

### First Draft
As a first draft, I was considering a concept like this:

https://github.com/user-attachments/assets/44592cb3-da1a-4871-a5aa-d8e05951a91e

While the above first draft solution technically does resolve the original problem, I realized there could be a more eloquent approach.

### Potentially Another Design

I know the data provided to display `workflows` is fairly minimal _(just versions and workflow names)_, but if it were ever expanded to include more detailed information, such as last modified or last ran, then it may be worth considering a more table orientated design. Additionally, this approach would allow for an intuitive way to filter for specific `WfSpecs`, provided there is relevant data to filter through. 

`Shadcn's` table example:
https://ui.shadcn.com/docs/components/table
image


coltmcnealy-lh:
Thanks @bryson-g! You get 10 points for the best PR description I've ever seen.

I'm not a frontend developer, so I can't review the code, but Hazim or Mijaíl will get to it tomorrow.

dashboard/src/app/(authenticated)/components/Search.tsx
```diff
@@ -35,7 +35,7 @@ export const Search: FC<{}> = () => {
           <RefreshCwIcon className="h-8 w-8 animate-spin text-blue-500" />
         </div>
       ) : (
-        <div className="min-h-[360px]">
```
HazimAr, line 38: Any reason why to remove the min height. I believe it was there so the footer with the "data per load" select would be far below if there was only one WfSpec

dashboard/src/app/(authenticated)/components/VersionSelector.tsx
```diff
@@ -21,17 +21,19 @@ export const VersionSelector: FC<Props> = ({ path, currentVersion, versions, loa
...
-                className="relative block cursor-pointer select-none p-2 hover:bg-slate-300"
-                key={version}
-                value={version}
-                as={Link}
-                href={`${path}/${version}`}
-              >
-                {version}
-              </Listbox.Option>
-            ))}
+            {[...versions].reverse().map(version => {
```
HazimAr, line 24: Nice job using `.reverse()` instead of `.sort` as it is exponentially more efficient 😎

dashboard/src/app/(authenticated)/components/tables/WfSpecTable.tsx
```diff
@@ -1,30 +1,57 @@
...
-      {pages.map((page, i) => (
-        <Fragment key={i}>
-          {page.results.map(({ name, majorVersion, revision }: WfSpecId) => (
-            <div key={`${name}.${majorVersion}.${revision}`} className="my-2 flex gap-2">
-              <Link className="underline hover:no-underline" href={`/wfSpec/${name}/${majorVersion}.${revision}`}>
-                {name}
-              </Link>
+      <div className="flex max-h-[600px] flex-col overflow-auto">
+        {wfSpecs.map(wfSpec => (
+          <Fragment key={wfSpec.name}>
```
HazimAr, line 35: Good job using the `wfSpec.name` as the key instead of the `index` I believe we should use `wfSpec.id` though because there might be multiple `WfSpec`s with the same name

Is what I was going to say but I realize that that `wfSpec.name` is unique as the map is pointed to the latest version. So good job!

HazimAr (approved):

dashboard/src/app/(authenticated)/components/Search.tsx
bryson-g: I removed the `min-h-[360px]` to bring the load footer closer to the end of the `WfSpec` list for these reasons:

- Less mouse travel time
- More visibly obvious to the user
- Less white space looks cleaner

Even with just one `WfSpec,` I'm not sure how a distanced footer would be more beneficial. I'm always open to hear other's thoughts, so please let me know what you think.

HazimAr: No that make sense I was just curious why the change but what you said tracks. 😎

dashboard/src/app/(authenticated)/components/tables/WfSpecTable.tsx
```diff
@@ -1,30 +1,57 @@
...
+      .reverse()
+      .forEach(({ name, majorVersion, revision }: WfSpecId) => {
+        if (!specMap.has(name)) {
+          specMap.set(name, { name, latestVersion: `${majorVersion}.${revision}` })
+        }
+      })
+
+    return Array.from(specMap.values())
+  }, [pages])
+
```
mijailr, line 30: I see there might be cases where the results might not be complete due to pagination. So I suspect it would get the latest version wrong. 

Have you tried that?

dashboard/src/app/(authenticated)/components/Search.tsx
mijailr: The default space is there for view consistency user can predict where the position of elements would be so I would prefer to keep it.

dashboard/src/app/(authenticated)/components/tables/WfSpecTable.tsx
bryson-g: > I see there might be cases where the results might not be complete due to pagination. So I suspect it would get the latest version wrong.
> 
> Have you tried that?

This is a valid issue that you bring up. After analyzing the data that is sent to the front end page, I don't believe there is a suitable solution for us here. It seems like a limitation from the server and the data being sent. 

How do you think we could solve this?

HazimAr: This is also an issue the server has, @coltmcnealy-lh and I talked about this in terms of of trying you query user tasks for "the latest 10" and the problem is the server can't do that. I'm not sure if this issue is related but I believe this issue might be easier to solve as the server could have a different RPC that just sends WFSpec and its latest version instead of each individual WfSpec

bryson-g: To clarify the situation here for anyone else reading this:

image

There's a pagination problem with WfSpecs display. The page shows WfSpecs with a limit on how many items load at once (`items per load`), but the UI incorrectly assumes that the most recent version of a WfSpec will always be within this loaded set.

Example of the problem:
If items_per_load = 10
- The latest version might actually be the 11th item or later
- Current UI would miss this latest version by only looking at the first 10 loaded items

This is a design flaw because the UI needs to account for cases where the latest version exists beyond the initial pagination limit.

coltmcnealy-lh: There is `rpc GetLatestWfSpec` that gets the latest version of a `WfSpec`

coltmcnealy-lh: @bryson-g and @HazimAr  check out this RPC: https://littlehorse.dev/docs/api#getlatestwfspec

coltmcnealy-lh: We can also add a flag to `rpc SearchWfSpec` that filters so that it only returns the latest version of each `WfSpec`. That might be the easiest solution, but it requires backend changes. @HazimAr maybe that would be a good "warmup" for you to get used to the internals of the server.

Alternatively, this would probably be a half-day of effort for @Snarr

HazimAr: I agree it would be a good warmup but I don't think that is necessary now

Also if we used GetLatestWfSpec then it will fix the problem of being uncertain of the latest version but the per load will still be off 

Because it would return 10 items but 8 are unique so it would call 8 times to get the latest version

HazimAr (changes requested):
Lets account for the pagination case and RPC suggestion @coltmcnealy-lh gave.

bryson-g:
@mijailr, Continuing from the conversation we had:

As one of the potential solutions for this pagination issue, we considered paginating through every page to ensure the latest workflow for each unique WfSpec.

However, I believe it would be more efficient to use the `GetLatestWfSpec` RPC instead. For example, loading the first 10 items and filtering for unique names, then calling the `GetLatestWfSpec` RPC for each listed WfSpec.

The one problem I see with the latter approach is that the `Items per load` number may be lower than the number of `WfSpecs` that are actually rendered on the page. This may be slightly confusing for users. When compared to the former approach, we already paginate through all of the `WfSpecs`, so we could display all unique `WfSpecs` in a scroll area and completely remove the `Items per load` option.

What are your thoughts on how exactly how we should proceed?

HazimAr (approved):

HazimAr:
@bryson-g please fix tests and then its ready to merge

bryson-g:
@HazimAr tests passed now. had to delete `package-json.lock` and push a new clean install. lmk if this is fine with u

PR #1126, opened 2024-11-13
feat(dashboard): adds lhd_oauth_callback_url_internal

MateoRojas:
Allows to configure `NEXTAUTH_URL_INTERNAL` in case `NEXTAUTH_URL` cannot be reached by the dashboard server. This is important because the dashboard server calls itself when rendering the OAuth providers and in certain cases, like local development within kind the `NEXTAUTH_URL` cannot be reached.


mijailr (approved):

coltmcnealy-lh:
Good work

PR #1133, opened 2024-11-16
fix(dashboard): resolve expression correctly

bryson-g:


dashboard/src/app/(authenticated)/(diagram)/components/Modals/TaskRun.tsx
```diff
@@ -91,7 +91,7 @@ function AttemptErrorExceptionOutput({ attempt }: { attempt: TaskAttempt }) {
       <pre className="overflow-auto">
         {attempt.error && attempt.error.message}
         {attempt.exception && attempt.exception.message}
-        {attempt.output && getVariableValue(attempt.output)}
+        {attempt.output && String(getVariableValue(attempt.output))}
```
bryson-g, line 94: This will still not render if `attempt.output` is not a truthy value. Still thinking about how this should be resolved. 

I think removing the left-hand side of the expression on line 94 would be fine?

HazimAr:
this will be resolved with #1112

PR #1134, opened 2024-11-16
feat(dashboard): Execute WfSpec from dashboard

muhammadqamar:
so we need to execute wfSpec from dashboard, there will be a button on WfSpec page for each version, 
The button will open a dialog modal and show all available variables required and public. after form validation , we will get WfRun in response and will redirect to WfRun page where we can see all variables values.
we also have option to mark a field as null. there is a checkbox which will allow the value to be null. 
image
image
image


mijailr (dismissed):
The form should be handled with Form Actions. Use this as an example or inspiration https://github.com/vercel/next.js/tree/canary/examples/next-forms

muhammadqamar:
> The form should be handled with Form Actions. Use this as an example or inspiration https://github.com/vercel/next.js/tree/canary/examples/next-forms

as per example. that would be part of react 19
// useActionState is available with React 19 (Next.js App Router)

HazimAr:
> > The form should be handled with Form Actions. Use this as an example or inspiration https://github.com/vercel/next.js/tree/canary/examples/next-forms
> 
> as per example. that would be part of React 19 // useActionState is available with React 19 (Next.js App Router)

`useActionState` is a new feature you don't need to implement next form actions

Form actions were introduced in Next 13 I believe but I know they are available with our version (Next 14)

https://nextjs.org/docs/app/api-reference/components/form#action-function-props

HazimAr (changes requested):
The `useActionState` is a new feature you don't need for implementing next form actions. However, we should use it when upgrading to Next 15.

Form actions were introduced in Next 13 as a part of the bigger app router upgrade I believe but I know they are available with our version (Next 14)

https://nextjs.org/docs/app/api-reference/components/form#action-function-props

Please also use Shad/cn implementation of `react-hook-form` instead of plain `react-hook-form`
https://ui.shadcn.com/docs/components/form

muhammadqamar:
> Currently, users have to input their `JSON` in plain string format. This comes with a ton of downsides
> 
> * For people who are not technical they wont really understand what JSON is or how to use it.
> * Users can mess up the JSON schema and miss a `}` or `,` very easily
> * For `JSON_ARR` its hard to see all your objects you have in the array
> 
> `JSON_ARR` & `JSON_OBJ` need better UX that allows people to use UI instead of passing in a plain string.
> 
> This will
> 
> * we can sanitize the JSON to make sure it's correct before sending it to the server.
> * people can mess up JSON if they use a plain string.
> * less technical people will be able to interact with it.
> * they should be able to `add` and `remove` objects or properties with buttons.
> 
> We do not have to force the user to use the UI for building `JSON` objects, we can have a "advanced mode" that will allow the user to just put in plain text.
> 
> This doesn't have to be in the same PR if we want to push this way to prod, but we need to keep track of that improvement if we decide.

agreed, that can be part of another enhancement, but we added check on FE if there is any syntax issue while writing object or array of ibjects

Snarr (approved):

HazimAr (approved):

Issue #1137, opened 2024-11-16
Move Tenant Information to the URL


HazimAr:
Currently we store the tenantId in cookie but we should move it to the URL.
This will allow users to be able to share specific page that has specific tenant data with other users.
We still need to allow the user to be able to change the current tenant to any tenant they have access to

PR #1139, opened 2024-11-18
refactor(dashboard): tenant cookie -> dynamic route slug

bryson-g:
resolves https://github.com/littlehorse-enterprises/littlehorse/issues/1137

- Added `[tenantId]` dynamic route
- Moved file directories
- Updated broken imports
- Modified Navigation component to prepend `tenantId` to the passed `href` prop.
  - This was done to make the component less tedious and easier to use, but I think we may want to force the href to include the slug anyways to avoid potential confusion. LMK.
 - Support root page redirect to `default` `tenantId` within `next.config.mjs`
- Ensuring every `redirect` & `href` is accounted for with the new route to maintain functionality


sauljabin:
> @sauljabin @coltmcnealy-lh This will help us with detecting metrics for each tenant.

really good catch, yes

dashboard/src/contexts/WhoAmIContext.tsx
```diff
@@ -23,7 +23,7 @@ export const WhoAmIContext: FC<PropsWithChildren<WhoAmIContextProps>> = ({ child
     if (!tenantId) {
       setTenant(tenants[0])
     }
-  })
+  }, [])
```
HazimAr: good catch as this would run on every render and we only need this to run 1 time

dashboard/src/app/page.tsx
```diff
@@ -0,0 +1,17 @@
+import { redirect } from 'next/navigation'
+import getWhoAmI from './getWhoami'
+
+export default async function Home() {
+  const whoAmI = await getWhoAmI()
```
HazimAr, line 5: I was thinking why not do this in middleware but yes Next.js doesn't support node apis in `middleware.ts`

HazimAr (approved):
You had some missing hrefs that needed to be updated with the new [tenantId] url so I went ahead and fixed everything.

Also for future reference tenantId when being used in the code should be coming from the URL not from `whoamI` context or the cookie. I have already fixed that as well

dashboard/src/contexts/WhoAmIContext.tsx
HazimAr: I removed this as this was only for setting the cookie.

PR #1142, opened 2024-11-18
fix(server): Update exception when anonymous principal not found in store

Snarr:
Adjusts exception message to be more descriptive and use gRPC Status Code UNAVAILABLE when Anonymous Principal not found in store.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streams/topology/core/WfService.java
+++ b/server/src/main/java/io/littlehorse/server/streams/topology/core/WfService.java
@@ -1,6 +1,8 @@
 package io.littlehorse.server.streams.topology.core;
 
+import io.grpc.Status;
 import io.littlehorse.common.LHConstants;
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/littlehorse/server/streams/topology/core/WfService.java
```diff
@@ -113,7 +117,9 @@ public PrincipalModel getPrincipal(PrincipalIdModel id) {
 
         if (principalModel == null && id.getId().equals(LHConstants.ANONYMOUS_PRINCIPAL)) {
             // If Anonymous Principal missing from store (should never happen...)
-            throw new IllegalStateException("Anonymous Principal missing from store...");
+            log.warn(
```
coltmcnealy-lh, line 120: This could almost be `info` or `debug`

coltmcnealy-lh (approved):

PR #1158, opened 2024-11-24
fix(server): Support blank `SearchPrincipalRequest`

### LINKED ISSUE #1152: Support Blank `SearchPrincipalRequest`
All of our other metadata search rpc's (`Search{WfSpec,TaskDef,ExternalEventDef,UserTaskDef,WorkflowEventDef,Tenant}` just show a list of all available ID's if you don't specify what you're searching for.

The following behavior (what we have currently) breaks that tradition and is also annoying from the user perspective.
```
->lhctl search principal
Error: at least one of the flags in the group [tenantId isAdmin] is required
Usage:
  lhctl search principal [flags]

Flags:
      --earliestMinutesAgo int   Search only for Principals that were created no more than this number of minutes ago (default -1)
  -h, --help                     help for principal
      --isAdmin                  List only Principals that are admins
      --latestMinutesAgo int     Search only for Principals that were created at least this number of minutes ago (default -1)
      --tenantId string          List Principals associated with this Tenant ID

Global Flags:
      --bookmark bytesBase64   Optional bookmark for paginated scans.
      --configFile string      Configuration File Location (default "${HOME}/.config/littlehorse.config")
      --limit int32            Guideline for number of response items to fetch per request. (default 100)
```

Snarr:
Removes requirement for one of `principal_criteria` field to be submitted with each `SearchPrincipalRequest` from server and lhctl.

Resolves #1152.

### DIFF
--- a/lhctl/internal/principal.go
+++ b/lhctl/internal/principal.go
@@ -200,7 +200,6 @@ func init() {
 	searchPrincipalCmd.Flags().Bool("isAdmin", false, "List only Principals that are admins")
 	searchPrincipalCmd.Flags().Int("earliestMinutesAgo", -1, "Search only for Principals that were created no more than this number of minutes ago")
 	searchPrincipalCmd.Flags().Int("latestMinutesAgo", -1, "Search only for Principals that were created at least this number of minutes ago")
-	// searchPrincipalCmd.MarkFlagsOneRequired("tenantId", "isAdmin")
 
 	deployCmd.AddCommand(deployPrincipalCmd)
 	deleteCmd.AddCommand(deletePrincipalCmd)
--- a/lhctl/internal/principal.go
+++ b/lhctl/internal/principal.go
@@ -1,11 +1,12 @@
 package internal
 
 import (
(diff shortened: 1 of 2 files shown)


lhctl/internal/principal.go
```diff
@@ -199,7 +200,7 @@ func init() {
 	searchPrincipalCmd.Flags().Bool("isAdmin", false, "List only Principals that are admins")
 	searchPrincipalCmd.Flags().Int("earliestMinutesAgo", -1, "Search only for Principals that were created no more than this number of minutes ago")
 	searchPrincipalCmd.Flags().Int("latestMinutesAgo", -1, "Search only for Principals that were created at least this number of minutes ago")
-	searchPrincipalCmd.MarkFlagsOneRequired("tenantId", "isAdmin")
+	// searchPrincipalCmd.MarkFlagsOneRequired("tenantId", "isAdmin")
```
coltmcnealy-lh, line 203: We can delete this line.

Snarr: Good catch

PR #1166, opened 2024-11-27
feat(sdk-python): Port in-line expressions to Python SDK

Snarr:
This PR ports the in-line expression changes to the Python SDK.


sdk-python/littlehorse/workflow.py
```diff
@@ -445,6 +501,21 @@ def searchable_on(
     def required(self) -> "WfRunVariable":
         self._required = True
         return self
+    
+    def with_default(self, default_value: Any) -> WfRunVariable:
+        self.__set_default(default_value)
+
+        return self
+
+    def __set_default(self, default_value: Any) -> None:
```
sauljabin, line 510: 1 underscore for internal 2 for language related ones.

example `__init__` 2 underscore

sauljabin, line 510: 1 underscore for internal 2 for language related ones.

example __init__ 2 underscore

sdk-python/littlehorse/workflow.py
```diff
@@ -284,25 +318,53 @@ def with_json_path(self, json_path: str) -> "NodeOutput":
...
+    def remove_key(self, other: Any) -> LHExpression:
+        return LHExpression(self, VariableMutationType.REMOVE_KEY, other)
 
 
 class WfRunVariable:
     def __init__(
         self,
         variable_name: str,
         variable_type: VariableType,
+        parent: WorkflowThread,
```
sauljabin, line 354: if the default value is None use `= None` here

Optional[WorkflowThread]

sauljabin (approved):
I have approve it, but before merging check if it is possible to get rid of the dependency at WfRunVariable and use the WorkflowCondition.

The other WfRunVariable parameter are attributes of the class, WorkflowThread would be a dependency that maybe we can avoid, if we can the better

sdk-python/littlehorse/workflow.py
Snarr: I agree with your proposal, it is unnecessary to depend on the parent condition methods when creating a new WorkflowCondition is so simple. If the code for creating a new WorkflowCondition were more complex, it may make sense to centralize it in one place and rely on this dependency injection or another class to hold the methods.

There is *one* additional method that may not make sense to port without injecting the parent thread dependency:

```python
    @Override
    public void assignTo(Serializable rhs) {
        parent.mutate(this, VariableMutationType.ASSIGN, rhs);
    }
```

Without injecting the parent WorkflowThread at the constructor, this Thread may not make sense or be easy to use. Instead, we would have to pass it to the method as an argument, which would not save the user much code and (in my opinion) be less readable.

sauljabin: You are right, I did not see the mutate

sdk-python/littlehorse/workflow.py
```diff
@@ -136,6 +136,40 @@ def to_variable_assignment(value: Any) -> VariableAssignment:
     )
 
 
+class LHExpression:
+    def __init__(self, lhs: Any, operation: VariableMutationType, rhs: Any) -> None:
+        self._lhs = lhs
+        self._operation = operation
+        self._rhs = rhs
+        pass
```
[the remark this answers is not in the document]
Snarr: Forgot to remove this, good catch

PR #1169, opened 2024-11-29
fix(sdk-java): remove output schema in task methods with void returned type

KarlaCarvajal:
-  Change LHTaskSignature to not set the output schema when there is not a returned type in LHTaskMethods


Snarr (approved):
LGTM!

PR #1171, opened 2024-12-02
feat(server): allow casting var assns

coltmcnealy-lh:


coltmcnealy-lh:
Closing this as it requires a proper proposal, and I think we want to take a more well thoguht-out approach. #1566 will track the design and implementation of this feature in a better way.

Issue #1090, opened 2024-11-03
Show `UTESaved` User Task Saved Events in User Task Audit Log


coltmcnealy-lh:
Clicking "Inspect UserTaskRun" should show the history of when the user task was saved and by whom it was saved.

PR #1183, opened 2024-12-07
feat(dashboard): user task audit log

bryson-g:
resolves [#1090](https://github.com/littlehorse-enterprises/littlehorse/issues/1090)

## Description
- New page containing a table UI for `UserTask` saved events.
- `View Audit Log` button on the `UserTask` node.
- `View Audit Log` button on the `UserTaskDef` page.
- Global `Details` component that can be used to refactor the project in the future.
  - For refactoring the multiple `Details.tsx` components that are not unified currently.
- Implemented retrieval, parsing, and ordering of saved events data.
- Pagination of events to not overcrowd the screen in large circumstances.

## Considerations
We may want to include more relevant data that users may want to know, such as:
- The saved state input values
- Other events
  - Completion
  - Reassignment
  - Cancellation

### Audit Log Pagination
https://github.com/user-attachments/assets/c99fe89e-b95c-40f6-85ef-50110f8d4f45

### Audit Log No Pagination
image

### No Save History Found
image

### Navigation From UserTaskDef View
https://github.com/user-attachments/assets/466cec04-0c6e-4368-99aa-b440cf7c38f5

### Navigation From WfRun View
https://github.com/user-attachments/assets/be5eec7b-2e74-4bdf-ab9c-45ec38320fea


dashboard/src/app/[tenantId]/(diagram)/components/NodeViewButton.tsx
Comment (HazimAr) [inline dashboard/src/app/[tenantId]/(diagram)/components/NodeViewButton.tsx:1]: If the entire purpose of this button is just to redirect them to another page then you should use the built in `LinkWithTenant`.

Another thing to note is when using buttons make sure to use `Button` from shad/cn and if you want it to look like a link then use the prop `variant`

dashboard/src/app/[tenantId]/userTaskDef/[...props]/components/Versions.tsx
```diff
@@ -1,3 +1,4 @@
+'use client'
```
Comment (HazimAr) [inline dashboard/src/app/[tenantId]/userTaskDef/[...props]/components/Versions.tsx:1]: good catch putting this here even tho the parent component its client, this is best practice in-case this components parent changes to be a server component

dashboard/src/app/[tenantId]/userTaskDef/[...props]/components/UserTaskDef.tsx
```diff
@@ -178,6 +179,17 @@ export const UserTaskDef: FC<Props> = ({ spec }) => {
...
+                          <TableCell>
+                            <Button
+                              onClick={() =>
+                                router.push(
+                                  `/${tenantId}/userTaskDef/audit/${userTaskRun.id?.wfRunId?.id}/${userTaskRun.id?.userTaskGuid}`
+                                )
+                              }
+                            >
+                              View Audit Log
+                            </Button>
```
Comment (HazimAr) [inline dashboard/src/app/[tenantId]/userTaskDef/[...props]/components/UserTaskDef.tsx:191]: Same with this, please use `LinkWithTenant` and if you want the default styling of `Button` then you can use `asChild` prop and put the `Button` as a child of `LinkWithTenant`

dashboard/src/app/[tenantId]/components/LinkWithTenant.tsx
```diff
@@ -1,11 +1,18 @@
...
 
-const LinkWithTenant = ({ ...props }: ComponentProps<typeof NextLink>) => {
+const LinkWithTenant = ({ linkStyle, ...props }: ComponentProps<typeof NextLink> & { linkStyle?: boolean }) => {
   const { tenantId } = useParams()
-  return <NextLink {...props} href={`/${tenantId}${props.href}`} />
+  return (
+    <NextLink
+      {...props}
+      href={`/${tenantId}${props.href}`}
+      className={cn(props.className, { 'text-blue-500 underline': linkStyle })}
```
Comment (HazimAr) [inline dashboard/src/app/[tenantId]/components/LinkWithTenant.tsx:13]: why not just pass "text-blue-500 underline" when using the component. I would be okay with this if we were using this in many places but right now its just one? Even if we were using in many places I don't see a problem with just referencing it there instead of the main component.

HazimAr (changes requested):

dashboard/src/app/[tenantId]/components/LinkWithTenant.tsx
Comment (bryson-g) [reply dashboard/src/app/[tenantId]/components/LinkWithTenant.tsx:13]: even though the hyperlink style isn't being used in multiple places, it is still a likely styling choice that will be used in the future. and adding a prop to style it was intended to avoid the possibility of links being styled differently across the app. like, what if someone used `text-blue-300` instead of `text-blue-500`, or what if they didn't add an underline. 

minor stuff, but just remaining consistent was the idea.

however, upon further review with your other comments, I believe using the `shadcn's` `link` `variant` provides an even greater global styling enforcement across the app.

Comment (HazimAr) [reply dashboard/src/app/[tenantId]/components/LinkWithTenant.tsx:13]: thats a good point i forgot about the the `link` `variant`. I agree with your first paragraph if there was not variant though.

Comment (bryson-g) [reply dashboard/src/app/[tenantId]/components/LinkWithTenant.tsx:13]: a majority of the `LinkWithTenant` references include `className's` with the same exact styling across all these different files and use cases. I think resolving this issue by implementing `shadcn's` `link` `variant` `Button` is due for a refactor in its own separate issue & PR.

for now, I have implemented a component that fits for a smaller use case within `UserTask.tsx`. 

image

HazimAr (approved):

PR #1193, opened 2024-12-12
fix(lhctl): Add `<revisionNumber>` argument to `wfSpec delete` description

Snarr:
Adds missing argument to the description for the `lhctl wfSpec delete` command.


KarlaCarvajal (approved):
LGTM, it would be nice to have just one parameter for the version, but I think in the future right?

PR #1195, opened 2024-12-13
feat: add extra tags to metronome beats

sauljabin:


sauljabin:
It includes some python lints 😅

mijailr (approved):

Issue #1201, opened 2024-12-16
SearchScheduledWfRunRequest should support querying for pagination


bryson-g:
- `SearchScheduledWfRunRequest` does not contain the necessary properties for querying w/ pagination usages.
  - `SearchScheduledWfRunRequest` should behave as similar as possible in this aspect to `SearchWfRunRequest`.

Screenshot of proto buffs showing the difference between `SearchWfRunRequest`, which has properties supporting querying, vs `SearchScheduledWfRunRequest`:
<img width="1177" alt="image" src="https://github.com/user-attachments/assets/0cf2cadd-5d03-42d0-8b1a-a46cb308ca2d" />

PR #1202, opened 2024-12-17
feat(dashboard): display scheduledWfRuns

### LINKED ISSUE #1087: Show `ScheduledWfRun`s on the Dashboard
We support `rpc SearchScheduledWfRun` in `lhctl`. However, on the dashboard there is no way to view schedules (cronjobs) for a `WfSpec`.

On the `WfSpec` page, we should see scheduled runs. Each `ScheduledWfRun` should show:
* Its schedule
* The input variables for it

bryson-g:
Resolves #1087

## Description
- Tab design between `WfSpecs` & `ScheduledWfSpecs`.
- `SelectionLink` component used for `WfSpec` list & `ScheduledWfSpec` for consistent design.
- Stateful loading pattern for `ScheduledWfSpecs`, since proto buff does not support queries. -https://github.com/littlehorse-enterprises/littlehorse/issues/1201
- Filter `ScheduledWfSpecs` by time window for next expected run datetime.

## Considerations
- We may want another approach for filtering `ScheduledWfRuns`, such as by the cron expression categories themselves.
- More relevant data may want to be displayed for each listed `ScheduledWfRun`, in which case a table UI would be good:
 
https://ui.shadcn.com/docs/components/data-table
image

### Before
https://github.com/user-attachments/assets/0a973e57-70f7-4389-aac7-b8eb9776f3cb

### After
https://github.com/user-attachments/assets/a7a68e60-c333-414f-9dee-99e9c02c14b6


HazimAr:
I think a table would be great addition not only for scheduled Wfs but just normal `WfRun`s

dashboard/src/app/[tenantId]/(diagram)/wfSpec/[...props]/actions/getScheduleWfSpec.ts
```diff
@@ -1,20 +1,30 @@
...
-    majorVersion: parseInt(majorVersion) || 0,
-    revision: parseInt(revision) | 0,
-  })
+
+  return Promise.all(
+    (
+      await client.searchScheduledWfRun({
+        wfSpecName: name,
+        majorVersion: parseInt(majorVersion) || 0,
+        revision: parseInt(revision) | 0,
```
Comment (HazimAr) [inline dashboard/src/app/[tenantId]/(diagram)/wfSpec/[...props]/actions/getScheduleWfSpec.ts:22]:
```suggestion
        revision: parseInt(revision) || 0,
```

dashboard/src/app/[tenantId]/(diagram)/wfSpec/[...props]/components/WfRuns.tsx
```diff
@@ -10,6 +10,8 @@ import { useParams, useSearchParams } from 'next/navigation'
 import { FC, Fragment, useMemo, useState } from 'react'
 import { PaginatedWfRunIdList, searchWfRun } from '../actions/searchWfRun'
 import { WfRunsHeader } from './WfRunsHeader'
+import { SelectionLink } from '@/app/[tenantId]/components/SelectionLink'
+import { parseExpression } from 'cron-parser'
```
Comment (HazimAr) [inline dashboard/src/app/[tenantId]/(diagram)/wfSpec/[...props]/components/WfRuns.tsx:14]: I may be blind but where is this being used in this file.

HazimAr (changes requested):

dashboard/src/app/[tenantId]/(diagram)/wfSpec/[...props]/actions/getScheduleWfSpec.ts
Comment (bryson-g) [reply dashboard/src/app/[tenantId]/(diagram)/wfSpec/[...props]/actions/getScheduleWfSpec.ts:22]: noice

dashboard/src/app/[tenantId]/(diagram)/wfSpec/[...props]/components/WfRuns.tsx
Comment (bryson-g) [reply dashboard/src/app/[tenantId]/(diagram)/wfSpec/[...props]/components/WfRuns.tsx:14]: i suppose the linter does not remove unused imports 🤔

Comment (HazimAr) [reply dashboard/src/app/[tenantId]/(diagram)/wfSpec/[...props]/components/WfRuns.tsx]: I believe there is a VSCode setting that allows you to remove unused code on save. I would look into that. Otherwise lets see if we can add that to the linter. Either way is fine for me. Linter is probably what we should do though.

HazimAr:
Looks like lh-standalone broke, please fix.

HazimAr (approved):

PR #1210, opened 2024-12-17
feat(sdk-go): Handle LHTaskExceptions, Errors, and Panics

### LINKED ISSUE #494: Allow sending EXCEPTION from Go and .NET Task Workers

Snarr:
Adds handling for LHTaskExceptions, Errors, and Panics thrown during the lifecycle of a TaskRun. Serializes all of these types of exceptions and returns them with the ReportTaskRun object.

# Example

The following workflow demonstrates an example usage of the LHTaskException type. Users should define an `error` as the last return value of their Task Method, and then return an instance of the `lhproto.LHTaskException` object.

```go
func Greet(name *string) (string, error) {
	if *name == "" {
		return "", &LHTaskException{
			Name:    "missing-name",
			Message: "Error, name not passed to Greet TaskRun",
			Content: &lhproto.VariableValue{Value: &lhproto.VariableValue_Int{Int: 558}},
		}
	} else if *name == "obi-wan" {
		return "hello there", nil
	} else {
		return "hello, " + *name, nil
	}
}
```

Resolves #494


coltmcnealy-lh (approved):
Looks good, I might also suggest using a different struct rather than the `&lhproto.LHTaskException` directly

PR #1212, opened 2024-12-17
fix(sdk-dotnet): fix server connections management

KarlaCarvajal:
- Add new channel connection only if it do not exist
- Free resources when a thread has finised the process


eduwercamacaro (approved):

KarlaCarvajal:
The code makes the worker stop when the bootstrap server down, so I will close it!

Issue #866, opened 2024-06-06
Show large variable values like YMLs


diablouma:
We can have variables that have a lot of content, in this case an entire YML:

![Image](https://github.com/littlehorse-enterprises/littlehorse/assets/4926247/1456c5dd-63c8-439b-8026-094f7b8ac300)

We need to improve the UX for those scenarios.

HazimAr:
@bryson-g will be assigned to this

PR #1213, opened 2024-12-18
feat(dashboard): large variable display

bryson-g:
resolves [#866](https://github.com/littlehorse-enterprises/littlehorse/issues/866)

## Description
- Created `Overflow` component which can be used anywhere to truncate and display full text on click w/ modal
  - This component auto-formats JSON, and perhaps we will want to support other formats like YAML
- Created `CopyButton` component used for copying text to clipboard
- Created TS logic for dynamic rendering of text on `WfRun's` `variables` section
  - Variables that overflow show new `Overflow` component, while text within bounds appears normal

### Before
image

### After
https://github.com/user-attachments/assets/7061744f-eb74-4c8b-987c-8bbd97805c00


HazimAr:
Nice work on componentizing it only have one question.

How will a user know that it is clickable?

dashboard/src/app/utils/tryFormatAsJson.ts
HazimAr, line 1: Don't we already have a function that does this, I swear remembering seeing this in something, maybe in the variables.

dashboard/src/app/[tenantId]/components/CopyButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const [copied, setCopied] = useState(false)
+
+  return (
+    <Button
+      variant="ghost"
+      size="icon"
+      onClick={() => {
+        navigator.clipboard.writeText(value)
+        setCopied(true)
+        setTimeout(() => setCopied(false), 1000)
```
Comment (HazimAr) [inline dashboard/src/app/[tenantId]/components/CopyButton.tsx:22]: Minor issue with this is that the state of the timeout isn't saved. This is not a big deal but it will technically cause UI issues if the user try's to copy it more than once in 1 second.

I'm sure there is a hook that better handles setTimeout

dashboard/src/app/[tenantId]/components/OverflowText.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const [isOverflowing, setIsOverflowing] = useState(false)
+
+  useEffect(() => {
+    const element = textRef.current
+    if (element) {
+      setIsOverflowing(element.scrollWidth > element.clientWidth)
+    }
+  }, [text])
+
+  const formattedText = tryFormatAsJson(text)
```
Comment (HazimAr) [inline dashboard/src/app/[tenantId]/components/OverflowText.tsx:26]: Why not use the same function that the variables use in the `taskRun` modal?

HazimAr (changes requested):

dashboard/src/app/utils/tryFormatAsJson.ts
bryson-g: its in a PR that isn't merged yet, and probably wont be, considering this PR is the better solution for the same problem

dashboard/src/app/[tenantId]/components/CopyButton.tsx
Comment (bryson-g) [reply dashboard/src/app/[tenantId]/components/CopyButton.tsx:22]: technically perhaps, but honestly i think it's fine:

https://github.com/user-attachments/assets/c0104dc1-c191-4a23-a21d-46cdd869e69c

dashboard/src/app/[tenantId]/components/OverflowText.tsx
Comment (bryson-g) [reply dashboard/src/app/[tenantId]/components/OverflowText.tsx:26]: can u link me the location of the function plz

HazimAr (approved):

PR #1219, opened 2024-12-25
refactor(lhctl): Standardize `lhctl` command formatting and parsing

Snarr:
Standardize the formatting and parsing of `lhctl` commands according to the [docopt](http://docopt.org) format and the [Cobra](https://cobra.dev/#getting-started) docs. I am not totally loyal to either of the above formats, but having a standardized format is better than not having one, so if you disagree with these choices please provide an alternative.

## Changes

1. Move argument validation into the `Args` field of commands
    - Why? To better organize our code and reduce the number of places we do our own bespoke validation
2. Optional arguments surrounded by `[brackets]` in help messages
(e.g. `lhctl search externalEvent [<externalEventDefName>]`
3. Required arguments are usually formatted with `<arrows>` and nothing more, but when two arguments are required together but optional in total, they appear as `[(<arg1> <arg2>)]`. See `docopt: Required elements` for more info.
4.  When necessary, check if a Flag has been `Changed` by the user before operating upon it.
    - Why? Flags are always parsed with a default value, like for strings they will always parse an empty string `""`. This messes up some parts of our code, like in the `put principal` command where the command tried to `parseAcls` on an empty string, even when the user did not provide any `acls`.


lhctl/internal/external_event.go
```diff
@@ -73,13 +80,17 @@ Returns a list of ObjectId's that can be passed into 'lhctl get externalEvent'.
 	returned. The time bound applies to the time that the ExternalEvents
 	were created.
 `,
+	Args: cobra.MaximumNArgs(1),
```
eduwercamacaro, line 83: it seems that we can use`cobra.ExactArgs(1)` instead, right?  then we can remove the validation on line [L#87](https://github.com/littlehorse-enterprises/littlehorse/pull/1219/files#diff-ef8ee86529c6a3aa1c4971184d98231012215689c15324dfbc5d79f3705f9d9fR86)

eduwercamacaro: Oh! nvm.. `externalEventDef` is optional

eduwercamacaro (approved):
This is really helpful! Thanks!

PR #1222, opened 2025-01-04
build(dashboard): next 14 -> 15

### LINKED ISSUE #1223: Next 14 -> Next 15
Upgrade to the new version

hazimoarafa:
closes #1223


HazimAr:
@bryson-g I'll let you take over this PR, the only thing is you need to test with the new next auth version as the old version doesn't support next 15

HazimAr:
@MateoRojas We will need to update the ENVs on the operator and merge those 2 PRs at the same time

dashboard/package.json
```diff
@@ -12,57 +12,61 @@
...
-    "react": "^18.3.1",
-    "react-dom": "^18.3.1",
-    "react-hook-form": "^7.53.2",
-    "reactflow": "^11.11.3",
-    "sonner": "^1.7.0",
-    "tailwind-merge": "^2.3.0",
+    "littlehorse-client": "file:../sdk-js",
+    "lucide-react": "^0.469.0",
+    "next": "15.1.3",
+    "next-auth": "^5.0.0-beta.25",
```
MateoRojas, line 36: This seems to be using a beta release, that's problematic for a couple of reasons the main one being that it means the contract and/or functionality could change in the future. Can we hold off on updating this until a final version is completely available?

mijailr: We can have a golden agreement, like: "only use beta dependencies only if new features are mandatory for business value, otherwise use stable" 

Also bundling two unrelated major upgrades should be avoided.

hazimoarafa:
We can upgrade this at a later date, no need for it now.

PR #1229, opened 2025-01-07
fix(server): Taskqueue ordering

eduwercamacaro:
This PR includes a major refactor for the class `OneTaskQueue` in order to stabilize TaskRun latencies. 

There are several problems related to the task rehydration process:
- TaskQueue ordering is not really guaranteed because the current algorithm aims to execute rehydration only when the task is empty leading to a scenario where the new-arriving tasks gets executed *before* the ones that are serialized on disk.
- Task rehydration happens inside a synchronized block (lock/unlock) which means that the streams threads will be blocked until the rehydration process releases the lock. This is really bad for application instance stability.
- It is really tricky to implement an algorithm that guarantees that all serialized tasks will be rehydrated, that caused many bugs in the past. 

Changes included on this PR are really simple:
- Uses a `java.util.Set` implementation to represent a queue that keeps scheduledTasks in memory until `PollTaskRequest.onNext` takes place. 
- Includes a virtual limit that changes the behavior of the task queue when it is reached: Queue items only store a reference to the taskRunId, and PollTaskRequest.onNext will resolve the ScheduledTask by executing a get to the stateStore.  Note that It adds an extra latency for the pollTasks when this limit is reached.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streams/taskqueue/OneTaskQueue.java
+++ b/server/src/main/java/io/littlehorse/server/streams/taskqueue/OneTaskQueue.java
@@ -4,7 +4,11 @@ import io.littlehorse.common.model.ScheduledTaskModel;
 import io.littlehorse.common.model.getable.objectId.TenantIdModel;
 import io.littlehorse.sdk.common.LHLibUtil;
 import io.littlehorse.server.streams.topology.core.RequestExecutionContext;
-import java.util.*;
+
+import java.util.LinkedHashSet;
+import java.util.LinkedList;
+import java.util.Objects;
+import java.util.Queue;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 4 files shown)


sauljabin (approved):
I look at the canary metrics too.
Good code.

server/src/main/java/io/littlehorse/server/streams/taskqueue/OneTaskQueue.java
```diff
@@ -160,127 +126,85 @@ public boolean onTaskScheduled(TaskId streamsTaskId, ScheduledTaskModel schedule
...
-                        new TrackedPartition(
-                                trackedPartition.hasMoreDataOnDisk(), trackedPartition.lastRehydratedTask(), nextTask));
-            } else {
-                // case 2
                 hungryClients.add(requestObserver);
+                return null;
             }
-        } finally {
-            lock.unlock();
+            return pendingTasks.removeFirst();
```
KarlaCarvajal, line 134: why do you need to remove the first pending task if there were no pending tasks?

Issue #473, opened 2023-12-19
Dangling Exit Node on Python + Go SDK


diablouma:
[https://github.com/littlehorse-enterprises/littlehorse/pull/142|https://github.com/littlehorse-enterprises/littlehorse/pull/142|smart-link] 

That fixed it in Java.

coltmcnealy-lh:
This needs to be looked at. Could be a bug, but it's unclear whether it is still a problem in Java and Go, especially after @MateoRojas made significant refactors to how we do conditional branching prior to `0.7.0`

PR #142, opened 2023-11-01
MINOR: Remove Dangling Exit Node

coltmcnealy-lh:
Fixes issue in Java SDK where we add an unneeded exit node if the user calls `WorkflowThread#fail()` as the last invocation in the workflow thread building function. No impact on actual workflow execution; just makes the dashboard look confusing.

PR #1231, opened 2025-01-08
fix(sdk-go, sdk-python): Remove dangling exit nodes on early returned WorkflowThreads

Snarr:
This PR removes the dangling exit node left on WorkflowThreads that return early using the `WorkflowThread.fail()` method in `sdk-python` and `sdk-go`.

**Resolves Issue #473.**

**Applies the changes in #142 to `sdk-python` and `sdk-go`.**

## Before
A WorkflowThread that reads as follows:
```
wf.execute("some-task")
wf.fail("my-failure", "failure_message)
```
Would effectively stop at the Node created by `wf.fail()`. But the WorkflowThread builder in the SDK would add an additional ExitNode anyways, resulting in this dangling node:

image

## After
The Python and Go SDKs will now check if the last node in a WorkflowThread is an exit node before adding another one.
image

### DIFF
--- a/sdk-go/littlehorse/wf_lib_internal_test.go
+++ b/sdk-go/littlehorse/wf_lib_internal_test.go
@@ -9,7 +9,31 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestCanMakeSearcjableVariable(t *testing.T) {
+func TestEarlyReturnOnExitNode(t *testing.T) {
+	wf := littlehorse.NewWorkflow(func(t *littlehorse.WorkflowThread) {
+		t.Execute("some-task")
+		t.Fail("failure message", "my-failure", nil)
+	}, "my-workflow")
+
+	putWf, err := wf.Compile()
+	if err != nil {
+		t.Error(err)
+	}
+
+	entrypoint := putWf.ThreadSpecs[putWf.EntrypointThreadName]
+
+	exitNodeCount := 0
+
+	for _, value := range entrypoint.Nodes {
+		if value.GetExit() != nil {
+			exitNodeCount += 1
+		}
+	}
+
+	assert.Equal(t, exitNodeCount, 1)
+}
+
+func TestCanMakeSearchableVariable(t *testing.T) {
 	wf := littlehorse.NewWorkflow(func(t *littlehorse.WorkflowThread) {
 		t.AddVariable(
 			"my-var", lhproto.VariableType_BOOL,
--- a/sdk-go/littlehorse/wf_lib_internal.go
+++ b/sdk-go/littlehorse/wf_lib_internal.go
@@ -72,9 +72,11 @@ func (l *LHWorkflow) compile() (*lhproto.PutWfSpecRequest, error) {
 				function(&thr)
 
 				// Now add exit node to make a sandwich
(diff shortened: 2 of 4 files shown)


sdk-go/littlehorse/wf_lib_internal_test.go
```diff
@@ -9,7 +9,31 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestCanMakeSearcjableVariable(t *testing.T) {
+func TestEarlyReturnOnExitNode(t *testing.T) {
```
KarlaCarvajal, line 12: is this a unit test?, is not this code executed in the production code? 

```exitNodeCount := 0

	for _, value := range entrypoint.Nodes {
		if value.GetExit() != nil {
			exitNodeCount += 1
		}
	}```

So if the wfSpec has 2 nodes and the fail one is always executed the exitNodeCount should allways be 1 right?

Comment (KarlaCarvajal) [reply sdk-go/littlehorse/wf_lib_internal_test.go:12]: is it possible to test the last node type?

File: sdk-go/littlehorse/wf_lib_internal_test.go [hunk 2]
```diff
@@ -9,7 +9,31 @@ import (
...
+	wf := littlehorse.NewWorkflow(func(t *littlehorse.WorkflowThread) {
+		t.Execute("some-task")
+		t.Fail("failure message", "my-failure", nil)
+	}, "my-workflow")
+
+	putWf, err := wf.Compile()
```
KarlaCarvajal, line 18: I do not believe the workflow may fail when it is compiled, it should capture the failure in the Fail node, so it is not needed to test that an Error is thrown ```if err != nil {
		t.Error(err)
	}```

sdk-go/littlehorse/wf_lib_internal_test.go
Snarr: Yes, this is just for unit testing the SDK

In any given ThreadSpec, there should only be 1 entrypoint node and 1 exit node, so this test checks the `entrypoint` thread to make sure there is only 1 exit node.

In Go, the entrypoint.Nodes is a map of strings, so I don't think I can check the last node type.

sdk-go/littlehorse/wf_lib_internal_test.go
Snarr: You're right, there is no failure returned in our Go SDK for Workflow compilation. I'm not sure if we plan to add one, but right now the `.compile` method always returns `nil`.

Snarr: I just wrote this to keep the test consistent with the other tests in this library, but we should look into it further

coltmcnealy-lh: Our compile method right now does a `panic()` if something goes wrong. That's really bad in terms of GoLang hygiene...we shouldn't have done it that way. 

However, given the nature of compiling a `WfSpec` (normally happens in something like a pipeline), it isn't as bad as if the task worker did a `panic()`.

Not worth fixing right now...maybe for LH 2.0

coltmcnealy-lh (approved):

PR #1232, opened 2025-01-09
refactor(sdk-dotnet): rename methods and class from helpers

KarlaCarvajal:


sdk-dotnet/LittleHorse.Sdk.Tests/Helper/LhMappingTest.cs
```diff
@@ -7,7 +7,7 @@
 using Xunit;
 using Type = System.Type;
 
-public class LHMappingHelperTest
+public class LhMappingTest
```
Snarr, line 10: Can we make LH uppercase to match the rest of the code, or is this for some standard where we need to keep it lowercase?

Snarr (approved):

sdk-dotnet/LittleHorse.Sdk.Tests/Helper/LhMappingTest.cs
coltmcnealy-lh: I would second this.

coltmcnealy-lh (approved):
Looks good, I would imagine this requires a change to the `lh-site` docs repo though.

sdk-dotnet/LittleHorse.Sdk.Tests/Helper/LhMappingTest.cs
KarlaCarvajal: Thanks guys for the feedback, yesterday I that test class kept with an old name

PR #1233, opened 2025-01-09
chore(server): drop core repartition processor and stores

eduwercamacaro:
Removes unused repartition store and processor.


server/src/main/java/io/littlehorse/server/streams/ServerTopology.java
```diff
@@ -148,20 +145,6 @@ public static Topology initCoreTopology(
                 Stores.persistentKeyValueStore(CORE_STORE), Serdes.String(), Serdes.Bytes());
         topo.addStateStore(coreStoreBuilder, CORE_PROCESSOR);
 
-        // Repartition sub-topology
```
coltmcnealy-lh, line 151: Can we comment instead of delete?

coltmcnealy-lh (approved):

PR #1238, opened 2025-01-10
feat(dashboard): display taskdef usage by wfspecs

bryson-g:
## Description
- Allow the user to view a list of `WfSpecs` that are used by any particular `TaskDef`.
  - Via `WfSpec Usage` section under the `TaskDef` page. 
- Using same `WfSpec` display visuals as on the primary `WfSpecs` view to maintain consistency and usage expectations by the user

### Screenshot
image

### Scroll Multiple WfSpecs Handling
https://github.com/user-attachments/assets/ef3c684d-4bcf-4604-941b-843e624d6e26

### DIFF
--- a/dashboard/src/app/[tenantId]/page.tsx
+++ b/dashboard/src/app/[tenantId]/page.tsx
@@ -1,5 +1,7 @@
+import { lhClient } from '../lhClient'
 import { Search } from './components/Search'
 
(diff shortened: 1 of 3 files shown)


HazimAr (approved):
I would like to see another UI/UX of this I think there is a better way to display this info but for now it's good

PR #1242, opened 2025-01-10
fix(sdk-python): update poetry, authlib and grpcio dependencies

sauljabin:
I've tested the worker with and without oauth


coltmcnealy-lh (approved):
Thank you

PR #1259, opened 2025-01-21
perf(server): avoid rocksdb range scan on deletewfrun

coltmcnealy-lh:
- Intelligently remember which NodeRun's and TaskRun's and variables there were
- No range scans over prefix
- Journal externalEventId's since we can't guess them


server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/SubNodeRun.java
```diff
@@ -99,4 +100,12 @@ public WfRunModel getWfRun() {
     public NodeModel getNode() {
         return nodeRun.getNode();
     }
+
+    /**
+     * Returns the created sub-Getable for this NodeRunModel. For example, a TaskNodeRunModel would
+     * return the associated TaskRunModel.
+     */
+    public Optional<? extends CoreObjectId<?, ?, ?>> getCreatedSubGetableId() {
+        return Optional.empty();
```
Snarr, line 109: Why does this return an empty optional ?

coltmcnealy-lh: The majority of the subnoderuns don't actually create another getable. Only `TaskNodeRun`, `WorkflowEventNodeRun`, and `UserTaskNodeRun` create other getables.

eduwercamacaro (approved):

PR #1262, opened 2025-01-21
feat(sdk-dotnet): add execute task feature to workflow spec

KarlaCarvajal:
- Add implementation of NodeOutput
- Add VariableAssigment helper
- Add execute tasks implementation in workflow thread
- Add unit tests


sdk-dotnet/LittleHorse.Sdk/Workflow/Spec/WorkflowThread.cs
```diff
@@ -104,15 +109,54 @@ private string AddNode(string name, Node.NodeOneofCase type, IMessage subNode) {
...
     }
+    
+    public NodeOutput Execute(String taskName, params object[] args) {
+        CheckIfWorkflowThreadIsActive();
+        _parent.AddTaskDefName(taskName);
+        var taskNode = CreateTaskNode(
+            new TaskNode {TaskDefId = new TaskDefId {Name = taskName}}, args);
+        string nodeName = AddNode(taskName, Node.NodeOneofCase.Task, taskNode);
+        
+        return new NodeOutput(nodeName, this);
```
Snarr, line 133: Eventually this should return a TaskNodeOutput which has additional methods (configurable retries & exponential backoff) on top of the normal NodeOutput implementation. But I understand this is early in the SDK and you probably chose to keep it simple for now 😁

Snarr (approved):

sdk-dotnet/LittleHorse.Sdk/Workflow/Spec/WorkflowThread.cs
KarlaCarvajal: exactly, until I reach that point of usage I am not adding more code

sdk-dotnet/LittleHorse.Sdk/Helper/LHVariableAssigmentHelper.cs
```diff
@@ -0,0 +1,48 @@
+using LittleHorse.Sdk.Common.Proto;
+using LittleHorse.Sdk.Workflow.Spec;
+
+namespace LittleHorse.Sdk.Helper;
+
+internal static class LHVariableAssigmentHelper
+{
+    internal static VariableAssignment AssignVariable(object? variable)
```
eduwercamacaro, line 8:
```suggestion
    internal static VariableAssignment AssignVariable(object? variableValue)
```

KarlaCarvajal: it should be variable, as in some cases just the variable name is set, and in others the value

sdk-dotnet/LittleHorse.Sdk.Tests/Helper/LHVariableAssigmentHelperTest.cs
```diff
@@ -0,0 +1,85 @@
...
+       var variableAssigment = LHVariableAssigmentHelper.AssignVariable(nodeOutput);
+       
+       Assert.Equal(nodeOutput.NodeName, variableAssigment.NodeOutput.NodeName);
+       Assert.Equal(nodeOutput.JsonPath, variableAssigment.JsonPath);
+    }
+
+    [Fact]
+    public void VariableAssigment_WithNotDefinedObject_ShouldAssignObjectAsDefaultVariable()
+    {
+        var notDefinedObjects = new List<object>
```
eduwercamacaro, line 73: Please, use parametrized test instead of defining multiple asserts

sdk-dotnet/LittleHorse.Sdk/Helper/LHVariableAssigmentHelper.cs
eduwercamacaro: It is always a variable value, but sometimes the _value_ is just a reference to another variable.

sdk-dotnet/LittleHorse.Sdk/Utils/JsonHandler.cs
```diff
@@ -16,6 +17,13 @@ internal static string ObjectSerializeToJson(object o)
     
     internal static object? DeserializeFromJson(string json, Type type)
     {
-        return JsonConvert.DeserializeObject(json, type);
+        try
```
eduwercamacaro, line 20: uh oh.. this class seems redundant. I don't think we need to create an extra abstraction for Json serializers. IMO, it makes the client code more difficult to understand. wdyt? @Snarr

KarlaCarvajal: Hey Edu, I see your point, but the purpose of this extraction is to make be testable, as Jsons are used in many cases along the SDK and for the rest of the code we do not need to worry about it.

PR #1263, opened 2025-01-21
fix(sdk-go): Use `int32` on `WithRetries`, Test `WithExponentialBackoff`

Snarr:


KarlaCarvajal (approved):
LGTM, we need to take into account this on the others SDKs

PR #1265, opened 2025-01-23
fix(sdk-java): adds missing configs to sdk java

sauljabin:
- Uses an enum to define all the settings and then it exports them as a set with configNames()

### DIFF
--- a/sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
+++ b/sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
@@ -89,7 +89,8 @@ public class LHConfig extends ConfigBase {
             LHConfig.OAUTH_CLIENT_SECRET_KEY,
             LHConfig.NUM_WORKER_THREADS_KEY,
             LHConfig.TASK_WORKER_VERSION_KEY,
-            LHConfig.INFLIGHT_TASKS_KEY);
+            LHConfig.INFLIGHT_TASKS_KEY,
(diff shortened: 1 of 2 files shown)


sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
```diff
@@ -22,82 +22,96 @@
...
+        LHC_CA_CERT,
+        LHC_OAUTH_CLIENT_ID,
+        LHC_OAUTH_CLIENT_SECRET,
+        LHC_OAUTH_ACCESS_TOKEN_URL,
+        LHW_TASK_WORKER_ID,
+        LHW_TASK_WORKER_VERSION,
+        LHW_NUM_WORKER_THREADS,
+        LHC_GRPC_KEEPALIVE_TIME_MS,
+        LHC_GRPC_KEEPALIVE_TIMEOUT_MS,
+        LHC_INFLIGHT_TASKS
```
KarlaCarvajal, line 55: should we add those config keys in the other SDKs?

KarlaCarvajal (approved):

Snarr (approved):

MateoRojas (approved):

sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
sauljabin: It is a good idea, so maybe, we can discuss it

PR #1266, opened 2025-01-24
refactor(canary): add kafka config filters

sauljabin:


sauljabin:
Closed because it's not needed yet.
With @eduwercamacaro

PR #1269, opened 2025-01-27
fix(sdk-python): Enable adding defaults to `declare` var methods

Snarr:


coltmcnealy-lh (approved):
Thanks, but a unit test would be nice (:

PR #1272, opened 2025-01-30
feat(sdk-dotnet) add implementation for do-if conditional

KarlaCarvajal:
- Add implementation of NOP nodes with conditionals 
- Add example
- Refactor failing tests


Snarr (approved):
Code works, WfSpec produced matches Java 👍

Issue #891, opened 2024-06-18
Java SDK Serialization Erorr Yields Timeout


coltmcnealy-lh:
If I have a `LHTaskMethod` as follows:

```
@LHTaskMethod("some-task")
public UUID returnSomeUUID() {
    return UUID.randomUUID();
}
```

I will observe the `TaskRun` go into the `TIMEOUT` state. That is because we fail to serialize the `UUID` properly. We already have a code in our `TaskStatus` proto that is for output serialization errors. When we fail to serialize the output, we should rethrn an `LHTaskError` to the server, rather than fail silently without a stacktrace, which is what we do right now. 

- Output serializing error: https://littlehorse.dev/docs/api/#taskstatus
- LHTaskError: https://littlehorse.dev/docs/api/#lhtaskerror

PR #1275, opened 2025-02-04
fix(sdk-java): Fix JsonToStr implementation that stores JsonStrings as JsonObjects

Snarr:
## Background

This PR set out to solve issue #891, which pointed out that returning a `UUID` from a Java Task Worker would cause a Task Timeout on the Server.

## Cause of the bug

The Task Timeout was caused by the Task Worker SDK sorting a `UUID` Java Object into a `VariableValue.JsonObj` in our protobuf while serializing the object into a JSON string.

This is a problem because the server expected a `JsonObj` to serialize into a `Map<String, Object>`, but received a String instead.

### Example:
```java
@LHTaskMethod("get-uuid")
public UUID getUUID() {
    return UUID.randomUUID();
}
```

The above Task Method would return a `UUID` object, which would eventually get sorted into the `VariableValue.JsonObj` field. But when our JSON Serde Library "Jackson" tried to convert the `UUID` object to a string, it would output a JSON string, such as `"77b0f54a-c102-40e0-a50b-6b93a"`.

## Solution

Initially, my solution was to make sure every object serializes into a JSON Object. For example, `UUID` would serialize into `{"UUID": "77b0f54a-c102-40e0-a50b-6b93a"}`. But neither Jackson nor GSON could accomplish this 100% of the time. 

Instead, I replaced the Jackson library with `GSON`. GSON provides methods to determine whether or not any object would serialize into a JSON String, Array, or Object, which allows us to then sort it into the appropriate `VariableValue` type. 

### Conversions

Below is a table with examples on how a Java Object is serialized and then turned into a protobuf VariableValue type

| Java Type  | GSON Type  | VariableValue Type |
|------------|------------|---------------------|
|   POJO     | JsonObject |        JsonObj      |
| Collection | JsonArray  |        JsonArr      |
|   UUID     | JsonString |        String       |

Resolves: #891


sdk-java/src/test/java/io/littlehorse/sdk/common/LHVariableMapperTest.java
```diff
@@ -95,7 +96,7 @@ void shouldVerifyVariableTypeBeforeConvertingTheValue() {
         assertThatThrownBy(() -> asList(boolVal, Person.class))
                 .isInstanceOf(IllegalArgumentException.class)
                 .hasMessage("Not possible to convert variable type BOOL to Collection");
-        assertThatThrownBy(() -> as(boolVal, Person.class)).isInstanceOf(RuntimeException.class);
+        assertThat(as(boolVal, Person.class)).isNull();
```
Snarr, line 99: No longer throws a RuntimeException, just returns null

server/src/main/java/io/littlehorse/server/monitoring/StandbyStoresOnInstance.java
```diff
@@ -58,7 +56,6 @@ public long totalLag() {
      *
      * @return The number of registered partitions for this store.
      */
-    @JsonProperty("numberOfRegisteredPartitions")
```
Snarr, line 61: Note to reviewer: These properties are not used anywhere. If you check the metrics code that depends on this class, it does not use these Jackson Annotations in any way.

server/src/test/java/io/littlehorse/storeinternals/JsonVariableStorageManagerTest.java
```diff
@@ -59,16 +62,18 @@ public class JsonVariableStorageManagerTest {
...
+    private static Gson gson = new GsonBuilder()
+            .setObjectToNumberStrategy(ToNumberPolicy.LONG_OR_DOUBLE)
+            .create();
 
     @BeforeEach
     public void setup() throws Exception {
         initializeDependencies();
 
         // JSON_OBJ test
+        String content = new String(Files.readAllBytes(Paths.get("./src/test/resources/json-variables-example.json")));
```
Snarr, line 74: Note to self: Revisit

eduwercamacaro (approved):

sdk-java/src/main/java/io/littlehorse/sdk/common/LHLibUtil.java
```diff
@@ -154,18 +169,25 @@ public static VariableValue objToVarVal(Object o) throws LHSerdeError {
...
+                case PRIMITIVE:
+                    // Trims the quotes off the string
+                    out.setStr(jsonResult
+                            .getJsonStr()
+                            .substring(1, jsonResult.getJsonStr().length() - 1));
+                    break;
+                case NULL:
+                    break;
+                default:
+                    throw new LHSerdeError("Failed deserializing object to Json: " + o.toString());
```
KarlaCarvajal, line 190: I believe this message should say `Failed serializing object to Json`, a deserialization is the other way around parsing json into objects

Snarr: I agree, great catch Karla

PR #1286, opened 2025-02-12
fix(server): Report TaskRun failure when var deserialization error occurs

### LINKED ISSUE #1279: Server ReportTaskRun serialization failure perceived as TaskTimeout
When the LittleHorse Server fails to serialize a `ReportTaskRun` request, it returns an `LHApiException` before forwarding a Record through our Kafka Streams Topology reporting the state of the `TaskRun`. This means that  the `TaskRun` state is usually perceived as a `Timeout` by the server, since no state is reported. We should forward a `ReportTaskRun Deserialization Error` or something equivalent as the state of the `TaskRun` to our topology.

## Steps to reproduce

```
1. Register any `TaskDef` with the LittleHorse Server
2. Register any `WfSpec` that uses that `TaskDef` with the LittleHorse Server
3. Run a Task Worker for your `TaskDef`
4. Run the workflow and execute that Task
5. Intercept the `ReportTask` RPC call before it is sent to the server
6. Use an invalid `json_obj` value for the `ReportTask.Output`, such as a non-JSON formatted string `"test"`
8. The server will fail to serialize the `ReportTask` request and interpret the Task as timed out after the timeout period expires
```

Snarr:
When the server fails to deserialize a ReportTaskRun message, it should forward a failure through our Kafka Streams topology so that a workflow can recover from the serde failure and a user can understand what went wrong.

This PR also adds a Deserialization Error field to VariableValueModels that enables catching deserialization errors whenever invalid JSON_OBJ or JSON_ARR comes in through a VariableValue. This helps with ReportTaskRun, as we can find out exactly what part of the Task Run Report was invalid.

Resolves #1279


server/src/main/java/io/littlehorse/common/model/getable/core/variable/VariableValueModel.java
```diff
@@ -41,6 +43,10 @@ public class VariableValueModel extends LHSerializable<VariableValue> {
 
     private ExecutionContext context;
 
+    private boolean deserializationErrorPresent = false;
```
coltmcnealy-lh, line 46: Rather than having two fields with `Getter`s, I would suggest that we have a method:

```
public Optional<String> getDeserializationError() {
   ...
};
```

And also make the `deserializationErrorPresent` field nave `@Getter(AccessLevel.NONE)` on it.

coltmcnealy-lh (approved):
Approved, BUT:

- A minor refactor request.
- This is lacking an end-to-end test.

Great work! Also, thanks for bringing back old memories. Some of this code is so old...I remember the desk that I wrote it at. I miss those monitors that I had in Austin. But life is better now that I have a wonderful team working with me.

eduwercamacaro (approved):

PR #1290, opened 2025-02-14
fix(sdk-go): trim space on boolean on StrToVarVal

### LINKED ISSUE #1035: LHCTL for Windows:  Unable to give user response for boolean variable type
### Context

Note: This happened only on Windows OS.

so when ever there is a Boolean variable response require (true/false) . windows by default add /r at the end, which is not recognized by lhctl and give a syntax error.

might we need a check that bypass for windows user

image

### Steps To Reproduce

1. use windows OS
2. run a wfspec with usertask
3. execute a usertask with boolean type
4. give true or false as input
5. system will response syntax error

### Expected Behavior

lhctl bypass /r and complete the user task successfully without any syntax error

### Screenshots

_No response_

### Additional Context

_No response_

### Components

_No response_

bryson-g:
resolves #1035

### DIFF
--- a/sdk-go/littlehorse/lh_variables.go
+++ b/sdk-go/littlehorse/lh_variables.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"reflect"
 	"strconv"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


Snarr (approved):
Looks good 👍

PR #1301, opened 2025-02-18
chore(dashboard): pins nginx-nodejs image to version v1

MateoRojas:
Pins `ghcr.io/littlehorse-enterprises/alpine-nginx-nodejs/nginx-nodejs` to version `v1`. It's a good practice to pin the version of the artifacts to avoid unintentionally using a recently released version that might contain errors. It helps to keep builds consistent.


Snarr (approved):
Thank you for this 😄

mijailr (approved):
Thanks Mateo

PR #1316, opened 2025-02-19
fix(dashboard): wfrun blank page

bryson-g:
### Description
- Refactored the id pulled from the url, which is passed to the wfRun page component, to be formatted via WfRunId object instead of just the string
- Refactored SWR implementation to avoid running the getWfRun request if there is no valid ID present

### Video
https://github.com/user-attachments/assets/dd4c56b1-1451-49bc-8988-80197f55e8de

### DIFF
--- a/dashboard/src/app/hooks/useWfRun.tsx
+++ b/dashboard/src/app/hooks/useWfRun.tsx
@@ -1,19 +1,25 @@
 "use client";
 import useSWR from "swr";
-import { getWfRun } from "../actions/getWfRun";
+import { getWfRun, WfRunResponse } from "../actions/getWfRun";
+import { useState } from "react";
+import { WfRunId } from "littlehorse-client/proto";
 
 type Props = {
-    id: string
+    wfRunId: WfRunId
     tenantId: string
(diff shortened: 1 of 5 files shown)


dashboard/src/app/hooks/useWfRun.tsx
```diff
@@ -1,19 +1,25 @@
...
     tenantId: string
 }
 
-export function useWfRun({ id, tenantId }: Props) {
-    const { data, error, isLoading } = useSWR(`wfRun/${tenantId}/${id}`, () => {
-        return getWfRun({ ids: [id], tenantId })
+export function useWfRun({ wfRunId, tenantId }: Props) {
+    const [dataCache, setDataCache] = useState<WfRunResponse | null>(null)
+
+    const { error, isLoading } = useSWR(`wfRun/${tenantId}/${wfRunId.id}/${wfRunId.parentWfRunId?.id}`, async () => {
```
hazimoarafa, line 15: this doesn't matter but don you want to flip parentwf and normal wf in this string

PR #1317, opened 2025-02-19
fix(dashboard): nested parents

bryson-g:


dashboard/src/app/(authenticated)/[tenantId]/(diagram)/wfRun/[...ids]/page.tsx
```diff
@@ -8,19 +8,7 @@ import { WfRunId } from 'littlehorse-client/proto'
...
-      parentWfRunId: {
-        id: ids[0],
-      }
-    }
-  } else {
-    wfRunId = {
-      id: ids[0],
-    }
-  }
+  const wfRunId: WfRunId = ids.reduce((wfRunId, id, i) => (i === 0 ? { id } : { id, parentWfRunId: wfRunId }), {} as WfRunId);
```
Comment (mijailr) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/wfRun/[...ids]/page.tsx:11]: The list must be reversed, please use the previous implementation.

```suggestion
  const wfRunId = ids
    .reverse()
    .reduceRight<WfRunId | undefined>((parentWfRunId, id) => ({ id, parentWfRunId }), undefined)
```

Also, this should be done in the calling function. you can pass the string array from this page, so the receiver has the logic to generate the valid WfRunId.

mijailr (changes requested):

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/wfRun/[...ids]/page.tsx
Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/wfRun/[...ids]/page.tsx:11]: The previous implementation negates the `reverse` by immediately after doing `reduceRight`.
`reverse().reduceRight()` is the same as `reduce`.
https://stackoverflow.com/a/20753610}

Additionally, the old implementation doesn't work in this scenario because the initial value for the reduce function becomes improperly inserted in the first index. This is why the new implementation uses the `i == 0` conditional

Since this is a server component, I do agree that it should be ran in the client component.

Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/wfRun/[...ids]/page.tsx]: > > Also, this should be done in the calling function. you can pass the string array from this page, so the receiver has the logic to generate the valid WfRunId.
> 
> This is not mandatory, having it here is fine as well. It's just a preference. If we use the WfRunId in the component for multiple purposes, then we should have a utility function.

I think it's better practice to include it in the client component as you suggest, so we avoid potentially slowing down the server with any unnecessary operations (even though this operation is so simple that would never happen here, it's just good practice I suppose 😂 ).

mijailr (approved):

PR #1321, opened 2025-02-20
fix(server): Reject invalid `wf_run_id` in `RunWfRequest`

### LINKED ISSUE #1315: Reject `RunWf` requests with underscores in the `WfRunId`
The server uses underscores (`_`) as a delimiter for storing Parent and Child WfRunIds, like so: `parent-id_child-id`. This means that underscores should be treated as a special character and users should not be able to use them in their custom `WfRunIds` or we risk confusing the server.

## Bug

If a user does execute a single `WfRun` with an underscore in the `WfRunId`, the server interprets the `WfRun` as being a child workflow run tied back to a parent workflow run.

### Steps to Reproduce

1. Register a WfSpec with the server
2. Make a `RunWf` request for that WfSpec, with the `WfRunId` set to `yoda_grogu`
3. Search for the existing WfRuns under that WfSpec
4. The search should reveal that a workflow `grogu` was created, and that the workflow has a parent workflow of `yoda`

This is a problem for multiple reasons. Firstly, Yoda is not Grogu's parent. Secondly, the server now thinks that there should be a WfRun named `yoda` somewhere, even though it does not exist.

Snarr:
Checks `RunWfRequest` `wf_run_id` against the `LHUtil.isValidLHName()` method in both the API layer and the process layer.

Resolves #1315


server/src/main/java/io/littlehorse/server/LHServerListener.java
```diff
@@ -591,6 +591,14 @@ public void searchScheduledWfRun(SearchScheduledWfRunRequest req, StreamObserver
     @Override
     @Authorize(resources = ACLResource.ACL_WORKFLOW, actions = ACLAction.RUN)
     public void runWf(RunWfRequest req, StreamObserver<WfRun> ctx) {
+        if (req.getWfSpecName().equals("")) {
+            throw new LHApiException(Status.INVALID_ARGUMENT, "Missing required parameter 'wf_spec_name'");
+        }
+
+        if (!req.getId().equals("") && !LHUtil.isValidLHName(req.getId())) {
```
coltmcnealy-lh, line 598: I wouldn't check for `req.getId().equals("")`. Instead, use the protobuf _optional_ capability which checks for presence of the string. The way to do that is: `req.hasId()`.

I would also suggest adding this:
```
if (req.hasId() && req.getId().equals("")) throw new LHApiException(...);
```

coltmcnealy-lh (approved):
Pre-emptively approving with a comment to address

server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/RunWfRequestModel.java
```diff
@@ -93,6 +93,16 @@ public boolean hasResponse() {
 
     @Override
     public WfRun process(ProcessorExecutionContext processorContext, LHServerConfig config) {
+        if (wfSpecName.equals("")) {
+            throw new LHApiException(Status.INVALID_ARGUMENT, "Missing required argument 'wf_spec_name'");
+        }
+
+        if (id != null) {
```
sauljabin, line 100: if id is an attribute you can make a method this.isIdValid()

+ yo can use Strings.isNullOrEmpty

server/src/test/java/e2e/RunWfTest.java
```diff
@@ -0,0 +1,29 @@
...
+        try {
+            client.runWf(RunWfRequest.newBuilder()
+                    .setId("my_workflow")
+                    .setWfSpecName("basic-example")
+                    .build());
+        } catch (StatusRuntimeException exn) {
+            caught = exn;
+        }
+        assertNotNull(caught);
+        Assertions.assertThat(caught.getStatus().getCode()).isEqualTo(Code.INVALID_ARGUMENT);
```
sauljabin, line 27: try to use assertThrows

server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/RunWfRequestModel.java
Snarr: Thanks for the suggestion, I implemented the `isIdValid` method.

I am not sure if `Strings.isNullOrEmpty` would apply here, since the `id` _can_ be null but _shouldn't_ be an empty string. Null implies that the `id` wasn't set, but empty string implies that someone tried to set the `id` with no value.

Issue #1296, opened 2025-02-17
Use Code Font on Dashboard


coltmcnealy-lh:
Show values of variables and the returns of Task Functions (basically, everything that's a `VariableValue`) in `code font` so that it's easier to see. Especially important for JSON strings.

PR #1326, opened 2025-02-22
refactor(dashboard): use code font

bryson-g:
resolves [#1296](https://github.com/littlehorse-enterprises/littlehorse/issues/1296)
- Simple quick fix to use `Fira Code` font where variable values and outputs are used.
  - The `OverflowText` component is used anywhere that dynamic values are populated (i.e. variable values, outputs, etc), so the default font for this component has been made to use `Fira Code`.
- Implemented font via `tailwindcss.config.ts`.


dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx
```diff
@@ -37,7 +37,7 @@ export const OverflowText: FC<OverflowTextProps> = ({ text, className, variant }
               className
             )}
           >
-            <span className="truncate">{formattedText}</span>
+            <span className="truncate font-code">{formattedText}</span>
```
Comment (hazimoarafa) [inline dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx:40]: Why not use pre instead of span

dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx
```diff
@@ -51,14 +51,14 @@ export const OverflowText: FC<OverflowTextProps> = ({ text, className, variant }
               'bg-status-failed text-red-500': variant === 'error',
             })}
           >
-            <div className="max-w-full whitespace-pre-wrap break-words p-4">{formattedText}</div>
+            <div className="max-w-full whitespace-pre-wrap break-words p-4 font-code">{formattedText}</div>
```
Comment (hazimoarafa) [inline dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx:54]: Same here

dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx
```diff
@@ -51,14 +51,14 @@ export const OverflowText: FC<OverflowTextProps> = ({ text, className, variant }
...
           </div>
         </DialogContent>
       </Dialog>
     )
   }
   return (
-    <div ref={textRef} className={cn(className, "flex items-center justify-center h-full")}>
+    <div ref={textRef} className={cn(className, "flex items-center justify-center h-full font-code")}>
```
Comment (hazimoarafa) [inline dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx:61]: Same here

dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx
Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx:40]: not sure. i just went with `span` here. what are the benefits for `pre` in this case?

Comment (hazimoarafa) [reply dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx:40]: Just convention html standards

Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx:40]: idk if `pre` makes sense here because it `pre`serves the whitespace of the characters written within the element, but in this case we're using a variable value.

image

dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx
Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx:61]: div is a block element, vs pre/span being inline, which might change the styling on the page. i'd just leave it as div here.

dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx
Comment (hazimoarafa) [reply dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx:40]: Not with tailwind preflight

Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx:40]: https://tailwindcss.com/docs/preflight

i dont see any modifications to the `pre` tag tho

Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/components/OverflowText.tsx:40]: > https://tailwindcss.com/docs/preflight
> 
> i dont see any modifications to the `pre` tag tho

wait nvm maybe i should actually read the page where it says 'view full list of changes'
image

hazimoarafa (approved):

Issue #1294, opened 2025-02-17
dashboard: remove sign out if no auth


bryson-g:
<img width="1051" alt="Image" src="https://github.com/user-attachments/assets/69878ed6-10ab-41b2-a411-66600c02d8ea" />

PR #1327, opened 2025-02-22
fix(dashboard): conditionally render sign out

bryson-g:
resolves [#1294](https://github.com/littlehorse-enterprises/littlehorse/issues/1294)
- Conditionally renders sign out button depending on env var `LHD_OAUTH_ENABLED`
  - Do we want to use data from next-auth to determine if auth is enabled, or the env?

### DIFF
--- a/dashboard/src/app/(authenticated)/[tenantId]/components/Principal.tsx
+++ b/dashboard/src/app/(authenticated)/[tenantId]/components/Principal.tsx
@@ -1,20 +1,22 @@
-import { useWhoAmI } from '@/contexts/WhoAmIContext'
 import {
   DropdownMenu,
-  DropdownMenuSeparator,
-  DropdownMenuTrigger,
   DropdownMenuContent,
   DropdownMenuItem,
   DropdownMenuLabel,
+  DropdownMenuSeparator,
+  DropdownMenuTrigger,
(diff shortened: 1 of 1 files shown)


dashboard/src/app/(authenticated)/[tenantId]/components/Principal.tsx
```diff
@@ -1,20 +1,22 @@
...
 import { signOut } from 'next-auth/react'
-import { FC, Fragment } from 'react'
+import { FC } from 'react'
 function classNames(...classes: Array<string | boolean>) {
   return classes.filter(Boolean).join(' ')
 }
 
 export const Principal: FC = () => {
   const { user } = useWhoAmI()
+  const isAuthEnabled = process.env.LHD_OAUTH_ENABLED === 'true'
```
Comment (hazimoarafa) [inline dashboard/src/app/(authenticated)/[tenantId]/components/Principal.tsx:18]: Random question 

Where we check for oauth enabled in other files

Does it check for "true" or "True" or is it a truthy value

If it checks for "true" then you can merge

hazimoarafa (approved):

dashboard/src/app/(authenticated)/[tenantId]/components/Principal.tsx
Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/components/Principal.tsx:18]: image
`middleware.ts`

Issue #1332, opened 2025-02-24
UserTask with blank userId and userGroup


eduwercamacaro:
By definition user tasks can be either assigned to a `userId` or owned by a `userGroup`, but there are some cases where both `userId` and `userGroup` are null. Consider this example in:

```
return new WorkflowImpl(
            "example-ut",
            wf -> {
                WfRunVariable userId = wf.addVariable("userId", VariableType.STR);
                WfRunVariable userGroup = wf.addVariable("userGroup", VariableType.STR);
                wf.assignUserTask("my-task", userId, userGroup);
                wf.execute("my-task");
            }
        );
```

If you run this example without variables, the `UserTaskRun` will have null in both `userId` and `userGroup`.
I believe this can be solved by failing the UserTaskRun on arrival if the assignment is not valid.

eduwercamacaro:
https://github.com/littlehorse-enterprises/littlehorse/pull/1333 proposes a solution for this which fails with a `VAR_ERROR` if both `userId` and `userGroup` are null

PR #1333, opened 2025-02-24
fix(server): Validate null on UserTaskRun arrival for userId and userGroup

eduwercamacaro:
solves #1332


eduwercamacaro:
pinging @coltmcnealy-lh and @Snarr for review

Snarr (approved):

Issue #1341, opened 2025-02-27
Redirect to login page when token is invalid, inactive or expired


sauljabin:
### Context

The dashboard throws an error when the token is inactive and there is a session already opened.

### lhctl `~/.config/littlehorse.config` file

```
LHC_API_HOST=localhost
LHC_API_PORT=2023
LHC_API_PROTOCOL=TLS
LHC_OAUTH_CLIENT_ID=lhctl
LHC_OAUTH_SERVER_URL=http://localhost:8888/realms/lh
LHC_CA_CERT=/<PATH TO YOUR LH REPO>/local-dev/certs/ca/ca.crt
```

### Dashboard `.env` file

```
LHC_API_HOST=localhost
LHC_API_PORT=2023
LHC_API_PROTOCOL=TLS
LHD_OAUTH_ENABLED=true
LHC_CA_CERT=/<PATH TO YOUR LH REPO>/local-dev/certs/ca/ca.crt
NEXTAUTH_SECRET=anyrandomstring
KEYCLOAK_CLIENT_ID=dashboard
KEYCLOAK_CLIENT_SECRET=74b897a0b5804ad3879b2117e1d51015
KEYCLOAK_ISSUER_URI=http://localhost:8888/realms/lh
NEXTAUTH_URL=http://localhost:3000
```

### Setup with keycloak

```
./local-dev/issue-certificates.sh
./local-dev/setup.sh --keycloak
```

### Run server

```
./local-dev/do-server.sh oauth
```

### Run dashboard and login

```
npm install
npm run dev
```

Then login into the dashboard http://localhost:3000/api/auth/signin?callbackUrl=/

User: `user`
Pass: `password`

## Close session in keycloak

Go to http://localhost:8888/admin/master/console/#/lh/sessions and close the session:

![Image](https://github.com/user-attachments/assets/39bd49e5-e9c6-4c2d-a961-e552dc113c13)

## Important: restart the server

At this point there is a token in the frontend and in the lh server. But that token if invalid now.

Is you restart the server then when it receives the token again it's going to validate it and throws an error:

![Image](https://github.com/user-attachments/assets/a6c30d8e-e258-4805-b261-28df5f3f96e7)

In the backend:

![Image](https://github.com/user-attachments/assets/9ffc0e99-b26b-4ff1-a8ac-ff64ccf5071f)

### Expected Behavior

If the token is invalid it should not show an error to the user, it should redirect to the login page.
It should log an error or export an error metric.

[...]

PR #1342, opened 2025-02-27
fix: redirect on unauthenticated error

bryson-g:
- Redirects to the signin page in the `getWhoami` function if there is an `UNAUTHENTICATED` error


dashboard/src/app/getWhoami.ts
```diff
@@ -1,14 +1,27 @@
...
+    principal = await client.whoami({})
+  } catch (error: unknown) {
+    if (error instanceof ClientError && error.code === Status.UNAUTHENTICATED) {
+      redirect('/api/auth/signin')
+    } else {
+      throw error
+    }
+  }
+
+  const { id, perTenantAcls, globalAcls } = principal
```
mijailr, line 24: This won't work. We must validate the session in the middleware before reaching this point.

mijailr (changes requested):
To properly address this issue, we must:

- Verify if the token is expired.
- If expired but refresh token is valid, then request a refresh.
- If both refresh and token are expired then reset session and redirect to login.

All this must be handled in the middleware.

mijailr:
the problem lies here: https://github.com/littlehorse-enterprises/littlehorse/blob/f4c0850119ee486791782b753502107b7bf180c4/dashboard/src/middleware.ts#L9-L18

sauljabin:
Closes #1341

dashboard/src/app/getWhoami.ts
bryson-g: > This won't work. We must validate the session in the middleware before reaching this point.

@mijailr 
I agree, this validation should be done on the middleware, but there will be more tech debt if we add to the middleware because of the limitation of next not supporting node apis on the edge. If OAuth is enabled on dashboard, every request is validated by default, so this current fix does work.

PR #1350, opened 2025-03-11
feat(dotnet): Add tenantid header for grpc clients

eduwercamacaro:


sdk-dotnet/LittleHorse.Sdk.Tests/DefaultLHConfigVariables.cs
```diff
@@ -4,7 +4,7 @@ public static class DefaultLHConfigVariables
 { 
     public static string LHC_API_HOST => "localhost";
     public static int LHC_API_PORT => 2023;
-    public static string LHC_API_PROTOCOL => "PLAIN";
+    public static string LHC_API_PROTOCOL => "PLAINTEXT";
```
eduwercamacaro, line 7: @KarlaCarvajal 
Can you please confirm if this change is correct? 
I checked every sdk we have so far, and all of them expect `PLAINTEXT` for the default protocol.

KarlaCarvajal: The server configuration file introduces PLAIN as protocol listener, so I set the same in the file. According to local-dev files the config is ok
image

KarlaCarvajal: Just checked that the server expects PLAINTEXT as default protocol, so you are right.

KarlaCarvajal (approved):
LGTM, and kudos for eliminating unnecessary set accessors in variable configurations

sauljabin (approved):

Issue #1361, opened 2025-03-14
Idempotency for PutExternalEvent


eduwercamacaro:
Currently, `PutExternalEventRequest` ignores the idempotency key that we pass through the `guid` field. The implementation should honor the documentation for the idempotency documented on this request.

```proto
// Request used to create an ExternalEvent.
message PutExternalEventRequest {
  // The ID of the WfRun that this Event is sent to.
  WfRunId wf_run_id = 1;

  // The ID of the ExternalEventDef that this event implements.
  ExternalEventDefId external_event_def_id = 2;

  // Note that an ExternalEventDefId is a three-part ID:
  // 1. WfRunId
  // 2. ExternalEventDef Name
  // 3. A guid
  // The guid from part 3) can be optionally provided to the PutExternalEventRequest
  // in order to make it idempotent. It is a best practice to do so.
  optional string guid = 3;

  // The content of this event.
  VariableValue content = 5;

  // Optionally specify that this ExternalEvent may only be claimed by a specific ThreadRun.
  optional int32 thread_run_number = 6;

  // Optionally specify that this ExternalEvent may only be claimed by a specific NodeRun.
  // In order for this to be set, you must also set thread_run_number.
  optional int32 node_run_position = 7;
  reserved 4, 8;
}
```

eduwercamacaro:
I'm not sure how this should work when using specific `thread_run_number` and `node_run_position`.

PR #1363, opened 2025-03-14
fix(server): handles idempotency for external events

eduwercamacaro:
Solves #1361

### DIFF
--- a/lhctl/internal/post_event.go
+++ b/lhctl/internal/post_event.go
@@ -6,6 +6,7 @@ package internal
 import (
 	"errors"
 	"log"
+	"strconv"
 
 	"github.com/littlehorse-enterprises/littlehorse/sdk-go/lhproto"
 	"github.com/littlehorse-enterprises/littlehorse/sdk-go/littlehorse"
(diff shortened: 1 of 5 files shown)


sauljabin (approved):
lhctl and sdk-java were tested

lhctl/internal/post_event.go
```diff
@@ -40,6 +41,23 @@ lhctl postEvent <wfRunId> <externalEventName>
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		wfRunIdStr, eedName := args[0], args[1]
+		guid := cmd.Flag("guid").Value.String()
+		threadRunNumber := int32(-1)
+		nodeRunNumber := int32(-1)
```
Snarr, line 46: Since these are initialized to -1, will that have any negative effect on the server? I'm not sure if we perform any server-side validation that the number is greater than 0.

eduwercamacaro: I shouldn't send negative values to the server, I'll add a validation that prevents this to happen. 👍

eduwercamacaro:
@Snarr I applied your suggestions. Could you please check these changes?

Snarr (approved):

Issue #1364, opened 2025-03-14
Remove Awaitility dependencies from the sdk-java


eduwercamacaro:

coltmcnealy-lh:
Closed by #1365

PR #1365, opened 2025-03-14
chore(sdk-java): Remove awaitility dependency

eduwercamacaro:
solves #1364

### DIFF
--- a/sdk-java/src/main/java/io/littlehorse/sdk/worker/LHTaskWorker.java
+++ b/sdk-java/src/main/java/io/littlehorse/sdk/worker/LHTaskWorker.java
@@ -25,7 +25,6 @@ import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import lombok.extern.slf4j.Slf4j;
-import org.awaitility.Awaitility;
 
 /**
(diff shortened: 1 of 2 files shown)


sauljabin (approved):

sdk-java/src/main/java/io/littlehorse/sdk/worker/LHTaskWorker.java
```diff
@@ -161,21 +160,23 @@ public void registerTaskDef() {
...
-                                        .equals(Code.NOT_FOUND);
-                    })
-                    .until(() -> {
-                        this.taskDef = grpcClient.getTaskDef(
-                                TaskDefId.newBuilder().setName(taskDefName).build());
-                        return true;
-                    });
+            long start = System.currentTimeMillis();
+            long timeout = start + Duration.ofSeconds(2).toMillis();
+            do {
```
coltmcnealy-lh, line 165: Can we put a sleep for 10ms in here somewhere? So we don't spam the crap out of the server with a busy loop?

eduwercamacaro: Done ✅

PR #1370, opened 2025-03-18
fix(sdk-java): Improve Array detection when parsing TaskDef input types

### LINKED ISSUE #1360: Map java array object as JSON_ARR and not as JSON_OBJ
When registering a `TaskDef` with a list like this:

```java
@LHTaskMethod("my-task")
public void myTask(List<String> names) {

}
```

the `sdk-java` uses `JSON_ARR` as type:

```json
{
  "id": {
    "name": "my-task"
  },
  "inputVars": [{
    "type": "JSON_ARR",
    "name": "names",
    "maskedValue": false
  }],
  "createdAt": "2025-03-14T20:07:01.480Z"
}
```

but when using an array:

```java
@LHTaskMethod("my-task")
public void myTask(String... names) {

}
```

it declares a `JSON_OBJ`:

```json
{
  "id": {
    "name": "my-task"
  },
  "inputVars": [{
    "type": "JSON_OBJ",
    "name": "names",
    "maskedValue": false
  }],
  "createdAt": "2025-03-14T20:08:31.904Z"
}
```

It should be `JSON_ARR` in both cases.

Snarr:
This PR implements an additional method for detecting Array classes that users define for their TaskDef input parameters.

It uses the method `.isArray()` to determine if a given Java class is an Array, thus covering cases like `String[]` where the old implementation did not.

Resolves #1360.


sdk-java/src/test/java/io/littlehorse/sdk/common/LHLibUtilTest.java
```diff
@@ -60,6 +61,16 @@ void shouldIncludeParentWfRunidOnToString() {
         Assertions.assertThat(LHLibUtil.wfRunIdToString(id)).isEqualTo(parentId + "_" + childId);
     }
 
+    @Test
```
sauljabin, line 64: Can we add a test for Book[]?

sauljabin (approved):

sauljabin:
what happens with the worker? is it deserializing to a  String[]?

eduwercamacaro (approved):

Issue #1372, opened 2025-03-18
Cannot Mutate Workflow Variables to NULL


hazimoarafa:
Currently, attempting to assign a variable to NULL results in an error:

"errorMessage":  "Failed evaluating edge with sink node 3-nop-NOP: Mutating variable actions-to-take with operation ASSIGN: Coercing to or from NULL not supported."

This limitation prevents assigning a workflow variable to NULL, which should be possible.

Steps to Reproduce

The following workflow logic fails when trying to assign actionsToTake from a task result:

// Step 2: Determine if there are any actions to take
actionsToTake.assign(wf.execute(LHConstants.DETERMINE_ACTIONS_TASK, transcript).timeout(60 * 5));

// Step 3: Either run workflows or create user task
wf.doIfElse(actionsToTake.isEqualTo(null),
        noActionsHandler -> {
            // Require a human to verify that no actions are needed to be taken
        },
        yesActionsHandler -> {
            // Run the identified workflows
            yesActionsHandler.execute(LHConstants.RUN_WORKFLOWS_TASK, actionsToTake);
        });

Expected Behavior
	•	Assigning a variable to NULL should be supported.
	•	The doIfElse condition should correctly evaluate whether actionsToTake is NULL.

Actual Behavior
	•	Assignment fails with an error about coercion to/from NULL.

Additional Context

Task Worker:

```java
	@LHTaskMethod(LHConstants.DETERMINE_ACTIONS_TASK)
	public List<WorkflowExecutionRequest> determineActions(String transcript) {
		ChatRequest request = ChatRequest.builder()
				.messages(new UserMessage(
						"Analyze this customer support call transcript and determine what actions should be taken. If you decide no actions are needed please double check the message you sent back because you tend to mess up.:\n\n"
								+ transcript))
				.toolSpecifications(updateCustomerRecordsSpec, escalateCaseSpec, sendFollowUpEmailSpec)
				.build();
		ChatResponse response = model.chat(request);
		System.out.println(response.aiMessage());

		if (response.aiMessage().toolExecutionRequests() == null || response.aiMessage().toolExecutionRequests().isEmpty()) {
			return null;
		}

		return response.aiMessage().toolExecutionRequests().stream()
				.map(item -> new WorkflowExecutionRequest(item.id(), item.name(), item.arguments())).toList();
	}
```

[...]

PR #1374, opened 2025-03-19
fix(server): Support null assignments for workflow variables

eduwercamacaro:
solves #1372


eduwercamacaro:
@Snarr can you take at this please?

Snarr (approved):

Snarr:
Looks good 👍 I see this allows us to assign Null to a VariableValue, but have we tested returning Null from a Task Worker in our SDKs?

eduwercamacaro:
> Looks good 👍 I see this allows us to assign Null to a VariableValue, but have we tested returning Null from a Task Worker in our SDKs?

Yes, currently it is possible to return null values in a TaskWorker because the NodeOutput doesn't not enforce any type conversion or cast.

PR #1378, opened 2025-03-20
feat(server): Support virtual threads on network pool

eduwercamacaro:


server/src/main/java/io/littlehorse/server/streams/util/AsyncWaiters.java
```diff
@@ -34,12 +35,12 @@ public class AsyncWaiters {
...
     private final Lock eventWaiterLock = new ReentrantLock();
 
-    public AsyncWaiters(ScheduledExecutorService executor) {
+    public AsyncWaiters() {
         commandWaiters = new ConcurrentHashMap<>();
         eventWaiters = new HashMap<>();
-
-        executor.scheduleAtFixedRate(this::cleanupOldCommandWaiters, 0, 10, TimeUnit.SECONDS);
-        executor.scheduleAtFixedRate(this::cleanupOldWorkflowEventWaiters, 1, 10, TimeUnit.SECONDS);
+        ScheduledExecutorService scheduledService = Executors.newScheduledThreadPool(2);
```
coltmcnealy-lh, line 41: Do we really want to create two full platform threads for this work? That seems pretty wasteful. Can we find another threadpool somewhere to use for it? (At the very least we should use just one thread)

coltmcnealy-lh (approved):
These changes seem reasonable as it reduces the place we are passing the `networkThreadPool` (which, in all honesty, does  alot more than just networks) around.

And it decreases total thread usage by 14 (16 was the default, and we created a new threadpool of size 2). I like it overall. Good work!

coltmcnealy-lh (approved):
Looks good still! I'll need to pull it down and poke around but I love this.

coltmcnealy-lh:
Fixes #1574

Issue #1160, opened 2024-11-25
Address "Impossible" Log in `HandlingFailureHaltReasonModel`


coltmcnealy-lh:
When running our end-to-end tests, I observe the following log on the server:

```
01:55:53 WARN  [LH] HandlingFailureHaltReasonModel - Impossible: handling failure for a WaitThreadNode and found a non-terminated child
```

The e2e tests pass as expected. This means that some assumption that was made when developing `HandlingFailureHaltReasonModel.java` is incorrect. We should investigate what assumption that was and either remove the log or fix the bug.

eduwercamacaro:
IMO, the only thing missing here is handle the case when one of the child threads is in `HALTED` state. We should change to this instead:
```java
else if (wft.getThreadStatus() != LHStatus.COMPLETED || wft.getThreadStatus() != LHStatus.HALTED) {
    log.warn("Impossible: handling failure for a WaitThreadNode "
            + "and found a non-terminated child");
}
```


to

PR #1409, opened 2025-03-28
fix(server): Halted state should be considered as resolved when handling halt reasons on child threads

eduwercamacaro:
solves #1160


server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/haltreason/HandlingFailureHaltReasonModel.java
```diff
@@ -68,7 +68,8 @@ public boolean isResolved(ThreadRunModel haltedThread) {
                 for (WaitForThreadModel wft : handledNode.getWaitThreadsRun().getThreads()) {
                     if (wft.getThreadStatus() == LHStatus.ERROR || wft.getThreadStatus() == LHStatus.EXCEPTION) {
                         originalThatFailed.handledFailedChildren.add(wft.getThreadRunNumber());
-                    } else if (wft.getThreadStatus() != LHStatus.COMPLETED) {
+                    } else if (wft.getThreadStatus() != LHStatus.COMPLETED
```
Snarr, line 71: I am confused by the OR statement. If the WorkflowThread status is COMPLETED, then it will not be HALTED, and this will be true. If it is HALTED, then it will not be COMPLETED, and then this will be true. If it is some other status, then this will be true. Am I missing something?

eduwercamacaro: Oh you are right. Thank you. I think what we want is `&&`

eduwercamacaro: We want to print this warning when the WaitForThread status is not `COMPLETED` nor `HALTED`

Snarr (approved):

PR #1410, opened 2025-03-28
feat(sdk-python): load properties form dict

sauljabin:
Currently it is possible to load configurations from a dictionary, example:

```python
lh_config = LHConfig()
lh_config.configs = {"LHC_API_HOST": "localhost"}
```

But this PR enhances the way the sdk-python loads configurations.
- It encapsulates loading configurations from dictionaries
- It filters the configurations
- Change the name of the internal configuration object showing that it should be private

Now:

```python
lh_config = LHConfig()
lh_config.load({"LHC_API_HOST": "localhost"})
```


eduwercamacaro (approved):

Snarr (approved):
I like this solution a lot!

sdk-python/littlehorse/config.py
```diff
@@ -71,37 +71,53 @@ class LHConfig:
...
-                for key, value in self.configs.items()
+                for key, value in self._configs.items()
             ]
         )
 
-    def load(self, file_path: Union[str, Path]) -> None:
-        """Loads configurations from properties file.
+    def load(self, source: Optional[Union[str, Path, dict[str, Any]]] = None) -> None:
+        """Loads configurations from a source dictionary or properties file.
+        Environment variables have higher precedence than the loaded by this method.
```
Snarr, line 91: > ... than the loaded by this method

Maybe we can update this to:

> ... than the configurations loaded by this method

sauljabin: done

Snarr (approved):

PR #1412, opened 2025-03-31
feat: enable strong typing on external events

coltmcnealy-lh:


coltmcnealy-lh:
@Snarr and @eduwercamacaro the commit above is my proposed protobuf structure, balancing compatibility and forwards evolution. I have done my best to incoroporate all of your feedback from last week.

I will proceed with the implementation in subsequent commits.

Snarr:
@coltmcnealy-lh I am in support of this proposed structure 👍 As long as we rename `PedroReturnType` to something more descriptive in the final implementation 😉

coltmcnealy-lh:
As of now, we are missing five things:

1. Test to make sure that the validation actually works properly.
2. Convenience utility to create `PutExternalEventDefRequest` messages with type info.
3. Validating ExternalEventNode mutations against the type info in the `ExternalEventDef`
4. Validating Interrupt `INPUT` `VariableDef`'s against the type info in the `ExternalEventDef`
5. Deprecating untyped `ExternalEventDef`s

I think item `1` should be done in this PR and everything else should be an issue.

dashboard/src/app/utils/variables.ts
```diff
@@ -67,3 +67,19 @@ export const getTypedContent = (contentType: string, contentValue: string) => {
...
+ * Old server versions may keep around both old and new Variables, so this function
+ * determines which typing strategy a Variable uses.
+ */
+export const getVariableDefType = (varDef: VariableDef): VariableType => {
+  if (varDef.typeDef) {
+    return varDef.typeDef.type
+  } else if (varDef.type) {
+    return varDef.type
+  }
+  throw new Error('Variable must have type or typeDef.')
```
Snarr, line 84: This case should NEVER be reached unless there is a really bad bug on the server.

PR #1417, opened 2025-04-01
fix(lhctl): Update WfRun related commands to use proper arg rules

### LINKED ISSUE #1404: Change `wfSpecName` to argument in `lhctl search wfRun` command
Currently, `wfSpecName` is a required flag on the `lhctl search wfRun` command. I think, however, the command won't work if you don't supply this flag. So in that case, it should just be a positional argument.

Snarr:
This PR fixes some style issues with the `WfRun` related commands in `lhctl`, like a Required Argument being listed as a Flag. See https://github.com/littlehorse-enterprises/littlehorse/tree/master/lhctl for more info on our CLI style guide.

Resolves #1404


lhctl/internal/wf_run.go
```diff
@@ -36,25 +36,24 @@ var getScheduledWfRun = &cobra.Command{
 }
 
 var searchWfRunCmd = &cobra.Command{
-	Use:   "wfRun",
+	Use:   "wfRun <wfSpecName>",
```
coltmcnealy-lh, line 39: I like this change. However, there is one reason why we may want to be hesitant: in the future we will want to be able to search for `WfRun`'s by their `parent_wf_run_id`. One `WfRun` can have different `child WfRun`'s of different `WfSpec`s.

Do we want to allow searching for all `WfRun`'s of a certain type or do we want to require that a `wf_spec_name` is passed into the request? I could see the dashboard wanting to show all children of a given `WfRun` regardless of the `WfSpec`.

coltmcnealy-lh (approved):
Thanks!

coltmcnealy-lh:
I also realized we'll need to update the docs, specifically:
- the quickstart readme's
- the docs page for `lhctl`
- Any of our examples readmes

lhctl/internal/wf_run.go
Snarr: Let's make the `wfSpecName` argument optional, but still an argument. When/If we add searching by `parent_wf_run_id`, we can use it that as a flag.

PR #1421, opened 2025-04-04
chore(server): update kafka to 4.0.0

coltmcnealy-lh:
- Kafka 3.9.1 resolves this issue, which we have run into at times: https://issues.apache.org/jira/browse/KAFKA-12679


eduwercamacaro (approved):

coltmcnealy-lh:
Well, let's wait for the release....in our pipelines we trust :joy: 

I don't want to update to 4.0.0 though without doing some _seriously_ intense performance testing on AWS. But 3.9.1 was a pretty small release compared to 3.8.0, and we are still before LittleHorse 1.0, so I am fine with this after Kafka releases their stuff.

gradle.properties
```diff
@@ -1,6 +1,6 @@
 version=0.0.0-development
 group=io.littlehorse
-kafkaVersion=3.8.0
+kafkaVersion=3.9.1
```
mjsax, line 3: I would wait until 3.9.1 was released before merging this PR :)

coltmcnealy-lh: When Dr. Streams himself says to wait, I will wait!

However, by the time 3.9.1 is released, we will probably have finished the 4.0 testing already. We are waiting for Stirmzi 0.46 so that we can stress-test the 4.0 Streams client against the 4.0 brokers.

In particular, KIP-890 changes the transactions protocol, which will definitely affect LittleHorse. 890 isn't enabled by default in 3.9.x so I felt it was safe to upgrade without extensive testing, especially given that there are some known issues in 3.8.x that affect LH and are scheduled to be fixed in 3.9.1.

But I am eager for 4.0, as it fixes [KAFKA-17455](https://issues.apache.org/jira/browse/KAFKA-17455), which is the biggest issue affecting us.

coltmcnealy-lh: Actually, it looks like you backported `17455` to the `3.9.x` branch. Thanks so much @mjsax ! This is a huge help. As I said I'm nervous about 4.0 due to KIP-890, so this will buy us time to do more testing before the upgrade.

coltmcnealy-lh:
@eduwercamacaro you're running a soak test with 4.0 right now, aren't you? If that test passes, then let's close this PR and move to Kafka 4.0

PR #1427, opened 2025-04-11
refactor(server): increase default task timeout 15s -> 60s

hazimoarafa:


server/src/main/java/io/littlehorse/common/LHConstants.java
```diff
@@ -42,7 +42,7 @@ public class LHConstants {
             INTERNAL_ERROR,
             USER_TASK_CANCELLED));
 
-    public static final int DEFAULT_TASK_TIMEOUT_SECONDS = 15;
+    public static final int DEFAULT_TASK_TIMEOUT_SECONDS = 60;
```
eduwercamacaro, line 45: "Slow is the new down."

In the context of a machine task, 15 seconds is high enough in common cases where low latencies are expected. Why are we encouraging our users to have 60 seconds in the task timeout by default? This seems like a lot.

hazimoarafa: I get where you’re coming from, but 15 seconds isn’t cutting it. I ran into this firsthand when calling AI APIs like OpenAI — even successful calls can easily take 20–50 seconds. These aren’t rare edge cases, they’re becoming really common.

The whole point of a default is that it should just work without the user needing to think about it. If someone tries LittleHorse for the first time and calls an LLM or any slower API, they’ll hit a timeout and assume LH is broken. Sure, they can override it — but that kind of defeats the purpose of having a useful default in the first place.

This change isn’t about encouraging slow tasks — it’s about making sure the platform works out of the box for the workloads people are actually running today.

We’re not making slow the default, we’re making success the default.

eduwercamacaro: I'm sorry, but I don't agree with any of those statements.

This won't simply work for the typical deployment of a task worker because this timeout adds unnecessary latencies to the `WfRun` causing unexpected behavior, and users must be aware of this timeout in the majority of the cases.

However, I would agree on a proposal that allows users to overwrite the default timeout at the TaskDef level (maybe at task def registration). 

```java
   @LHTaskMethod(name = "my-slow-task", timeoutSeconds = 60)
```

hazimoarafa: @coltmcnealy-lh @Snarr what are your thoughts here?

mijailr: I don't believe this justifies modifying the default task timeout. Tasks that require longer execution times should leverage other features provided by the LH server to handle them asynchronously.

For example, in cases like the AI endpoint scenario shared by @hazimoarafa, where there are network calls between upstreams, the service implementing @LHTaskMethod should trigger an external event or a workflow event and reply immediately upon receiving the request, without waiting for the long-running operation to complete.

coltmcnealy-lh: At first, I agree with the Grumpy Maintainer (@eduwercamacaro) on this. However, I think this requires some more thought.

I'll ask some questions:

**In what case does the `TIMEOUT` error occur?** Hint: they happen when there's a `TaskClaimEvent` and no `ReportTaskRun` within 15 seconds (or whatever the timeout is). I have two specific cases in my mind.

**For each of these cases, what would happen if we wait a little bit longer before moving the `TaskRun` to `ERROR`? Would it be better to transition more quickly or wait longer for the `ReportTaskRun` to come in?**

And lastly:

**How likely is each case? Would we expect more of one or the other?**

After going through those questions, my opinion changed. @eduwercamacaro and @mijailr if you think through it I'm curious to see where you end up.

Snarr: Colt, following your train of thought here:

**In what case does the `TIMEOUT` error occur?**
As you mentioned, the `TIMEOUT` error will occur when a Task has been claimed but not reported back in the allotted Timeout time.

The two cases that come to my mind are:
1. Task Attempt takes too long
2. Task Worker dies

**For each of these cases, what would happen if we wait a bit longer before moving the `TaskRun` to `TIMEOUT` error?**

In the first case, waiting longer may mean receiving the Task Attempt.

In the second case, waiting 45s longer will mean the user's workflow fails more slowly, but not by much. It is marginal in most use cases.

**How likely is each case?**

I would say Case 1 is much more likely than Case 2.

You've led me to believe that a 60s default timeout could be a good improvement that has more pros than cons. I understand @eduwercamacaro and @mijailr's apprehensions though. It is good to be cautious about this, as we must be diligent about changing server defaults and configurations based on momentary problems.

hazimoarafa: I’m with Colt and Jacob on this one. Bumping the timeout to 60 seconds makes a lot of sense, especially for beginners who might not expect timeouts to happen so quickly. It’s way more user-friendly when you’re dealing with slower tasks, like AI API calls, that can easily take 20–50 seconds. Advanced users can still tweak things as needed, but the longer default timeout will prevent new users from getting frustrated with errors that make it seem like the platform is broken. It’s just a more forgiving approach that fits better with how things work today.

Snarr: @eduwercamacaro I do like the proposal of having a `withDefaultTaskTimeout` configuration, I suggested something similar to Colt in our discussions, but I was thinking more at the WorkflowThread level:

```java
new WorkflowImpl(
            "example-basic",
            wf -> {
                wf.setDefaultTimeoutSeconds(60);

                WfRunVariable theName = wf.addVariable("input-name", VariableType.STR).searchable();
                wf.execute("greet", theName);
            }
        );
```

We can discuss this model more in a discussion post or an issue perhaps. There may be other configurations we want to set at the global level, or other ways we can design this to make it more clear to users that the option is available.

Issue #1395, opened 2025-03-25
Reject empty strings as User Task User and UserGroup IDs on the server


Snarr:
We should reject empty strings as User Task User and UserGroup IDs on the server.

```java
public UserTaskOutputImpl assignUserTask(String userTaskDefName, Object userId, Object userGroup) {
  ...
  if (userGroup != null) {
      VariableAssignment userGroupAssn = assignVariable(userGroup);
      utNode.setUserGroup(userGroupAssn);
  }
  ...
}

```

The current method definition validates that the UserGroup is not null, but it does not validate that the User or UserGroup is not an empty string.

I cannot think of any value gained from setting the userGroup as an empty string, but I can name a ton of ways it could break other systems if developers don't take this option into account. For that reason, we should reject cases where users set an empty string for a User or UserGroup.

We should also have a larger discussion about whether or not any ID on the server should accept an empty string.

## Requirements

- Reject empty strings as UserTask User and UserGroupIds
  - Add this validation at the API level and the processor level of the Server
  - Add E2E test(s) to check that your added validation is correct

bryson-g:
UserIDs also have this same problem. Let's use the same fix for that too.
We should also do the same validation for any ID anywhere.

Context from @Snarr : 

![Image](https://github.com/user-attachments/assets/2380995d-589f-4d6c-a1bd-ca741af2d225)

PR #1428, opened 2025-04-13
fix(server): reject empty string values for user and user group Id's #1395

akuppa9:
- Added validation to ensure that `userId` and `userGroup` are not empty strings.

- Throws an IllegalArgumentException if an empty string is detected for either parameter.


Snarr:
Thanks for contributing!

To address issue #1395 completely we should also validate this on the server-side. `sdk-java` is just 1 of 4 SDKs for communicating with LittleHorse, and we can't trust that every client request will have this validation. Therefore, we should add this validation at the API level and also the processor lever in the server.

Snarr (changes requested):
We need some
- Server-side validation at the API level and the processor level
- E2E test

If you feel comfortable, you can also port this to the other SDKs  `lhctl`. If not, we can open an issue to tackle that.

Let me know if you have any questions, thanks again for submitting this contribution.

coltmcnealy-lh (commented):
I think @mauricio-nasser implemented this in #1395 ?

PR #1434, opened 2025-04-16
fix(sdk-java): set active thread to mutate a parent variable from child nested threads

KarlaCarvajal:
- Add unit tests

### DIFF
--- a/sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
+++ b/sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
@@ -58,12 +58,10 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Queue;
 import lombok.Getter;
-import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 @Getter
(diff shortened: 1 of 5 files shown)


sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
```diff
@@ -77,7 +75,8 @@ final class WorkflowThreadImpl implements WorkflowThread {
     private Queue<VariableMutation> variableMutations;
 
     public WorkflowThreadImpl(String name, WorkflowImpl parent, ThreadFunc func) {
-        this.parent = parent;
+        this.parent = Objects.requireNonNull(parent, "Parent workflow can not be null.");
+        ;
```
coltmcnealy-lh, line 79: Extra `;`

coltmcnealy-lh (approved):
Thanks!

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WorkflowThreadImpl.java
KarlaCarvajal: thanks!

Issue #1468, opened 2025-04-24
lhctl hangs after deleting externalEventDef


mitchell-h:
if you issue `lhctl delete externalEventDef <defName>` `lhctl` successfully deletes the `ExternalEventDef` but never returns.  You are forced to `CTRL + C` to exit the command and return to shell.

coltmcnealy-lh:
Steps to reproduce are just delete an `externalEventDef`? Normally, when it hangs, it means the "async waiter" on the server side is never completed.

mitchell-h:
Yes.  Reproduce with just doing `lhctl delete externalEventDef $NAME`.

coltmcnealy-lh:
Thanks @eduwercamacaro for fixing this. Closed by #1469

PR #1469, opened 2025-04-24
fix(server): Send an empty response for DeleteExternalEventDefRequest

eduwercamacaro:
solves #1468


eduwercamacaro:
@coltmcnealy-lh Can you confirm if an protobuf.Empty is ok for these cases? I think it is ok

coltmcnealy-lh:
> @coltmcnealy-lh Can you confirm if an protobuf.Empty is ok for these cases? I think it is ok

Yes, confirmed. If you look at the `rpc` I think it's `rpc DeleteExternalEventDef(ExternalEventDefId) returns (Empty) {}`

coltmcnealy-lh (approved):

Snarr (approved):

Issue #880, opened 2024-06-15
Schema Support on `VariableDef`


coltmcnealy-lh:
## Background

The `VariableDef` proto right now is as follows:
```
message VariableDef {
  // The Type of the variable.
  VariableType type = 1;

  // The name of the variable.
  string name = 2;

  // Optional default value if the variable isn't set; for example, in a ThreadRun
  // if you start a ThreadRun or WfRun without passing a variable in, then this is
  // used.
  optional VariableValue default_value = 3;
}
```

The only typing information we have is the `VariableType` which is simply an enum and contains no schema information. We are lacking the following functionality:

* Regex validation for `STR` variables. For example, enforcing that it matches an `email`.
* Schema for any `BYTES` variables. For example, we could include information to denote that the `BYTES` field is a protobuf, an Avro record, etc.
* Any form of Json Schema for `JSON_OBJ` or `JSON_ARR` variables. Currently, we don't even validate that it is a proper valid Json String, much less check for presence of any fields.

## Proposal

I propose to introduce a new Tenant-scoped `GlobalGetable` object called `VariableSchema`. It would look like this:

```
message VariableSchemaId {
  string name = 1;
  int32 version = 2;
}

message VariableSchema {
  // Id of the schema
  VariableSchemaId id = 1;

  // human-readable description
  string description = 2;

  oneof schema {
    // An Open-API v3 Schema
    OpenApiV3Schema open_api = 3;
    
    // Protobuf schema
    ProtoBufSchema proto_schema = 4;

    // String Regex
    StringRegex string_regex_schema = 5;
  }
}
```

Protobuf already has a well-defined "Proto Descriptor" API for sharing protobuf schemas. For OpenAPIv3, we can use the [ApiCurio Data Models Library](https://www.apicur.io/datamodels/). Lastly, String regexes should be easy enough to match using a `Pattern`.

## Discussion
The discussion section takes it for granted that LittleHorse should adopt some form of schema management solution. Instead we focus on the specific implementation proposed above.

### Benefits
The two most crucial benefits of what is proposed above are:
1. No additional external dependencies are introduced to the server.
2. Because the schemas can be managed inside the global KTable, there is no performance penalty (compared with using an external Schema Registry in which we would have to make external network calls).

Further benefits are that:
* The schemas can be viewed inside our dashboard.
[omitted internal LH Cloud reasons]

### Drawbacks
* Implementing schema management like this will be considerable effort.
[...]

PR #1471, opened 2025-04-26
doc(arc): add struct and structdef proposal

coltmcnealy-lh:
Proposal for the ticket #880 .

### DIFF
--- a/proposals/000-struct-and-structdef.md
+++ b/proposals/000-struct-and-structdef.md
@@ -1,0 +1,470 @@
+# Proposal: `StructDef` and `Struct`
+
+Author: Colt McNealy
+
+We want schemas on the LH Server for several reasons:
+
+* Add type safety to WfSpec’s.
+* Allow users of the upcoming _Output Topic_ feature to safely perform joins and transformations without worrying about type casting failures.
+* Better developer experience so that @sauljabin is happier :smile:
+
+We already have strong typing for primitive types (`INT`, `STR`, `DOUBLE`, `BOOL`). However, the `JSON_OBJ` and `JSON_ARR` fields are YOLO Wild West, which has caused problems and errors for our users.
+
+This proposal proposes to introduce a new type of variable, a `Struct` (with an associated `StructDef` metadata object). The `Struct` is intended to replace the `JSON_ARR` and `JSON_OBJ` fields in the long term, and if we implement it properly (with sufficient support in the SDK-side), ideally we will have equally or even more convenient developer experience while introducing strong typing.
+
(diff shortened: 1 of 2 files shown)


proposals/000-struct-and-structdef.md
```diff
@@ -0,0 +1,495 @@
...
+
+1. The `car` `StructDef` does not exist.
+2. The `car` `StructDef` is incompatible with the `Car` object (only in the second case above).
+
+#### Automatic `StructDef` Creation
+
+The `LHTaskWorker` should be extended to allow automatically creating a `StructDef` as per the following code:
+
+```java
+@LHStructDef(name = "car")
```
Snarr, line 351: Could have a version as well

Snarr (approved):

proposals/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+Proposals are similar to Kafka Improvement Proposals (KIP's) in Apache Kafka and [Strimzi Proposals](https://github.com/strimzi/proposals) in Strimzi. Proposals should be written before significant features which have significant impact on:
+
+* Any changes to the Protocol Buffer API (in the [`schemas` directory](../schemas/)).
+* Significant changes to the SDK API in Java, Go, Python, or C#.
+* Significant changes to the internal architecture of the Server which might affect performance.
+
+Proposals should include:
+
+* The motivation for the new feature or change.
```
Snarr, line 15: Let's make a proposal template

coltmcnealy-lh: Hmm. I'm not a huge fan of templates, but we can have one as a guideline which is optional. Thanks!

proposals/000-struct-and-structdef.md
coltmcnealy-lh: Hmm. There's no version in the `Workflow` utility for registering `WfSpec`s (and there's no version in the `rpc RegisterWfSpec`). The idea is that the `rpc PutStructDef` behaves the same as `rpc PutWfSpec`.

eduwercamacaro (approved):

Snarr:
@coltmcnealy-lh 

From my conversations with @eduwercamacaro, we were wondering what the purpose of `NO_SCHEMA_UPDATES` is. I couldn't pinpoint an exact use case where the server would need to receive a `PutStructDefRequest` that doesn't intend to update a `StructDef`.

There is one case where the client doesn't want to update a `StructDef` if it already exists, but in that case why can't the client just call `GetStructDef` to see that it already exists, and then conditionally call `PutStructDefRequest`?

PR #1472, opened 2025-04-28
test(sdk-python): Add tests validating lambda support

Snarr:


sdk-python/tests/test_workflow.py
```diff
@@ -1621,6 +1675,15 @@ def test_entrypoint_is_a_function(self):
             str(exception_context.exception),
         )
 
+    def test_no_error_on_lambdas_for_if_statements(self):
+        def my_entrypoint(thread: WorkflowThread) -> None:
+            thread.do_if(thread.condition(5, Comparator.EQUALS, 5), lambda wf: wf.execute("my-task"))
+
+        try:
+            Workflow("my-wf", my_entrypoint)
```
KarlaCarvajal, line 1683: Maybe I am wrong, but I believe lambdas with 1 line of code are supported in python, this test will never fail, so I think, with the one above should be enough.

Snarr: It could fail if we change something internally about the way we handle WorkflowThreads, like adding back the If Statements that validate the params and return types (which I removed in the last release). I think it is worthwhile to keep

KarlaCarvajal: But, all tests (test_do_if_with_lambda, test_no_error_on_lambdas_for_if_statements) using that structure will fail
 ```python
wf.do_if(
                wf.condition(5, Comparator.EQUALS, 5),
                lambda wf: wf.execute("my-task")
            )```

coltmcnealy-lh (approved):

PR #1480, opened 2025-04-30
fix(sdk-python): add validation to throws an error when do_else is called multiple times

KarlaCarvajal:


Snarr (approved):
Thanks for catching this bug! Great improvement

PR #1484, opened 2025-05-01
doc(proposals): adds output topic proposal

coltmcnealy-lh:


proposals/001-output-topic.md
```diff
@@ -0,0 +1,366 @@
...
+    updated_at TIMESTAMPTZ DEFAULT now()
+);
+```
+And then insert a new row for every single `WfRun`. This would allow you to do analytics based on your orders.
+
+This can be accomplished with an _Output Topic_ that publishes updates to `WfRun` data in real time to Apache Kafka.
+
+### Workflows as Streams
+
+Another motivation for the Output Topic is that updates to your `WfRun`s can be streams of events. For example, the following use-cases have come up:
```
KarlaCarvajal, line 62: I think, this part `updates to your `WfRun`s can be streams of events` will be more clear if we specify that events are external events.

coltmcnealy-lh: They are not external events though.

proposals/001-output-topic.md
```diff
@@ -0,0 +1,366 @@
...
+    WfSpec wf_spec = 1;
+    TaskDef task_def = 2;
+    ExternalEventDef external_event_def = 3;
+    WorkflowEventDef workflow_event_def = 4;
+    UserTaskDef user_task_def = 5;
+  }
+}
+```
+
+There is no need to register `Principal`s, since they're not scoped to the tenant, and there's not much useful business info in the creation of `Principal`s. Obviously, since the Output Topic is tenant scoped, we don't put `Tenant`s there.
```
KarlaCarvajal, line 298: Metadata and Execution Output topics will be created per tenant, but this text is saying that output topics will not include tenant info, so end users with multiple tenants might not know where the info comes from. Is the tenant going to be added as part of the topic name?

coltmcnealy-lh: Yes, that's a good point. I think the proposal should mention the naming strategy for the topics.

coltmcnealy-lh:
> For example, are data engineers familiarized with protobuf?, I believe the standard is json.

Great point, @sauljabin. The reason I chose protobuf is that:

- Server implementation is easiest if we send it out in protobuf and we avoid the performance penalty of dealing with JSON.
- Our SDK's will automatically include the ability to deserialize the output topic events.

In the future we could even open a proposal to allow our SDK's to integrate with the Kafka Consumer API by providing deserializers or something like that.

If needed, our users can write a simple stateless Kafka Streams app (or a Kafka Connect Converter) that converts the records from proto to json.

proposals/001-output-topic.md
```diff
@@ -0,0 +1,376 @@
...
+### Metadata and Execution Data
+There are two types of data in LittleHorse:
+1. Metadata, such as `WfSpec`, `TaskDef`, etc.
+2. Execution data, such as `WfRun`, `TaskRun`, etc.
+
+Metadata is small, relatively static, and global to a cluster. Execution data is large, partitioned, and constantly changing. Consumers doing stream processing on Execution data will often need access to Metadata in order to properly make sense of the Execution Data.
+
+Therefore, we will separate metadata and execution data into two topics:
+1. The `metatadata-output` topic, which is a single-partition, compacted topic containing metadata updates.
+2. The `execution-output` topic, which is a multi-partition, non-compacted topic containing execution data updates.
```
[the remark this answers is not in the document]
coltmcnealy-lh: Fantastic question. I think that should be done by the cluster operator/admin not the creator of the Tenant. Maybe we can use LHS_CLUSTER_PARTITIONS?

PR #1490, opened 2025-05-05
perf(server): avoid sending extra writes to rocksdb

coltmcnealy-lh:
In the GetableManager#commit() method, we previously put every single GetableToStore into rocksdb. THat's problematic because when we *read* a getable without modifying it, it ends up inside the buffer. This PR reduces the writes to RocksDB by saving the previous protobuf and comparing to the protobuf at the time of commit. This costs a bit more CPU at the server-side but reduces the number of writes to rocksdb.

### DIFF
--- a/server/src/test/java/io/littlehorse/storeinternals/GetableManagerTest.java
+++ b/server/src/test/java/io/littlehorse/storeinternals/GetableManagerTest.java
@@ -452,4 +452,109 @@ public class GetableManagerTest {
                 Arguments.of(variable, 3),
                 Arguments.of(externalEvent, 4));
     }
+
+    @Test
+    void dontStoreGetableWhenNotModified() {
+        String varName = "my-str";
+        String wfRunId = "my-wf-run-id";
+        String valueBefore = "valueBefore";
+        String anotherValue = "anotherValue";
+        VariableModel actualVariable = TestUtil.variable(wfRunId);
+        actualVariable.getId().setName(varName);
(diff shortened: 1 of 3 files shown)


Snarr (approved):
Amazing 🎉

server/src/test/java/io/littlehorse/storeinternals/GetableManagerTest.java
```diff
@@ -452,4 +452,109 @@ private static Stream<Arguments> provideGetableObjectsAndIds() {
                 Arguments.of(variable, 3),
                 Arguments.of(externalEvent, 4));
     }
+
+    @Test
+    void dontStoreGetableWhenNotModified() {
```
eduwercamacaro, line 457: IMO, these two tests are very confusing to me. Even though you'r adding a comment for each section on the test, I don't think the implementation shows the real expected behavior. I'm wondering whether it is possible to use the `Sensor` instance to exactly know the number of `put`s in the store, or we can create a wrapper for the `KeyValueStore` instance that allows of to know the number of puts.

eduwercamacaro (approved):
This implementation is not ideal as it adds extra `toProto` calls, maybe I can add another PR to propose an alternative to this in the future.

eduwercamacaro:
Bonus check:

Is it possible to run the e2e tests and then compare the number of entries in the changelog topics and/or the number of writes into rocksdb?

PR #1500, opened 2025-05-07
feat(server): implements output topic metadata management

coltmcnealy-lh:
TODO:
- handle kafka topic creation if LHS_SHOULD_CREATE_TOPICS == true
- Construct models for the output topic records
- Intercept and send the records in the ProcessorExecutionContext


eduwercamacaro (approved):

server/src/main/java/io/littlehorse/common/LHServerConfig.java
```diff
@@ -353,6 +354,11 @@ public int getClusterPartitions() {
         return Integer.valueOf(String.class.cast(props.getOrDefault(LHServerConfig.CLUSTER_PARTITIONS_KEY, "12")));
     }
 
+    public int getOutputToppicPartitions() {
```
KarlaCarvajal, line 357: here we have a misspelled word `toppic` getOutputToppicPartitions

Issue #1309, opened 2025-02-19
Update the `rpc PutTenant` to allow tenant-wide output topic configuration.


coltmcnealy-lh:

coltmcnealy-lh:
The first step of this issue is done by #1500 and will be completed by #1502

coltmcnealy-lh:
Closed by #1502

PR #1502, opened 2025-05-08
feat(server): create output topics if configured to do so

coltmcnealy-lh:
This is a DRAFT that would finish the second part of #1309

### DIFF
--- a/server/src/main/java/io/littlehorse/common/LHServerConfig.java
+++ b/server/src/main/java/io/littlehorse/common/LHServerConfig.java
@@ -5,6 +5,7 @@ import io.grpc.ChannelCredentials;
 import io.grpc.ServerCredentials;
 import io.grpc.TlsChannelCredentials;
 import io.grpc.TlsServerCredentials;
+import io.littlehorse.common.model.getable.global.acl.TenantModel;
 import io.littlehorse.common.util.LHProducer;
 import io.littlehorse.common.util.RocksConfigSetter;
 import io.littlehorse.sdk.common.config.ConfigBase;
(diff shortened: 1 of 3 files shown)


coltmcnealy-lh:
@eduwercamacaro and @Snarr I want to talk more about this. There are a few points here so I'll number them.

1. For performance reasons, we don't want to be creating topics _inside_ the Kafka Stream Thread. Can we delegate the topic creation to the network thread pool or something else like that?
2. I assume that we should say "setting `LHS_SHOULD_CREATE_TOPICS` is never recommended for production use." Given that, how do we want to handle cases where topic creation fails? I see a few options:
  a. Just silently fail, since it's only for local development and not production.
  b. Retry on a scheduled executor.
  c. Expose an error message on the `Tenant` resource itself somehow. (Does this mean that we'd have to do the creation _inside_ the `MetadatProcessor` so we can modify the `Tenant` resource?
3. What if `LHS_SHOULD_CREATE_TOPICS` is set to false, and when we try to produce to the output topic it fails? I think we should swallow the failure and have a metric that says "can't product to output topic" so cluster administrators can monitor+alert on it.

My opinion is that for production deployments the output topics should be properly created _before_ creating the `Tenant`, just like you should create the changelog and command topics _before_ starting the server. So perhaps the design of this feature isn't that important.

We could (in the future) also introduce a few more configs:
- `LHS_SHOULD_CREATE_OUTPUT_TOPICS` which defaults to the value of `LHS_SHOULD_CREATE_TOPICS`
- Output topic retention as a _server-level_ configuration.

eduwercamacaro:
> For performance reasons, we don't want to be creating topics inside the Kafka Stream Thread. Can we delegate the topic creation to the network thread pool or something else like that?
I believe we can handle the output topic creation in the request context after the `PutTenantRequest` command is completed.

> Expose an error message on the Tenant resource itself somehow. (Does this mean that we'd have to do the creation inside the MetadatProcessor so we can modify the Tenant resource?

I'm in favor of this idea, but I have to say that adding a output_topic_config in the tenant is not my favorite thing on this implementation.

> What if LHS_SHOULD_CREATE_TOPICS is set to false, and when we try to produce to the output topic it fails? I think we should swallow the failure and have a metric that says "can't product to output topic" so cluster administrators can monitor+alert on it.

Totally agree

coltmcnealy-lh:
Closing in favor of #1509 which includes these changes

PR #1508, opened 2025-05-10
feat(server): Add `StructDef`s

Snarr:
This PR adds `StructDef` objects to the protobuf Schema and to the server according to the specification in [Proposal #000 Struct and StructDef](https://github.com/littlehorse-enterprises/littlehorse/blob/master/proposals/000-struct-and-structdef.md). I plan to implement the ADR incrementally, and this is the first step in doing so. The next step will be adding client-side helpers for generating StructDefs.

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/getable/global/structdef/InlineStructDefModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/getable/global/structdef/InlineStructDefModel.java
@@ -1,15 +1,11 @@
 package io.littlehorse.common.model.getable.global.structdef;
 
 import com.google.protobuf.Message;
-import io.grpc.Status;
 import io.littlehorse.common.LHSerializable;
-import io.littlehorse.common.exceptions.LHApiException;
-import io.littlehorse.common.util.LHUtil;
 import io.littlehorse.sdk.common.exception.LHSerdeException;
 import io.littlehorse.sdk.common.proto.InlineStructDef;
 import io.littlehorse.sdk.common.proto.StructFieldDef;
 import io.littlehorse.server.streams.topology.core.ExecutionContext;
-import java.text.MessageFormat;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Map.Entry;


server/src/main/java/io/littlehorse/common/model/getable/global/taskdef/TaskDefModel.java
```diff
@@ -86,7 +86,6 @@ public void initFrom(Message p, ExecutionContext context) {
         for (VariableDef entry : proto.getInputVarsList()) {
             inputVars.add(VariableDefModel.fromProto(entry, context));
         }
-        returnType = LHSerializable.fromProto(proto.getReturnType(), ReturnTypeModel.class, context);
```
Snarr, line 89: NOTE: This line is written twice, so I am removing one of the calls

server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/TypeDefinitionModel.java
```diff
@@ -59,6 +59,12 @@ public boolean isPrimitive() {
         return false;
     }
 
+    public static TypeDefinitionModel fromProto(TypeDefinition proto, ExecutionContext context) {
```
eduwercamacaro, line 62: I don't think we need this method as it is already defined in a parent class `LHSerializable`. So users of this class can do something like this:
```java
TypeDefinitionModel.fromProto(proto.getFieldType(), TypeDefinitionModel.class, context);
```

eduwercamacaro: Nvm. I just realized that this method is better because it doesn't use reflection to create a new instance.

eduwercamacaro (approved):

schemas/littlehorse/acls.proto
```diff
@@ -117,6 +117,9 @@ enum ACLResource {
 
     // Refers to `WorkflowEventDef` and `WorkflowEvent`
     ACL_WORKFLOW_EVENT = 8;
+
+    // Refers to `StructDef` and `Struct`
+    ACL_STRUCT = 9;
```
coltmcnealy-lh, line 122: Hmm. Just to be super nit-picky, I would say maybe change the comment to:

```
  // Refers to `StructDef`
```

There are no rpc's that directly edit `Struct`s, as far as I know. Everything can _use_ a `Struct`. Aside from `ACL_WORKFLOW_EVENT` and `ACL_STRUCT`, all of the other acl resources involve RPC's that modify the stuff directly (eg. `rpc RunWf`, `rpc ReportTaskRun`, `rpc PutExternalEvent`, `rpc CompleteUserTaskRun`).

schemas/littlehorse/service.proto
```diff
@@ -252,6 +268,8 @@ service LittleHorse {
   rpc DeleteWfRun(DeleteWfRunRequest) returns (google.protobuf.Empty) {}
   // Deletes a TaskDef.
   rpc DeleteTaskDef(DeleteTaskDefRequest) returns (google.protobuf.Empty) {}
+  // Deletes a StructDef.
+  rpc DeleteStructDef(DeleteStructDefRequest) returns (google.protobuf.Empty) {}
```
coltmcnealy-lh, line 275: I'm wondering if we want to refactor all of these `DeleteFoo` rpc's to return the object that was deleted or throw `NOT_FOUND` otherwise.

schemas/littlehorse/service.proto
```diff
@@ -355,6 +373,35 @@ message PutTaskDefRequest {
...
+    // Allowed to make fully compatible (both backward-and-forward compatible)
+    // changes to the `struct_def` in this request.
+    FULLY_COMPATIBLE_SCHEMA_UPDATES = 1;
+  }
+
+  // The name of the StructDef.
+  string name = 1;
+
+  // The descripton of the StructDef.
+  optional string description = 2;
```
coltmcnealy-lh, line 385: This makes me happy. We should update all of the other metadata rpc's and `Getable`s as well.

server/src/main/java/io/littlehorse/common/model/getable/global/structdef/InlineStructDefModel.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public Class<InlineStructDef> getProtoBaseClass() {
+        return InlineStructDef.class;
+    }
+
+    public void validate() {
+        for (Entry<String, StructFieldDefModel> field : fields.entrySet()) {
+            if (!LHUtil.isValidLHName(field.getKey())) {
```
coltmcnealy-lh: I'm not sure we want to enforce that field names are valid hostnames. The motivation for that convention was that the Getable ID's became keys in RocksDB, and we use special characters in RocksDB to delimit things (eg `/` to delimit `/{tenant}/{storeableType}/{getableType}/etc`, and `_` and `__` to delimit tags).

I think we should allow field names like `graduationYear` or `graduation_year` etc. This is especially salient if we start inferring field names automatically via reflection and c

But that begs a bigger question: should we _encourage_ a certain standard? For example, the pythonic way is `graduation_year` but the Javonic way is `graduationYear`. In LH, a `StructDef` and `Struct` are intended to be used in multiple languages.

In proto, everything is `snake_case` field names but then in each language it's serialized into the local dialect.

server/src/main/java/io/littlehorse/common/model/getable/global/structdef/StructDefModel.java
```diff
@@ -0,0 +1,107 @@
...
+        return List.of();
+    }
+
+    @Override
+    public Class<StructDef> getProtoBaseClass() {
+        return StructDef.class;
+    }
+
+    public void bumpVersion() {
+        id.setVersion(id.getVersion() + 1);
```
coltmcnealy-lh: This _might_ break things given how the storage managers work. At least in the Core processor, if you:

```
WfRunModel wfRun = getableManager.get(wfRunId);
wfRun.setId(otherId);
getableManager.commit();
```
You might end up overwriting the old `WfRunId` and not even saving the new one...some weird things could happen.

@eduwercamacaro what do you think? That sounds like we need to put some more checks + unit tests in the `GetableManager`. Also I think in the metadata topology we don't have as much magic, so this might still be fine over there.

server/src/test/java/e2e/StructDefLifecycleTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+        PutStructDefRequest updatedStructDef = PutStructDefRequest.newBuilder()
+                .setName("car-3")
+                .setAllowedUpdates(AllowedStructDefUpdateType.FULLY_COMPATIBLE_SCHEMA_UPDATES)
+                .setStructDef(InlineStructDef.newBuilder()
+                        .putFields(
+                                "model",
+                                StructFieldDef.newBuilder()
+                                        .setFieldType(
+                                                TypeDefinition.newBuilder().setType(VariableType.STR))
```
coltmcnealy-lh: Beautiful test. Can we add another test case that does something like change `model` from `INT` to `STR`?

server/src/main/java/io/littlehorse/server/streams/topology/core/WfService.java
```diff
@@ -83,6 +85,25 @@ public WorkflowEventDefModel getWorkflowEventDef(WorkflowEventDefIdModel id) {
...
+     */
+    public StructDefModel getStructDef(String name, Integer version) {
+        Supplier<StructDefModel> findStructDef = () -> {
+            final StructDefModel storedResult;
+
+            if (version != null) {
+                storedResult = metadataManager.get(new StructDefIdModel(name, version));
+            } else {
+                storedResult =
+                        metadataManager.getLastFromPrefix(StructDefIdModel.getPrefix(name), StructDefModel.class);
```
coltmcnealy-lh, line 99: I feel young again reading this code. It's been two years since I've touched the equivalents for `WfSpec`s and `TaskDef`s.

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutStructDefRequestModel.java
```diff
@@ -0,0 +1,134 @@
...
+        throw new LHApiException(Status.INVALID_ARGUMENT, errorMessage.toString());
+    }
+
+    @Override
+    public Class<PutStructDefRequest> getProtoBaseClass() {
+        return PutStructDefRequest.class;
+    }
+
+    public void validate() {
+        if (!LHUtil.isValidLHName(name)) {
```
coltmcnealy-lh: Small nitpick. Is it necessary to put this in a helper method? It kind of threw me off since I thought `validate()` was going to do the compatibility checking. I think (since this method is so short) we can just put the lines directly in `process()`

coltmcnealy-lh (approved):
Beautiful work.

server/src/test/java/e2e/StructDefLifecycleTest.java
Snarr: Already added in my local changes from Friday, will push soon 👍

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutStructDefRequestModel.java
Snarr: No it is not necessary in this instance, good catch. I was just following a pattern I setup in other Model classes where it did make more sense, like giving `StructDefModel` its own `validate()` method that can be used in multiple request classes like Validate and Put StructDef.

server/src/main/java/io/littlehorse/common/model/getable/global/structdef/InlineStructDefModel.java
Snarr: I think we should enforce or convert to a certain standard. 

Assuming we infer field names using reflection in all languages one day:

If you generate a Car `StructDef` in Python (using reflection on class fields), but then want to move the `StructDef` to Java (using reflection on class fields), the fields should match up automatically to a common standard so that the `StructDef` can be evolved properly while keeping your code inline with a given language's conventions.

This aligns with a notion that I have always had that a `WfSpec` generated with Java should match the `WfSpec` generated with Python, as should `StructDef`s assuming we use reflection for inferring fields eventually.

Snarr: - Should the server reject a request if the StructDef field names don't a specific naming style (`snake_case`, `camelCase`, etc...)
- If no, what kinds of names should the server reject? Perhaps ones with hyphens `-` and other special characters. Maybe only A-Z, 0-9, and underscores.
- The SDKs should automatically convert from the given languages standard to the LittleHorse WfSpec naming standard

Snarr: My opinion:

- The server should only accept characters A-Z, 0-9, and underscores
- Should the server reject a request if the `StructDef` field names don't follow a specific naming style? I say no, we should accept most naming styles compatible with the above characters like how you can break the naming conventions in Java
- The SDKs should automatically convert to Java style `camelCase` field names

eduwercamacaro: As a platform, Littlehorse server should enforce a certain standard by rejecting the requests that does not follow the naming standard for structs fields. Therefore, client sdks must send a valid name for the fields within a schema. 

I propose to enforce snake case as the standard for Littlehorse server.
However, I think defining which characters are allowed deservers a separate discussions, for we should cover all the possible valid names for a field.

server/src/main/java/io/littlehorse/common/model/getable/global/structdef/StructDefModel.java
eduwercamacaro: I'm always in favor of immutability as it is less error prone and it is usually more predictable. 

I do think the `bumpVersion` should return a new StructDefModel with a higher version.

server/src/main/java/io/littlehorse/common/model/getable/global/structdef/InlineStructDefModel.java
Snarr: @eduwercamacaro That sounds good. As for the characters allowed discussion, I want to address that matter before releasing `StructDef`s so that users do not create fields with characters that will one day be invalid. I want to address everything that we can before `StructDef`s V1 is released.

eduwercamacaro: @Snarr Absolutely. 
We should address it before releasing schemas.

PR #1518, opened 2025-05-13
(sdk-python): update do_if method usage in examples

KarlaCarvajal:


KarlaCarvajal:
Hi team, @Snarr @eduwercamacaro please take a look at this code. I will be happy to receive some feedback :1st_place_medal:

sdk-python/examples/parallel_approvals/parallel_approvals.py
```diff
@@ -56,24 +54,39 @@ def is_user(user_thread: WorkflowThread) -> None:
...
-            condition=condition, if_body=is_user_group, else_body=is_user
-        )
+        approval_thread.do_if(condition, is_user_group).do_else(is_user)
 
     def my_entrypoint(wf: WorkflowThread) -> None:
-        approvals_var = wf.add_variable("approvals", VariableType.JSON_ARR)
-        (wf.add_variable("item-url", VariableType.STR).searchable())
-        (wf.add_variable("status", VariableType.STR, "PENDING").searchable())
+        approvals_var = wf.declare_json_arr("approvals")
+        (wf.declare_str("item-url").searchable())
```
sauljabin, line 61: I believe that the parentheses are not needed

sdk-python/littlehorse/workflow.py
```diff
@@ -1944,7 +1952,8 @@ def do_if(
 
         self._validate_initializer(else_body)
         self._do_if(condition, if_body).do_else(else_body)
-        return None
+
+        return WorkflowIfStatement(self, "", "")
```
sauljabin, line 1956: can you explain why?

sauljabin (approved):

sdk-python/littlehorse/workflow.py
Snarr: Karla is adding this for the edge case presented in the test `test_should_throw_an_error_compiling_a_wf_when_else_body_and_do_else_are_used`.

KarlaCarvajal: Hi @sauljabin, it was because an error with `poetry run mypy .` that showed that having an  `Optional[WorkflowIfStatement]` can be None and it does not have the `do_else` or `do_else_if` methods, so this solution is returning an WorkflowIfStatement with empty params, and it adds the validations during `do_else` and `do_else_if` calls. What do you think?

sauljabin: It's okay, I mean, by general rule try to avoid returning null (even in java), when it is not possible well, return null.
So it is good to return an empty object instead of none.

Snarr (approved):

PR #1524, opened 2025-05-14
fix(sdk-dotnet): replace threadvar type by typedef and fix unit tests

KarlaCarvajal:


sdk-dotnet/LittleHorse.Sdk/Workflow/Spec/WfRunVariable.cs
```diff
@@ -1,3 +1,4 @@
+using Google.Protobuf.Reflection;
```
Snarr: Why did this add a new import? 🤔

Snarr (approved):

PR #1529, opened 2025-05-15
docs(proposals): Revise `Proposal #000` protobuf based on implementation

Snarr:
Revises the protobuf definitions in `Proposal #000` based on implementation findings in #1508.


proposals/000-struct-and-structdef.md
```diff
@@ -234,16 +240,19 @@ message PutStructDefRequest {
   // The name of the `StructDef`.
   string name = 1;
 
+  // The description of the `StructDef`.
+  optional string description = 2;
+
   // The actual schema for the `StructDef`.
-  InlineStructDef struct_def = 2;
+  InlineStructDef struct_def = 3;
```
KarlaCarvajal, line 247: I thought that protobuf does not allow to change the Ids of the fields in the structs. I will read more though

proposals/000-struct-and-structdef.md
```diff
@@ -89,6 +89,9 @@ message StructFieldDef {
 
   // The type of the field.
   TypeDefinition field_type = 3;
+  
+  // The default value of the field, which should match the Field Type.
+  optional VariableValue default_value = 4;
```
eduwercamacaro, line 94: We are trying to represent `null` values with an optional here, but the `VariableValue` proto already covers this case by using `one_of`. You can represent `null` in a LittleHorse type in this way:


```java
        VariableValue build = VariableValue.newBuilder().build();
        VariableValueModel variableValueModel = VariableValueModel.fromProto(build, executionContext);
        assertThat(variableValueModel.getVal()).isNull();
```

Based on this, what do you think making this field mandatory? That will simplify the implementation because we will handle null checks just at the VariableValue level

eduwercamacaro (approved):

proposals/000-struct-and-structdef.md
Snarr: This makes a lot of sense, I forgot about that.

Snarr: Note: There are several other places in our protobuf where we use optional VariableValues when maybe we should use null VariableValues, we should investigate those:

- VariableDef
- Failure.Content
- ReportTaskRun.Log_Output
- TaskAttempt.Log_Output

eduwercamacaro: Exactly and then we end up handling null checks twice:
- `ReportTaskRun.Log_Output` is null?
- `ReportTaskRun.Log_Output.value` is null?

coltmcnealy-lh: Good point. I agree with this, let's not be like JavaScript which has `undefined` and `null` as separate values.

While I was having my coffee this morning, though, I realized that the Server is more like Java than like GoLang with regards to "nullable" and "not-nullable." 

In Java, if you pass an object, like `Integer`, it has a reference which might be null or might point to something. A primitive type, like `int`, it will never be null, and defaults to zero.

In GoLang, there are things and there are pointers to things. A thing can be a primitive or even a struct, and every type has a "default" value (like zero or `""`). You can choose whether something is required (value) or a pointer, which can be `nil`.

The Server treats everything like a Java object. It can still be null. It somewhat prevents us from being as strong as Go and Rust in terms of type-checking...it makes me a little bit grumpy but Java is overall fine, so it should be okay.

coltmcnealy-lh: Separately, I think I'd like to push for `optional VariableValue default_value = ?;` and getting rid of the `bool optional = 2;` field. Why?

```
{
  optional: false,
  field_type: {...},
  default_value: {
    str: "asdf"
  }
}
```
The above is valid protobuf with that schema. But what does it mean? Someone might be confused by why do we have a default value for a required field.

I would propose the following protobuf:

```
// A `StructFieldDef` defines a field inside a `StructDef`.
message StructFieldDef {
  // The name of the field.
  string name = 1;
 
  // The type of the field.
  TypeDefinition field_type = 3;
  
  // The default value of the field, which should match the Field Type. If not
  // provided, then the field is treated as required.
  optional VariableValue default_value = 4;
}
```

Now, there is only one field which defines what happens in the case that the user does not provide a value. There are three cases:

1. We want to reject the value (fail validation). To do this, do not set the `default_value` field.
2. We want to accept the value (pass validation) and leave the field as null. To do this, set the `default_value` field to the empty `VariableValue`.
3. We want to accept the value and fill the field with a default value. To do this, set the `default_value` field to whatever you want (matching the appropriate type, of course!).

Snarr: I agree with Colt on the fact that a `Required` field should not have a default and I hesitantly agree that `Optional` fields must always have a default. 

I am confused on your 3 cases @coltmcnealy-lh. Are you saying:

- 1. If the optional `default_value` field is not set, we reject the request? Doesn't that defeat the whole purpose of it being optional? I am confused.

Snarr: @coltmcnealy-lh About your proto proposal, you want to store the name twice? I don't think we need to store the name in the `StructFieldDef`

coltmcnealy-lh: Good point. I think originally we had `repeated StructDefField fields`, which means the `StructDefField` necessarily needed a `string name` inside it.

However, given that we're moving it into a `map<string, StructDefField>`, we can (and probably _should_..?) remove the `name` inside the `StructDefField`.

coltmcnealy-lh:
There were offline comments asking about 'why do we need `AllowedStructDefUpdateType`?' I'll answer that here.

The idea is that client code will have utilities to register a `StructDef` using the `@LHStructDef` annotation over classes. Just like the "best practice" is to do `Workflow#reigsterWfSpec()` upon application startup, we will recommend that users register `StructDef`'s upon application startup.

In the modern world, there are often multiple instances of an application, and often there are updates to the code.

Users might want to either:

1. Fail application startup if there is any change to the `StructDef` (and be notified by a failing k8s healthcheck for example), because they don't want to make changes to the schema and if there are any changes then it was a mistake.
2. Allow changes to the schema by re-deploying their code, but ONLY if the changes are compatible.

In the future we may add more options, which is why this is an ENUM.

@Snarr

coltmcnealy-lh (approved):
This looks good. Let's get it merged :smile:

Snarr:
> There were offline comments asking about 'why do we need `AllowedStructDefUpdateType`?' I'll answer that here.
> 
> The idea is that client code will have utilities to register a `StructDef` using the `@LHStructDef` annotation over classes. Just like the "best practice" is to do `Workflow#reigsterWfSpec()` upon application startup, we will recommend that users register `StructDef`'s upon application startup.
> 
> In the modern world, there are often multiple instances of an application, and often there are updates to the code.
> 
> Users might want to either:
> 
> 1. Fail application startup if there is any change to the `StructDef` (and be notified by a failing k8s healthcheck for example), because they don't want to make changes to the schema and if there are any changes then it was a mistake.
> 2. Allow changes to the schema by re-deploying their code, but ONLY if the changes are compatible.
> 
> In the future we may add more options, which is why this is an ENUM.
> 
> @Snarr

So for the average Task Worker just trying to run a task that uses Structs, we `PutStructDef` when each application starts up as a way to ensure that the StructDef was not changed locally and that all Task Workers running agree with the server on the StructDef schema? Do I have this correct?

eduwercamacaro:
I'm wondering when the value for `AllowedStructDefUpdateType` should change in a regular deployment. 
In my head the `FULLY_COMPATIBLE_SCHEMA_UPDATES` value only applies during the first initialization, but once the schema is registered users should change it to `NO_SCHEMA_UPDATES`. am I right?

Snarr:
@coltmcnealy-lh after discussing with @eduwercamacaro, we think it might be more valuable if the `NO_SCHEMA_UPDATES` functionality, which serves to validate that a Schema has not updated, were apart of its own `RPC ValidateStructDef` call, where the request looks like this:

```proto

service LittleHorse {
  // Validate the schema of a new StructDef against an existing StructDef
  rpc ValidateStructDef(ValidateStructDefRequest) returns (google.protobuf.Empty) {}
}

// Request to validate a StructDef.
message ValidateStructDefRequest {
  // The name of the `StructDef` you want to compare against that already exists on the server
  string name = 1;

  // The actual schema for the current `StructDef`.
  InlineStructDef struct_def = 2;


  // The server will validate the new StructDef against
  // the existing version based on this compatibility type.
  //
  // If no compatibility type is specified, then the server will
  // validate if the StructDefs are the same.
  optional StructDefCompatibilityType compatibility_type = 3;
}

message ValidateStructDefResponse {
  bool is_valid = 0;
  
  // Leave room for indicating which fields are invalid?
}

enum StructDefCompatibilityType {
  // Allowed to make fully compatible (both backward-and-forward compatible)
  // changes to the `struct_def` in this request.
  FULL_COMPATIBLE = 0;
}
```

## Rejected Alternative

If you wish not to add another RPC for this behavior, we can just change the name of the `NO_SCHEMA_UPDATES` enum value to something that indicates it is used for validation. e.g. `VALIDATE_NO_UPDATES`.

coltmcnealy-lh:
I'd prefer to not have another rpc for this. It increases the amount of work a user needs in order to achieve the desired result and also has a slight vulnerability to race conditions between validating and registering.

Snarr:
Ok, so keep the functionality of `PutStructDef` and add the `ValidateStructDef` call as well. So `PutStructDef` can still do that `NO_SCHEMA_UPDATES` validation and be smart, but there's also an additional call that supports just doing validations and no puts. Got it 👍

coltmcnealy-lh (approved):

PR #1541, opened 2025-05-22
fix(lhctl): lhctl run should be compatible with old types of variabledefs

coltmcnealy-lh:
A previous fix in this file would fail if you tried to run a `WfSpec` created prior to `0.13.1`.


Snarr (approved):
I think I wrote the previous fix to the file but I wasn't aware that we'd have to support both versions in this release of `lhctl`. Whats the compatibility policy for this proto change? How long will we keep around the old and new methods in our codebases like this?

Issue #1543, opened 2025-05-24
Move Towards Strong Typing


coltmcnealy-lh:
We want to move the LittleHorse DSL towards strong typing as much as possible. This will be a multi-year effort to deprecate the last few places in our API where python-style Chulla Vida typing is allowed.

- [ ] Deprecate the `JSON_OBJ` and `JSON_ARR` types after `StructDef` and before 1.0
- [ ] Remove the `JSON_OBJ` and `JSON_ARR` types in 2.0
- [ ] Remove untyped `ExternalEventDef`s in 2.0
- [ ] Remove `VariableAssignment.NodeOutputReference` from the SDK's before 1.0
- [ ] Remove `VariableAssignment.NodeOutputReference` from the Server in 2.0

coltmcnealy-lh:
I am working on a formal Proposal with the details and will refer to this ticket in it.

PR #1549, opened 2025-05-24
chore(project): adds project lifecycle document

coltmcnealy-lh:


coltmcnealy-lh:
This adds the proposal needed for #1543

PR #1550, opened 2025-05-24
feat(proposals): strong typing deprecations proposal

coltmcnealy-lh:


gwenshap (commented):
Great idea! Strong typing will make it easier to have consistent behavior with fewer edge cases.

Also, it will help LLM-based agents write correct workflows (many rely on linter and type complaints via a language server as a feedback mechanism - so they can automatically fix the code they wrote).

coltmcnealy-lh:
> Great idea! Strong typing will make it easier to have consistent behavior with fewer edge cases.
> 
> Also, it will help LLM-based agents write correct workflows (many rely on linter and type complaints via a language server as a feedback mechanism - so they can automatically fix the code they wrote).

Great point...you need strong typing (with no opt-out) if you want Language Server support.

I'm also working on another _proposal_ at the moment which aims to have a "Child Workflow Node" wherein you can have a `Node` start and wait for another `WfRun` to complete. 

In LittleHorse, every `NodeRun` can return an output, and you can take the output of a `NodeRun` and mutate variables. Right now, we don't do any type-checking on those mutations: it's just Wild West Chulla Vida YOLO Anakin Skywalker style.

I want to make the Server reject any `WfSpec` that could result in runtime exceptions. In order for this to work with these new nodes, for example, we need the output of a `WfRun` to be of a predictable type. If your `WfSpec` just returns a random unspecified field from a JSON variable, how can we do that? Well, we can't. So I want to kill untyped stuff with fire.

proposals/002-move-to-strong-typing.md
```diff
@@ -0,0 +1,161 @@
...
+Valid classes in Java will be:
+* `String.class`
+* `Integer.class` or `Long.class`
+* `Boolean.class`
+* `Double.class` or `Float.class`
+* Any class with the `@LHStructDef` annotation.
+
+Other SDK's will provide equivalent functionality.
+
+We will then make the Server validate all variable mutations on `ExternalEventNode`s and also validate the declarations of the
```
Snarr, line 93: trailing off

proposals/002-move-to-strong-typing.md
```diff
@@ -0,0 +1,161 @@
...
+
+```java
+NodeOutput firstOutput= wf.execute("some-task");
+
+wf.execute("another-task", firstOutput);
+```
+
+The current implementation has some problems:
+
+* **Performance**: it involves sequentially iterating through all previous `NodeRun`'s in the
```
Snarr, line 109: trailing off

proposals/002-move-to-strong-typing.md
```diff
@@ -0,0 +1,161 @@
...
+1. Provide utilities that facilitate and encourage the registration of `ExternalEventDef` and `WorkflowEventDef` resources _with proper typing information, and then deprecate un-typed `ExternalEventDef` and `WorkflowEventDef`'s in LittleHorse 2.0.
+2. Deprecate the `VariableAssignment.NodeOutputReference` in 1.0 and remove it in 2.0.
+3. Deprecate `JSON_OBJ` and `JSON_ARR` when `StructDef` is ready (before 1.0), and remove it in 2.0.
+
+### Deprecating Untyped `ExternalEventDef`s
+
+We will first make it possible to register `ExternalEventDef`'s using our `Workflow` DSL. This utility will _only_ support strongly-typed `ExternalEventDef`'s, and will not support `JSON_OBJ` `ExternalEventDef`'s. It will look as follows:
+
+```java
+wf.waitForEvent("my-event").registeredAs(String.class);
```
Snarr, line 81: I am still skeptical about this design and I am open minded about alternative implementations. But I do not have any alternatives at the moment, just putting this out there.

Snarr (approved):
Approved with comments

I agree Strong Typing is important to the future LittleHorse, and with LittleHorse being built on Java it would be quite silly of us to go with a YOLO typing strategy.

proposals/002-move-to-strong-typing.md
coltmcnealy-lh: Hmm. The idea is that if you _don't_ use `.registeredAs()`, then you must manually create the `ExternalEventDef` before registering the `WfSpec`. The point of that is because there are use-cases for a single `ExternalEventDef` (eg. `document-signed`) to be used within multiple `WfSpec`s, so not all of the `WfSpec`'s should/can register it.

> I think I understand why this is necessary—to support old WfSpec code

I might be misunderstanding you here, but the purpose isn't to support old `WfSpec` code so much as to give users the option to register `ExternalEventDef`'s outside the `WfSpec`, which we will always continue to support.

Snarr: > I might be misunderstanding you here, but the purpose isn't to support old WfSpec code so much as to give users the option to register ExternalEventDef's outside the WfSpec, which we will always continue to support.

Ahhh okay that makes sense. I understand now that this will not be the only strategy we provide in our SDK for registering `ExternalEventDefs`, it will just be an option for doing it in the DSL.

eduwercamacaro (approved):

PR #1552, opened 2025-05-26
fix(server): fixes ineffective external event e2e test

coltmcnealy-lh:


coltmcnealy-lh:
Thanks @KarlaCarvajal !

PR #1555, opened 2025-05-27
fix(e2e): allow 500 ms between task run retry and retry in flaky test

KarlaCarvajal:


Snarr (approved):
Good fix, thanks for taking the time to analyze this and figure it out!

PR #1559, opened 2025-05-29
fix(sdk-dotnet): add validations to handle the cases when new nodes are added after a complete.

KarlaCarvajal:
Add validations to handle the cases when new nodes are added after a complete.


Snarr (approved):

sdk-dotnet/LittleHorse.Sdk.Tests/Workflow/Spec/WorkflowThreadTest.cs
```diff
@@ -1547,4 +1547,25 @@ void EntryPointAction(WorkflowThread wf)
...
+    {
+        var workflowName = "TestWorkflow";
+        var mockParentWorkflow = new Mock<Sdk.Workflow.Spec.Workflow>(workflowName, _action);
+
+        void EntryPointAction(WorkflowThread wf)
+        {
+            var input = wf.DeclareStr("input");
+            wf.Execute("any-task-def-name", input);
+            wf.Complete();
+            var exception = Assert.Throws<InvalidOperationException>(() =>
```
Snarr: Can we move the `Assert.Throws` outside of the WorkflowThread and validate it on the `new WorkflowThread()` call down below? Just to organize the test better. I don't know if this is possible in C# so let me know.

Snarr (approved):

sdk-dotnet/LittleHorse.Sdk.Tests/Workflow/Spec/WorkflowThreadTest.cs
KarlaCarvajal: thanks for the feedback, now these tests are more readable!!!

eduwercamacaro (approved):

PR #1561, opened 2025-05-30
docs(proposal): proposal for userComments on a UserTaskRun

JakeRose03:


proposals/003-add-user-task-comments.md
```diff
@@ -0,0 +1,74 @@
...
+message Command {
+    // ...
+
+    oneof command {
+        // ...
+        CommentUserTaskRunRequest comment_user_task_run = 28;
+    }
+}
+
+```
```
coltmcnealy-lh, line 236: Looks good. Maybe add the updates to the `message UserTaskRun`, and comment about how `rpc GetUserTaskRun` is sufficient to fetch comments?

KarlaCarvajal: Hi @coltmcnealy-lh, I think the userTaskRun will only handle a new event (commented), what do you mean by adding the updates there?

proposals/003-add-user-task-comments.md
```diff
@@ -0,0 +1,257 @@
...
+
+  // This is the id of the user 
+  string user_id = 2;
+
+  // This is the specific note/comment that a user wants to make 
+  // on a `UserTask`
+  string comment = 3;
+
+  // Flag representing if this comment has been edited 
+  boolean isEdited = 4;
```
Snarr: Don't forget to use `snake_case` in protobuf field names

proposals/003-add-user-task-comments.md
```diff
@@ -0,0 +1,257 @@
...
+  UserTaskRunId user_task_run_id = 1;
+
+  // The author of the comment being made on a UserTaskRun
+  string user_id = 2;
+
+  // The comment being made on a UserTaskRun
+  string comment = 3;
+
+  // Flag that represent whether a field has been edited
+  boolean isEdited = 4 ;
```
Snarr: Are we still incorporating editing functionality?

proposals/003-add-user-task-comments.md
```diff
@@ -0,0 +1,257 @@
...
+  string user_comment_id = 1;
+}
+
+```
+
+```proto
+  oneof event {
+    // ...
+    UTEComment commented = 6;
+    UTEDeleteComment comment_deleted = 7;
```
Snarr: How would UTEDeleteComment work under the hood? Would the UTEComment event actually be deleted from the UserTaskEvent list in the parent UserTaskRun? I see an example down below but I don't think that functionality is explicitly mentioned anywhere.

JakeRose03: I do not believe the event should be deleted , we can add a flag for hasBeenDeleted so frontend knows what  comments to display.

proposals/003-add-user-task-comments.md
JakeRose03: Yes, I believe this is the desired functionality

Snarr: Which RPC would enable this editing functionality in Option 1?

proposals/003-add-user-task-comments.md
Snarr: I don't have a strong opinion about this, I can see pros and cons for both deleting the event and not deleting the event, lets just make sure that we document `RPC DeleteCommentUserTaskRun` with comments about this behavior, otherwise users might be confused.

proposals/003-add-user-task-comments.md
JakeRose03: Option 1 has limited functionality. You would only be able  to append  new comments and then list comments

Snarr: Wait so how is it getting edited? What is the point of the `isEdited` field in Option 1? I might be missing something here, under what conditions would the `isEdited` field be true, and what conditions would make it false?

JakeRose03: My apologies it would be through the `CommentUserTaskRun` if the id is passed in it would overwrite the existing comment

KarlaCarvajal: @Snarr 
The same CommentUserTaskRun method will allow to insert or edit a comment through the optional field `user_comment_id`, if this come from the request, it should update the comment with the `is_edited` flag in true, and if it do not, it should insert a commnet with the desault flag in false.
 The other alternative is by creating a new event to edit comments, what do you think is the best option for you?

Snarr: @KarlaCarvajal Ok nice I see now that the latest changes include the `user_comment_id` field in the `CommentUserTaskRunRequest`. Thank you both for the clarification

KarlaCarvajal: yes, thanks for asking @Snarr. This made us detect some improvements

proposals/003-add-user-task-comments.md
```diff
@@ -0,0 +1,278 @@
...
+    }
+  ]
+}
+```
+
+`UserTaskRunEvents` should be append only. In order to sort UserComments the front end will have to lookup via `userCommentId` and sort by time. The most recent `UserTaskRunEvents`(UTEComment or UTECommentDeleted) will represent whether the comment has been inserted, edited, or deleted. 
+
+## Rejected Options
+### Option 2: Add comments through a list of new objects in the UserTaskRun
+It is possible to add comments through a new object which can be added as a repeated field in the `UserTaskRun` object.
```
Snarr, line 199: Could you add a short blurb about why we rejected this option? It's nice to have a full history of our decisions and some of that discussion happened offline

proposals/003-add-user-task-comments.md
```diff
@@ -0,0 +1,278 @@
...
+  // The author of the comment being made on a UserTaskRun
+  string user_id = 2;
+
+  // The comment being made on a UserTaskRun
+  string comment = 3;
+}
+
+message EditCommentUserTaskRunRequest {
+  // This the id of a individual user comment
+  int16 user_comment_id = 4 ;
```
Snarr: Small nitpick but you can set this field name to 1 and renumber the other field names if you want to preserve the ordering in the written format

proposals/003-add-user-task-comments.md
KarlaCarvajal: Hi @Snarr, we added the reason, could you check this PR one more time please?

Snarr: Looks great 👍

proposals/003-add-user-task-comments.md
KarlaCarvajal: @coltmcnealy-lh `comment_id_count` field was added to the `UserTaskRun`, this allows to start comment_ids from 0 every time a `UserTaskRun` is added.

eduwercamacaro (approved):

PR #1569, opened 2025-06-05
feat(server): add comment events to userTaskRun

KarlaCarvajal:
- Add new event to comment usert task runs in proto files
- Add commands to allow to map comment, editComment and deleteComment in UserTaskRuns
- Add models for comments' requests and events in server code
- Add a new listener to comment userTaskRuns
- Add e2e to verify comment behaviour in UserTaskRun

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/DeleteUserTaskRunCommentRequestModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/DeleteUserTaskRunCommentRequestModel.java
@@ -22,7 +22,6 @@ public class DeleteUserTaskRunCommentRequestModel extends CoreSubCommand<DeleteU
     private UserTaskRunIdModel userTaskRunId;
     private String userId;
 
-
     @Override
     public UserTaskRun process(ProcessorExecutionContext executionContext, LHServerConfig config) {
 
--- a/server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/EditUserTaskRunCommentRequestModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/EditUserTaskRunCommentRequestModel.java
@@ -23,8 +23,6 @@ public class EditUserTaskRunCommentRequestModel extends CoreSubCommand<EditUserT
     private String userId;
     private String comment;
 
-  
-
     @Override
     public UserTaskRun process(ProcessorExecutionContext executionContext, LHServerConfig config) {
         if (userCommentId == 0) {
--- a/server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/PutUserTaskRunCommentReqeustModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/PutUserTaskRunCommentReqeustModel.java
@@ -21,7 +21,6 @@ public class PutUserTaskRunCommentReqeustModel extends CoreSubCommand<PutUserTas
     private String userId;
     private String comment;
 
-
     @Override
     public UserTaskRun process(ProcessorExecutionContext executionContext, LHServerConfig config) {
         if (userId.isBlank()) {


schemas/internal/command.proto
```diff
@@ -48,6 +48,9 @@ message Command {
         SaveUserTaskRunProgressRequest save_user_task_run_progress = 27;
         PutCorrelatedEventRequest put_correlated_event = 28;
         UpdateCorrelationmarkerPb update_correlation_marker = 29;
+        CommentUserTaskRunRequest comment_user_task_run = 30;
+        EditCommentUserTaskRunRequest edit_comment_user_task_run = 31;
```
coltmcnealy-lh: This is an annoying request, but I think the request would sound better if named as `EditUserTaskRunCommentRequest edit_user_task_run_comment` and `DeleteUserTaskRunCommentRequest delete_user_task_run_comment`.

The `CommentUserTaskRun` makes sense. What I'm trying to do is stick to the pattern of `{Verb}{Noun}Request`.

The way it's currently written makes it seem like:

* verb: `Edit`
* noun: `CommentUserTaskRunRequest`

Which sounds like we're deleting the request; whereas in reality we want to delete the comment.

schemas/littlehorse/service.proto
```diff
@@ -145,6 +145,15 @@ service LittleHorse {
   // to model an entity.
   rpc ListUserTaskRuns(ListUserTaskRunRequest) returns (UserTaskRunList) {}
 
+  // Adds userComment to a UserTaskRun
+  rpc CommentUserTaskRun(CommentUserTaskRunRequest) returns (UserTaskEvent) {};
+
+  // Edits userComment with the correlated userCommentId 
+  rpc EditCommentUserTaskRun(EditCommentUserTaskRunRequest) returns (UserTaskEvent) {};
```
coltmcnealy-lh:
```suggestion
  rpc EditUserTaskRunComment(EditUserTaskRunCommentRequest) returns (UserTaskRun) {};
```

coltmcnealy-lh: I think in order to retain consistency with the rest of the LittleHorse GRPC API, return the full `UserTaskRun` object, not the event itself. That's because we currently don't return things which aren't Getable's with their own ID's.

schemas/littlehorse/service.proto
```diff
@@ -145,6 +145,15 @@ service LittleHorse {
   // to model an entity.
   rpc ListUserTaskRuns(ListUserTaskRunRequest) returns (UserTaskRunList) {}
 
+  // Adds userComment to a UserTaskRun
+  rpc CommentUserTaskRun(CommentUserTaskRunRequest) returns (UserTaskEvent) {};
```
coltmcnealy-lh:
```suggestion
  rpc CommentUserTaskRun(CommentUserTaskRunRequest) returns (UserTaskRun) {};
```

schemas/littlehorse/service.proto
```diff
@@ -145,6 +145,15 @@ service LittleHorse {
...
+
+  // Edits userComment with the correlated userCommentId 
+  rpc EditCommentUserTaskRun(EditCommentUserTaskRunRequest) returns (UserTaskEvent) {};
+
+  // Deletes a comment logically, this does not affect the userTaskEvent Log
+  rpc DeleteCommentUserTaskRun(DeleteCommentUserTaskRunRequest) returns (google.protobuf.Empty) {};
```
coltmcnealy-lh:
```suggestion
  rpc DeleteCommentUserTaskRun(DeleteUserTaskRunCommentRequest) returns (UserTaskRun) {};
```

This suggestion is a bit more debatable. When we delete a `Getable`, we currently return `Empty` (but I want to start returning the `Getable` not just `Empty`). 

I think here we can return the actual `Getable` without breaking the semantics of the rest of the API. I also think we maybe should make that full change before 1.0

schemas/littlehorse/user_tasks.proto
```diff
@@ -117,6 +117,9 @@ message UserTaskRun {
   // updated or re-assigned. Used internally to implement automated reassignment and reminder
   // tasks.
   int32 epoch = 12;
+
+  // The number of comments that have been made on this UserTaskRun.
+  int32 comment_id_count = 13;
```
coltmcnealy-lh: I don't believe we need this. You can infer it by walking the list.

server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/CommentUserTaskRunRequestModel.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public boolean hasResponse() {
+        return true;
+    }
+
+    @Override
+    public UserTaskEvent process(ProcessorExecutionContext executionContext, LHServerConfig config) {
+        if (userTaskRunId.getUserTaskGuid().isBlank()
+                || userTaskRunId.getWfRunId().getId().isBlank()) {
```
coltmcnealy-lh: I think the validation of `if (utr == null)` a few paragraphs below is perfectly sufficient here—an empty-string is _technically_ a valid `WfRunId` as of now.

server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/CommentUserTaskRunRequestModel.java
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        UserTaskEventModel userTaskEventModel = utr.comment(userId, comment);
+
+        WfRunModel wfRunModel = executionContext.getableManager().get(userTaskRunId.getWfRunId());
+        if (wfRunModel == null) {
+            throw new LHApiException(Status.DATA_LOSS, "Impossible: got UserTaskRun but missing WfRun");
+        }
+
+        wfRunModel.advance(new Date());
```
coltmcnealy-lh: This call is somewhat expensive, and I think we actually don't need to make this call here, since commenting a `UserTaskRun` won't cause a workflow to advance at all.

server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/DeleteCommentUserTaskRunRequestModel.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public Message process(ProcessorExecutionContext executionContext, LHServerConfig config) {
+
+        if (userCommentId == 0) {
+            throw new LHApiException(Status.FAILED_PRECONDITION, "The User Comment Id must be provided");
+        }
+        if (userTaskRunId.getUserTaskGuid().isBlank()
+                || userTaskRunId.getWfRunId().getId().isBlank()) {
+            throw new LHApiException(Status.INVALID_ARGUMENT, "The userTaskRunId must be provided.");
```
coltmcnealy-lh: Same comments here, we can just throw `NOT_FOUND`. An empty string is valid.

server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/DeleteCommentUserTaskRunRequestModel.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        utr.deleteComment(userCommentId);
+
+        WfRunModel wfRunModel = executionContext.getableManager().get(userTaskRunId.getWfRunId());
+        if (wfRunModel == null) {
+            throw new LHApiException(Status.DATA_LOSS, "Impossible: got UserTaskRun but missing WfRun");
+        }
+
+        wfRunModel.advance(new Date());
```
coltmcnealy-lh: Likewise, advance() not needed.

server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/CommentUserTaskRunRequestModel.java
KarlaCarvajal: really?, we did not know that it was possible to have empty WfRunIds. We will do a test with a wfRun with emptyId and change that validation. Thanks!

schemas/littlehorse/user_tasks.proto
```diff
@@ -189,8 +190,45 @@ message SaveUserTaskRunProgressRequest {
...
+
+  // The author of the comment being made on a `UserTaskRun`
+  string user_id = 3;
+
+  // The comment being made on a `UserTaskRun`
+  string comment = 4;
+}
+
+// Deletes a comment on a UserTaskRun
+message DeleteUserTaskRunCommentRequest {
```
Snarr, line 225: Maybe we should store the ID of the user who deleted the comment?

JakeRose03: Yes I agree with this.

schemas/littlehorse/user_tasks.proto
```diff
@@ -286,5 +347,14 @@ message UserTaskEvent {
 
     // Denotes that the `UserTaskRun` was saved.
     UTESaved saved = 5;
+
+    // Denotes that there was a comment on a `userTaskRun`
+    UTECommented comment_added = 6;
+
+    // Denotes that a comment on a `userTaskRun` has been edited
+    UTECommented comment_edited = 7;
```
coltmcnealy-lh, line 355: It's clever that you can re-use the same proto structure here.

coltmcnealy-lh (approved):
Great updates!

PR #1570, opened 2025-06-06
docs(proposal): Add `StructDef` naming conventions proposal

Snarr:


proposals/004-structdef-field-names.md
```diff
@@ -0,0 +1,226 @@
...
+
+#### Standard
+
+I propose that we follow the lower `camelCase` convention for `StructDef` Field names, meaning:
+
+- The field name must start with a lowercase letter
+  - Why no numbers? Java, Python, Go, class attributes cannot start with numbers
+- Subsequent characters must match the ranges `A-Z`, `A-Z`, and `0-9`
+  - Why no underscores? We want Java, .NET, and Go users to design compatible fields without breaking their individual language naming conventions. Since underscores are typically only used in constants in Java, and private member variables in `C#`, this character breaks support for these languages. Sorry Python, you can keep your `snake_case`, we will just convert them to `camelCase` during the `StructDef` compilation process.
+- For our case, field names will be `case insensitive`.
```
coltmcnealy-lh: Treating `thisField` and `thisfield` as the same is probably not the best UX. Also note that it's very easy to write a parser that converts `python_snake_case` into `properCamelCase`—ChatGPT should be able to work around it.

coltmcnealy-lh (approved):
Looks good except two things:

1. Let's be case sensitive if we are using `camelCase`.
2. Make a note that `_` is a somewhat special character in RocksDB

proposals/004-structdef-field-names.md
Snarr: Snake to camelCase will be more straight forward and easier for users to follow when porting between SDKs, I agree. My concern about case sensitivity does not have much to do with this.

See this example for why I don't want our fields to be case sensitive:

Java `carVINNumber` -> Converts to  `carVINNumber`
Python: `car_vin_number` -> Converts to `carVinNumber`

Doing conversion magic involving capitalization can lead to unpredictable effects, which is why I am in favor of making the fields case insensitive.

eduwercamacaro (approved):

PR #1571, opened 2025-06-06
Completable futures refactor

eduwercamacaro:
Major refactor for Command waiters:
- Grpc Listeners implement a Thread-per-Request approach
- Uses CompletableFutures as the mechanism to wait for commands to be executed

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/corecommand/CommandModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/corecommand/CommandModel.java
@@ -37,7 +37,7 @@ import io.littlehorse.common.util.LHUtil;
 import io.littlehorse.server.streams.topology.core.ExecutionContext;
 import io.littlehorse.server.streams.topology.core.ProcessorExecutionContext;
 import java.util.Date;
-import java.util.Objects;
+import java.util.Optional;
 import lombok.Getter;
 import lombok.Setter;
 
@@ -451,8 +451,9 @@ public class CommandModel extends AbstractCommand<Command> {
         return config.getCoreCmdTopicName();
     }
 
-    public String getCommandId() {
-        return Objects.requireNonNull(commandId);
+    @Override
+    public Optional<String> getCommandId() {
+        return super.getCommandId();
     }
 
     public boolean hasResponse() {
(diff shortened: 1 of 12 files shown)


server/src/main/java/io/littlehorse/common/model/corecommand/CommandModel.java
```diff
@@ -451,6 +451,10 @@ public String getTopic(LHServerConfig config) {
         return config.getCoreCmdTopicName();
     }
 
+    public String getCommandId() {
+        return Objects.requireNonNull(commandId);
```
coltmcnealy-lh: In certain cases, we rely on the `commandId` being null to determine whether or not to notify the Async Waiters that someone is waiting for a response. This would break that. Can we suppoort nulls?

server/src/main/java/io/littlehorse/server/LHInternalClient.java
```diff
@@ -0,0 +1,100 @@
...
+            ByteString bytes = response.getResult();
+
+            return (T) responseCls.getMethod("parseFrom", ByteString.class).invoke(null, bytes);
+        } catch (Exception exn) {
+            log.error(exn.getMessage(), exn);
+            throw new RuntimeException("Not possible");
+        }
+    }
+
+    private LHInternalsGrpc.LHInternalsBlockingStub getInternalAsyncClient(
```
coltmcnealy-lh, line 72: Why is this method called `getInternalAsyncClient()` if it returns a blocking stub?

server/src/main/java/io/littlehorse/server/streams/topology/core/processors/CommandProcessor.java
```diff
@@ -93,19 +96,17 @@ private void processHelper(final Record<String, Command> commandRecord) {
                 "{} Processing command of type {} with commandId {} with partition key {}",
                 config.getLHInstanceName(),
                 command.type,
-                command.commandId,
+                command.getCommandId(),
                 command.getPartitionKey());
         try {
             Message response = command.process(executionContext, config);
             executionContext.endExecution();
+            CompletableFuture<Message> completable =
```
coltmcnealy-lh: Please guard this inside the `if ()`. We shouldn't be creating `CompletableFuture`s if there is nothing waiting for a response

sauljabin (approved):

PR #1595, opened 2025-06-09
feat(server): Add Experimental Feature Flag for `StructDef`s

Snarr:
Adds a feature flag `LHS_X_ENABLE_STRUCT_DEFS` to the environment variables in `LHServerConfig` that takes in a true/false value to enable StructDef functionality in the Server. This will prevent clients from relying or trying to build upon the `StructDef` RPCs while they are in development.


server/src/test/java/e2e/StandaloneTestBootstrapper.java
```diff
@@ -101,6 +101,7 @@ private Properties getBaseServerConfig() {
         serverProperties.put(LHServerConfig.CORE_MEMTABLE_SIZE_BYTES_KEY, String.valueOf(1024L * 1024L * 8));
         serverProperties.put(LHServerConfig.ROCKSDB_TOTAL_MEMTABLE_BYTES_KEY, String.valueOf(1024L * 1024L * 100));
         serverProperties.put(LHServerConfig.ROCKSDB_TOTAL_BLOCK_CACHE_BYTES_KEY, String.valueOf(1024L * 1024L * 100));
+        serverProperties.put(LHServerConfig.X_ENABLE_STRUCT_DEFS_KEY, "true");
```
coltmcnealy-lh, line 104: This enables them by default. Is that the intended behavior?

Snarr: :O Isn't this just for tests? My intention was to enable them for the e2e testing framework so our tests would still validate and pass.

PR #1600, opened 2025-06-10
fix(dotnet-sdk): Downgrades examples from .NET 8 to .NET 6

MateoRojas:
The SDK and the tests are using .NET 6, and even though .NET 8 should be compatible with .NET, there .NET Unit Test framework insist that test should be run with .NET 6. So to keep a single version of .NET when developing it's better if everything uses .NET 6 for now.


KarlaCarvajal:
For real life examples is better to have them running in .Net8 as it is the most used version. I will recommend to keep .net version 8.0 as it is for external people usage.

eduwercamacaro (approved):
LGTM! thanks

PR #1604, opened 2025-06-11
feat(server): allow using direct io for rocksdb

coltmcnealy-lh:
In certain resource-constrained configurations in our soak tests, we have seen runaway page cache usage. In containerized environments we cannot limit page-cache memory (either via the application or via the container configuration). We can only limit total memory.

If page cache usage grows unbounded then we can encounter OOMKilled (137) as the container runtime enforces the overall memory limit on the container.

This commit introduces LHS_ROCKSDB_USE_DIRECT_IO which allows users to specify whether to bypass the page cache in rocksdb, relying instead on the Block Cache for performance. The Block Cache serves an equivalent purpose to the page cache except that it allows users to control the size via LHS_ROCKSDB_TOTAL_BLOCK_CACHE_BYTES.


coltmcnealy-lh:
After stopping the load and waiting 5 minutes, the page cache still remained at this level:
image


I will now test this with `LHS_ROCKSDB_USE_DIRECT_IO=true`. Identical setup to before.

coltmcnealy-lh:
With `LHS_ROCKSDB_USE_DIRECT_IO=true`, I ran the test for 10 minutes (as opposed to 8 minutes), and the memory usage never got above 515MB (as opposed to almost 800MB). 

Most importantly, the page cache usage topped out at `16605184`, which was 1/82nd as much as `1373495296`

image

Safe to say that this configuration solves the problem we observed in our soaks. 

QED

Issue #904, opened 2024-06-27
Add `CompleteUserTask` to Audit Log


Snarr:
In `user_tasks.proto`, we have a `message UserTaskEvent` that is stored in the audit log of a `UserTaskRun` for observability purposes. There are different types of `UserTaskEvent`s that can be logged, and we should add `SaveUserTask` and `CompleteUserTask` to this list of types once they are ready.

From `schemas/littlehorse/user_tasks.proto`:

```
message UserTaskEvent {
  ...
  oneof event {
    // Denotes that a TaskRun was scheduled via a trigger.
    UTETaskExecuted task_executed = 2;
  
    // Denotes that the UserTaskRun was assigned.
    UTEAssigned assigned = 3;
  
    // Denotes that the UserTaskRun was cancelled.
    UTECancelled cancelled = 4;
    // TODO: Add "save user task" and "complete user task" to the
    // audit log
  }
}
```

Snarr:
Partially resolved: https://github.com/littlehorse-enterprises/littlehorse/pull/946

PR #1607, opened 2025-06-11
docs(proposal): add completed UserTaskEvent

JakeRose03:
## Summary

This PR adds Proposal 005: `userTaskEvent` completed.
.

## Changes

- Adds Proposal 005 markdown file (`proposals/005-user-task-completed.md`)
- Updates `proposals/README.md` with Proposal 005 entry

## Related

Related to Issue #904


proposals/005-add-user-task-event-completed.md
```diff
@@ -0,0 +1,39 @@
...
+
+Add new message:
+```proto
+message UserTaskEvent {
+  //...
+
+  // Message denoting that the UserTaskRun has been completed
+  message UTECompleted {
+    // The ID of the TaskRun that was just completed
+    TaskRunId task_run_id = 1;
```
coltmcnealy-lh: Why is this a `TaskRunId`?

JakeRose03: ah, well I guess I meant `userTaskRunId` but it should just be a string correct?

coltmcnealy-lh: A `UserTaskRunId` is also a top-level protobuf object.

However, I don't see the need to put the `UserTaskRunId` in this messsage, because it is inherent in the `UserTaskRun` itself (of which the `UTECompleted` will be a part).

I am trying to think of what other information we can/should put in `message UTECompleted`. I can't think of too much:

* `user_id` is frozen after completion, so the user who completed it is already evident.
* The results are also frozen, so we can just look into the `results` to see what the payload was.
* The timestamp is already stored at the `UTEvent` level.

This means I think the `UTECompleted` could just be an empty protobuf for now. Unless @eduwercamacaro or @Snarr  you have other ideas for what can go in it?

PR #1613, opened 2025-06-12
Handle thread input variables for output topic

eduwercamacaro:


coltmcnealy-lh (changes requested):
On second thought, why should we not allow these methods to be produced?

Separately, I am working on tests for the output topic on another branch. If this isn't blocking / breaking anything, can you wait until tomorrow mid-morning?

coltmcnealy-lh:
I believe we should be able to put `INPUT` variables in the output topic. #1614 fixes this problem and allows that.

PR #1614, opened 2025-06-13
fix(server): output topic and handler input variables

coltmcnealy-lh:
- Fixes loading of variable types in case of exception handler / interrupt handler `INPUT` variables
- Fixes output topic handling of `StructDefModel`
- Avoids calling `rpc PutTenant` every test instance.


server/src/main/java/io/littlehorse/common/model/getable/core/variable/VariableModel.java
```diff
@@ -163,14 +166,21 @@ private boolean isIndexable() {
...
+            ReadOnlyMetadataManager metadataManager,
+            ReadOnlyGetableManager getableManager,
+            OutputTopicConfigModel config) {
         if (config.getDefaultRecordingLevel() == OutputTopicRecordingLevel.NO_ENTITY_EVENTS) {
             return false;
         }
 
         // Only PUBLIC_VAR variables should be pushed out.
-        ThreadVarDefModel variableDef =
-                getWfSpec(metadataManager).getAllVariables().get(id.getName());
```
eduwercamacaro, line 173: So, as far as I understand the `getAllVariables` filtered `INPUT` variables that's why we were receiving a NPE at this point. But, I have a question. Is there a valid use case to produce thread `INPUT` variables to the output topic? I think these type of variables are meaningless for a business use case as they server technical purposes within a wfspec.

test-utils/src/main/java/io/littlehorse/test/LHExtension.java
```diff
@@ -226,18 +228,26 @@ private void maybeCreateTenantAndPrincipal(TestContext testContext) {
...
-                            .getLhClient()
-                            .withDeadlineAfter(2, TimeUnit.SECONDS)
-                            .putTenant(PutTenantRequest.newBuilder()
-                                    .setId(testContext.getConfig().getTenantId().getId())
-                                    .build());
-                    return true;
-                });
+        try {
+            testContext.getLhClient().getTenant(testContext.getConfig().getTenantId());
+        } catch (StatusRuntimeException sre) {
```
eduwercamacaro: I believe we should try to register the tenant only when `rpc GetTenant` returns not found, so we can handle cases where the server is not ready and/or healthy yet.

server/src/main/java/io/littlehorse/common/model/getable/core/variable/VariableModel.java
coltmcnealy-lh: There are use-cases, for example catching exceptions with meaningful output, or interrupt handlers that have meaningful content.

test-utils/src/main/java/io/littlehorse/test/LHExtension.java
coltmcnealy-lh: Currently (before this commit) the test framework relies on the `rpc PutTenant` to wait until the server is ready.

That behavior is unchanged.

test-utils/src/main/java/io/littlehorse/test/LHExtension.java
```diff
@@ -226,18 +228,26 @@ private void maybeCreateTenantAndPrincipal(TestContext testContext) {
...
+            Awaitility.await()
+                    .atMost(Duration.ofSeconds(25))
```
Snarr: 25 seconds seems extreme, it should fail sooner right?

schemas/littlehorse/output_topic.proto
```diff
@@ -65,5 +66,8 @@ message MetadataOutputTopicRecord {
 
         // A UserTaskDef update
         UserTaskDef user_task_def = 5;
+
+        // A StructDef update
+        StructDef struct_def = 6;
```
Snarr, line 71: Nice

test-utils/src/main/java/io/littlehorse/test/LHExtension.java
coltmcnealy-lh: This code was originally intended to wait for the server + kafka to start up when you first start the e2e tests. It still does that.

However, what I'm fixing in this PR is that the `rpc PutTenant` actaully happens on every test class instance. I found this out because of output topic creation logging.

test-utils/src/main/java/io/littlehorse/test/LHExtension.java
eduwercamacaro: Yes, that behavior is unchanged, but I think we should change that behavior and try to put the tenant only when we receive `NOT_FOUND`.

sauljabin: this catch is not needed, you can use Awaitility

sauljabin (approved):
no special comments from my side

test-utils/src/main/java/io/littlehorse/test/LHExtension.java
coltmcnealy-lh: Fixing...

PR #1615, opened 2025-06-13
fix(server): don't log status runtime exceptions

coltmcnealy-lh:


server/src/main/java/io/littlehorse/server/LHServerListener.java
```diff
@@ -1131,16 +1131,21 @@ private <AC extends Message, RC extends Message> void processCommand(
                     futureResponse.get(LHConstants.MAX_INCOMING_REQUEST_IDLE_TIME.getSeconds(), TimeUnit.SECONDS);
             responseObserver.onNext((RC) response);
             responseObserver.onCompleted();
-        } catch (InterruptedException | ExecutionException e) {
+        } catch (InterruptedException e) {
             Throwable cause = e.getCause() == null ? e : e.getCause();
```
eduwercamacaro: I believe that we don't need to verify the exception's cause here. If we ever get a `InterruptedException` in this class it means that the application is being shutdown. In this case, I believe we response with a `UNAVAILABLE` status code. wdyt?

coltmcnealy-lh: Makes sense, I think that's a good improvement over what we had before

PR #1619, opened 2025-06-13
fix(lhctl): Throw error if `search wfRun` called with no args 

Snarr:
Before this, `lhctl` was throwing a panic when you called `lhctl search wfRun` with zero arguments, which isn't very helpful. This updates the args validator to throw an error message in that case.


coltmcnealy-lh:
Love it. Pad them GH stats!

Issue #1592, opened 2025-06-09
Update `PutExternalEventModel#hasResponse()` to return `false` when appropriate


coltmcnealy-lh:
We should devise a way for the `PutExternalEventModel#hasResponse()` to return `false` for `PutExternalEventRequest`s that did not originate from the grpc server.

I'm leaning towards using Kafka message headers on the `Command` so that we do not pollute the grpc API.

Perhaps the `processCommand()` in the server can automatically set those headers...

coltmcnealy-lh:
Finished by #1620

PR #1620, opened 2025-06-14
refactor(server): remove hasResponse from abstractcommand

coltmcnealy-lh:
- Rely upon presence of commandId to determine whether to create a Completable Future


coltmcnealy-lh:
Fixes #1592

PR #1630, opened 2025-06-18
feat(lhctl): commands for put, edit, and deleting comments on userTaskRun

JakeRose03:
### Summary
implements put edit and delete functionality to lhctl for 
comments on a userTaskRun

### Concerns
My only concern here is that lhctl read me states that flag use is for modifying the underlying behavior 
of the command.
In this implementation, since all fields are required and do not modify behavior of command they should be args

My concern is with unintended behavior. A client can easily input the comment in the userId vice versa.
If they make this mistake they can still edit a comment. but will not be able to remove the event from the audit log .

### DIFF
--- a/lhctl/internal/user_task_run.go
+++ b/lhctl/internal/user_task_run.go
@@ -238,7 +238,7 @@ This command allows you to attach feedback or notes to a specific UserTaskRun fo
 	Run: func(cmd *cobra.Command, args []string) {
 
 		userId, err := promptFor(
-			"Enter the userId of the person completing the task",
+			"Enter the userId of the person adding the comment",
 			lhproto.VariableType_STR,
 		)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


Snarr (approved):
I understand your concerns about users mixing up `userId` and user `comment`s, since `userId`s do not have any validation on the server side to prove that they are valid from what I understand.

I still don't think an option flag makes sense here still, since those usually represent optional data and we want to keep things consistent to our style guide.

If you think that this would be very confusing to users, maybe it would be best to prompt the user from the command line after they enter the `wfRunId` and `userTaskGuid`:

```
$ lhctl put userTaskRunComment <wfRunId> <userTaskGuid>
> Please type the user ID:
> __________
> Please type your comment:
> ________
```

We have a `promptFor` method in `user_task_run.go` that does this exactly when users are executing user tasks.

Through this alternative solution, we would keep our args/flags rules consistent while making it difficult for the user to mix up those two fields.

PR #1648, opened 2025-06-23
feat(dashboard): nextjs metrics

hazimoarafa:
image


dashboard/src/instrumentation-node.ts
```diff
@@ -0,0 +1,78 @@
...
+  ATTR_SERVICE_NAME,
+  ATTR_SERVICE_VERSION,
+} from "@opentelemetry/semantic-conventions";
+import { IncomingMessage } from "http";
+
+// manually setting a context manager to replace the no-op context manager
+context.setGlobalContextManager(new AsyncLocalStorageContextManager());
+
+const exporter = new PrometheusExporter({
+  port: 9464,
```
mijailr: Can this port passed as an environment variable? You can set a default value but some times we may need to change the port.

dashboard/src/instrumentation-node.ts
```diff
@@ -0,0 +1,78 @@
...
+});
+
+const detectedResources = detectResources({
+  detectors: [envDetector, processDetector, hostDetector],
+});
+
+const customResources = resourceFromAttributes({
+  [ATTR_SERVICE_NAME]: "Dashboard",
+  [ATTR_SERVICE_VERSION]: "0.0.0",
```
mijailr, line 36: Can this be set dynamically based on the dashboard version?

dashboard/src/instrumentation.ts
```diff
@@ -0,0 +1,5 @@
+export const register = async () => {
+  if (process.env.NEXT_RUNTIME === "nodejs") {
+    await import("./instrumentation-node");
+  }
```
mijailr: Not only if runtime, but also we may disable it entirely. Use an environment variable to opt to enable/disable.

dashboard/src/instrumentation-node.ts
GhostArrays: When we add dashboard version, there are some techniques that I want to explore first but yes.

mijailr (approved):

PR #1649, opened 2025-06-23
refactor(dashboard): all links are with tenant now

GhostArrays:


dashboard/src/app/[tenantId]/TaskDefs/[taskDefId]/TaskDefClient.tsx
```diff
@@ -52,10 +52,10 @@ export default function TaskDefClient({ taskDef }: TaskDefClientProps) {
     <div className="container mx-auto py-6">
       {/* Header */}
       <div className="mb-6 flex items-center gap-4">
-        <Link href={`/${tenantId}`} className="text-muted-foreground hover:text-foreground flex items-center gap-2">
+        <LinkWithTenant href="/" className="text-muted-foreground hover:text-foreground flex items-center gap-2">
```
Comment (mijailr) [inline dashboard/src/app/[tenantId]/TaskDefs/[taskDefId]/TaskDefClient.tsx:55]: I see this is just to go back, can't you simple use 

`href="../"` ?

Comment (GhostArrays) [reply dashboard/src/app/[tenantId]/TaskDefs/[taskDefId]/TaskDefClient.tsx:55]: yea that is true but that may not always be the case that its one level above it may change in the future and to avoid breaking links i think its fine to be a little more verbose here.

bryson-g (approved):

PR #1655, opened 2025-06-23
feat(sdk-java): Add SDK methods for validating and generating `StructDef`s

Snarr:
This PR includes:
- An update to `TypeDefinition`s that adds `StructDefId`s, `InlineStructDef`s, and `InlineArrayDef`s as valid types
- A new `@LHStructDef` annotation that you can add to Java classes to declare your `StructDef`s
- A new `@LHStructField` annotation that you can add to `@LHStructDef` getters/setters to optionally name your fields or declare them as masked
- A new `@LHStructIgnore` annotation to ignore certain properties in an `@LHStructDef` class
- Updates to the `LHTaskSignature` in `sdk-java` that allow us to support `StructDef`s as parameter and return types from Task Methods
- Updates to the `LHTaskWorker` in `sdk-java` that allow us to validate all `StructDef`s used in a Task Method before registering the Task Method
- New client side methods for registering and validating all `StructDef`s used in a Task Method
- Lots of work around the edges to make `StructDef`s blend smoothly with our other client annotations and functionality, like the `@LHType` annotation

### DIFF
--- a/dashboard/src/app/utils/variables.ts
+++ b/dashboard/src/app/utils/variables.ts
@@ -113,7 +113,17 @@ export const getTypedVariableValue = (
  */
 export const getVariableDefType = (varDef: VariableDef): NonNullable<VariableValue['value']>['$case'] => {
   if (varDef.typeDef) {
-    return getVariableCaseFromType(varDef.typeDef.type)
+    switch (varDef.typeDef?.definedType?.$case) {
+      case 'primitiveType':
+        return getVariableCaseFromType(varDef.typeDef.definedType.value)
+      case 'structDefId':
+      case 'inlineStructDef':
+        return 'struct'
+      case 'inlineArrayDef':
+        return 'jsonArr'
+      default:
+        throw new Error('Unknown variable type.')
+    }
   } else if (varDef.type) {
     return getVariableCaseFromType(varDef.type)
   }
--- a/sdk-python/littlehorse/worker.py
+++ b/sdk-python/littlehorse/worker.py
@@ -177,7 +177,9 @@ class LHTask:
         self._validate_match()
 
     def _validate_match(self) -> None:
-        task_def_vars = [to_type(var.type_def.primitive_type) for var in self.task_def.input_vars]
+        task_def_vars = [
+            to_type(var.type_def.primitive_type) for var in self.task_def.input_vars
+        ]
 
         callable_params = [
             param.annotation
--- a/sdk-python/tests/test_worker.py
+++ b/sdk-python/tests/test_worker.py
@@ -175,7 +175,9 @@ class TestLHTask(unittest.TestCase):
                 input_vars=[
                     VariableDef(
                         name="param",
-                        type_def=TypeDefinition(primitive_type=VariableType.STR, masked=False),
+                        type_def=TypeDefinition(
+                            primitive_type=VariableType.STR, masked=False
+                        ),
                     )
                 ]
             ),
@@ -224,11 +226,15 @@ class TestLHTask(unittest.TestCase):
             input_vars=[
                 VariableDef(
                     name="paramA",
-                    type_def=TypeDefinition(primitive_type=VariableType.STR, masked=False),
+                    type_def=TypeDefinition(
+                        primitive_type=VariableType.STR, masked=False
+                    ),
                 ),
                 VariableDef(
                     name="paramB",
-                    type_def=TypeDefinition(primitive_type=VariableType.INT, masked=False),
+                    type_def=TypeDefinition(
+                        primitive_type=VariableType.INT, masked=False
+                    ),
                 ),
             ]
         )
@@ -246,11 +252,15 @@ class TestLHTask(unittest.TestCase):
             input_vars=[
                 VariableDef(
                     name="param2",
-                    type_def=TypeDefinition(primitive_type=VariableType.INT, masked=False),
+                    type_def=TypeDefinition(
+                        primitive_type=VariableType.INT, masked=False
+                    ),
                 ),
                 VariableDef(
                     name="param1",
-                    type_def=TypeDefinition(primitive_type=VariableType.STR, masked=False),
+                    type_def=TypeDefinition(
+                        primitive_type=VariableType.STR, masked=False
+                    ),
                 ),
             ]
         )
@@ -271,11 +281,15 @@ class TestLHTask(unittest.TestCase):
             input_vars=[
                 VariableDef(
                     name="param1",
-                    type_def=TypeDefinition(primitive_type=VariableType.STR, masked=False),
+                    type_def=TypeDefinition(
+                        primitive_type=VariableType.STR, masked=False
+                    ),
                 ),
                 VariableDef(
                     name="param2",
-                    type_def=TypeDefinition(primitive_type=VariableType.INT, masked=False),
+                    type_def=TypeDefinition(
+                        primitive_type=VariableType.INT, masked=False
+                    ),
                 ),
             ]
         )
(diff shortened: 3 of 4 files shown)


Snarr:
> the `buildVariableDef` method contains conflicting logic that leads to incorrect `TypeDefinition` construction. Parameters with `@LHStructDef` annotated types are erroneously registered as primitive types with `masked=false`,

That's just not true.

server/src/main/java/io/littlehorse/common/model/getable/global/structdef/StructFieldDefModel.java
```diff
@@ -62,7 +62,7 @@ public void validate() {
                     Status.INVALID_ARGUMENT,
                     MessageFormat.format(
                             "StructFieldDef field type [{0}] is not compatible with the provided default value.",
-                            this.fieldType.getType().name()));
+                            this.fieldType.getPrimitiveType().name()));
```
eduwercamacaro: From what I can see, this line will fail if fieldType is not a primitive type. Like for example referencing to another StructDef

eduwercamacaro: Can we add a null check? It is not safe to rely on an output from `getPrimitiveType` as it is optional in the proto definition.

Snarr: @eduwercamacaro Yes, this is another case where I have not yet added the other `defined_type` cases to the server code. For now, it is impossible to declare a `Struct` value as the default value of a `StructFieldDef`, as `Struct`s don't exist on the server yet, so we cannot add the logic for the other cases here. But I can add a null check and maybe throw an error otherwise?

Snarr: Update: I just removed the `getPrimitiveType()` method entirely. Now, the entire field type will be printed regardless of which defined case is selected 👍

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/FixedSpawnedThreads.java
```diff
@@ -22,7 +22,7 @@ public WaitForThreadsNode buildNode() {
         WaitForThreadsNode.Builder waitNode = WaitForThreadsNode.newBuilder();
         for (SpawnedThread spawnedThread : spawnedThreads) {
             WfRunVariableImpl threadNumberVariable = (WfRunVariableImpl) spawnedThread.getThreadNumberVariable();
-            if (!threadNumberVariable.getType().equals(VariableType.INT)) {
+            if (!threadNumberVariable.getTypeDef().getPrimitiveType().equals(VariableType.INT)) {
```
eduwercamacaro, line 25: It could be useful for both internal validations and clients to have this check in the WfRunVariable level. I'm thinking on something like this: `WfRunVariable#canBeType(VariableType)`. What do you think?

I'm trying to move this check into one single place because I see this in other classes.

sdk-java/src/main/java/io/littlehorse/sdk/worker/LHTaskWorker.java
```diff
@@ -153,11 +163,59 @@ public boolean doesTaskDefExist() {
...
+            ValidateStructDefEvolutionResponse resp =
+                    grpcClient.validateStructDefEvolution(validateStructDefRequest.build());
+
+            if (!resp.getIsValid()) {
+                invalidStructDefs.add(structDef.getId().getName());
+            }
+        }
+
+        if (!invalidStructDefs.isEmpty()) {
+            throw new RuntimeException("Invalid StructDefs: " + invalidStructDefs.toString());
```
eduwercamacaro, line 204: I suggest creating a specific exception type here, so clients can catch them. RuntimeException is too generic and can be anything even a bug (for example NullPointerException)

eduwercamacaro: And we should added into the `registerTaskDef` method's java doc

sdk-java/src/main/java/io/littlehorse/sdk/worker/LHStructIgnore.java
```diff
@@ -0,0 +1,10 @@
+package io.littlehorse.sdk.worker;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface LHStructIgnore {}
```
eduwercamacaro, line 10: Lets add javadoc for this new annotation

sdk-java/src/main/java/io/littlehorse/sdk/worker/LHStructDef.java
```diff
@@ -0,0 +1,23 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+public @interface LHStructDef {
+    /**
+     * This corresponds to the name of the StructDef.
+     *
+     * @return the StructDef name.
+     */
+    String name();
+
+    String description() default "";
```
eduwercamacaro, line 22: Missing javadoc

examples/struct-def/src/main/java/io/littlehorse/examples/Person.java
```diff
@@ -0,0 +1,27 @@
+package io.littlehorse.examples;
+
+import io.littlehorse.sdk.worker.LHStructDef;
+
+@LHStructDef(name="person")
+public class Person {
+  public String firstName;
```
l4crito: will this work when we set the properties to private?

Snarr: not in my testing 😢 I will add tests covering that scenario

examples/struct-def/src/main/java/io/littlehorse/examples/StructDefExample.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public static void main(String[] args) throws IOException {
+        // Let's prepare the configurations
+        Properties props = getConfigProps();
+        LHConfig config = new LHConfig(props);
+
+        // New worker
+        LHTaskWorker worker = getTaskWorker(config);
+
+        // Register StructDefs
```
l4crito, line 41: So this means that all the struc defs will be registered with no NO_SCHEMA_UPDATES?, what if i have 2 models 
-Person
-Car
i want Car to be modified but not Person, can i accomplish that with the current setup?
may be move the compatibility to each definition

@LHStructDef(name="person", compatibility=StructDefCompatibilityType.NO_SCHEMA_UPDATES)

Snarr: @l4crito Great observation. I wanted to keep the `StructDefCompatibilityType` out of the LHStructDef annotation since it is a property of the Put/Validate requests, but this seems like a valid use case where that isn't sufficient.

We can add it to the StructDef annotation. We can (and should) give users an option for manually registering their StructDefs by saying `worker.registerStructDef(Car.class)`.

Snarr: @l4crito Another thing to keep in mind is that if you modify `Person`, it will increase in version number, and thus for `Car` to depend on the new version, `Car` would also need to be updated.

aka:

If `Car 0.0` depends on `Person 0.0`
And `Person 0.0` upgrades to `0.1`
We'd need a new version of `Car` that depends on `Person 0.1` (I think)

PR #1656, opened 2025-06-23
feat(server): Rocksdb config optimizations

eduwercamacaro:
This PR allows clients to increase Littlehorse's state stores throughput by increasing the number of background jobs and sub-compactions in Rocksdb.
Also, it fixes a misconfiguration at the global `WriteBufferManager` that controls write stalls generated by the write buffer.


server/src/main/java/io/littlehorse/common/LHServerConfig.java
```diff
@@ -1125,7 +1125,7 @@ private void initRocksdbSingletons() {
         long totalWriteBufferSize = Long.valueOf(getOrSetDefault(ROCKSDB_TOTAL_MEMTABLE_BYTES_KEY, "-1"));
         if (totalWriteBufferSize != -1) {
             this.globalRocksdbWriteBufferManager =
-                    new WriteBufferManager(totalWriteBufferSize, globalRocksdbBlockCache);
+                    new WriteBufferManager(totalWriteBufferSize, globalRocksdbBlockCache, true);
```
eduwercamacaro, line 1128: `allowStalls` is true now.

server/src/main/java/io/littlehorse/common/util/RocksConfigSetter.java
```diff
@@ -55,6 +55,7 @@ public void setConfig(final String storeName, final Options options, final Map<S
             // to .close() it to avoid leaks so that we can provide a global one.
             Cache oldCache = tableConfig.blockCache();
             tableConfig.setBlockCache(serverConfig.getGlobalRocksdbBlockCache());
+            tableConfig.setCacheIndexAndFilterBlocks(true);
```
eduwercamacaro, line 58: `BlockCache` will bound index and filter blocks.

server/src/main/java/io/littlehorse/common/LHServerConfig.java
coltmcnealy-lh: My mistake—I thought true was the default. Great catch!

coltmcnealy-lh (approved):
Nice!

PR #1660, opened 2025-06-24
feat(server): support masked correlation keys

coltmcnealy-lh:
- implements masked correlation keys in java
- allows users to keep sensitive data hidden when using it to correlate events
- TODO: python/go/csharp

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -87,6 +87,13 @@ jobs:
         with:
           node-version: 20
 
+      - name: Update Dashboard Version
+        working-directory: ./dashboard
+        run: |
+          SHORT_SHA=$(echo "${{ github.sha }}" | cut -c1-7)
+          VERSION="${{ github.ref_name }}-$SHORT_SHA"
+          sed -i "s|NEXT_PUBLIC_VERSION: \".*\"|NEXT_PUBLIC_VERSION: \"$VERSION\"|g" next.config.ts
+
       - name: Build Dashboard
         working-directory: ./dashboard
         run: |
@@ -113,6 +120,13 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 20
+
+      - name: Update Dashboard Version
+        working-directory: ./dashboard
+        run: |
+          SHORT_SHA=$(echo "${{ github.sha }}" | cut -c1-7)
+          VERSION="${{ github.ref_name }}-$SHORT_SHA"
+          sed -i "s|NEXT_PUBLIC_VERSION: \".*\"|NEXT_PUBLIC_VERSION: \"$VERSION\"|g" next.config.ts
 
       - name: Build Dashboard
         working-directory: ./dashboard
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -119,6 +119,17 @@ jobs:
         with:
           node-version: 20
 
+      - name: Update Dashboard Version
+        working-directory: ./dashboard
+        run: |
+          if [ "${{ inputs.tag }}" != "" ]; then
+            VERSION="${{ inputs.tag }}"
+          else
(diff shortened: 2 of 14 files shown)


coltmcnealy-lh:
> ### Bug: Masking Error in Event Node Serialization
> The `ExternalEventNodeRunModel` incorrectly masks the `correlationKey` during deserialization in `initFrom()` when `RequestExecutionContext` is supported and masking is enabled. This masked value (`LHConstants.STRING_MASK`) is then persisted back to storage via `toProto()`, leading to permanent loss of the original correlation key. Masking should only occur at the presentation layer or during serialization for display, not during data loading or persistence.
> 
> `server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/ExternalEventNodeRunModel.java#L65-L73`
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjYwYWFlNTUwLWMxZTYtNDMxNi1iMjhkLTdkZTIzM2ViZWY3YiIsImVuY3J5cHRpb25LZXkiOiJncGlrRm5SemM5UHRPRU54YUFJTlh6dmY4ZmZOU1ItTWI1b1BvRUJXWFRBIiwiYnJhbmNoIjoiMTYzMi9tYXNrLWNvcnJlbGF0ZWQtZXZlbnQtamF2YSJ9LCJpYXQiOjE3NTA3NDIxODcsImV4cCI6MTc1MTM0Njk4N30.VqIwZQKrFewY1HJZoheNTOn0j9HhjKgKd_w2JU5yMCpZdy9cuS9dzpfJiCFtQz1Dq7-nY1-sw4b3Gu_TsPqyLTbYVG88wj2qzc0RWi2yuzQwUZMcTG2G2cL68f2lPv61Nk3cocE7JwP-855f3VSCd4OuKw9sMhj75ZHatu4TVv7oa7tahrRyr8tTt7B1OY3ZItoORERrhtcpBBMFO2oGjMjK4DsjZw2k3tn0QPVInjV_SM1gcshKhfLPgITonOrXzE2kzW1mNLD5C_5c7p6hXQl4t6J6-xoXiX_sQ248hofJv3qYLj_i0CTf2I0qA3nARKypl0GRqRytnav7u7lwIQ)
> 
> _Was this report helpful? Give feedback by reacting with 👍 or 👎_

Well, copilot is an idiot

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/ExternalEventNodeOutputImpl.java
```diff
@@ -31,7 +31,16 @@ public ExternalEventNodeOutput registeredAs(Class<?> payloadClass) {
 
     @Override
     public ExternalEventNodeOutput withCorrelationId(Serializable correlationId) {
-        parent.addCorrelationIdToExtEvtNode(this, correlationId);
+        boolean shouldMaskAutomatically =
+                (correlationId instanceof WfRunVariableImpl) && ((WfRunVariableImpl) correlationId).isMasked();
```
eduwercamacaro, line 35: I think we should also cover the case when we receive a masked `NodeOutput`. Very difficult to do with the current structure I believe.

eduwercamacaro (approved):

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/ExternalEventNodeOutputImpl.java
coltmcnealy-lh: The idea is that we will cover this in the server side. Right now we cant without making a call to the server

PR #1667, opened 2025-06-24
fix(dashboard) use `getVariable`

hazimoarafa:
image


dashboard/src/utils/data/variables.ts
```diff
@@ -66,20 +66,4 @@ export const getTypedContent = (contentType: string, contentValue: string) => {
     default:
       return { str: contentValue }
   }
-}
-
```
Snarr, line 70: Cool if you relocate this code or refactor it but make sure the dashboard still supports both old and new VarDefs

PR #1676, opened 2025-06-25
refactor(dashboard): general cleanup

hazimoarafa:


dashboard/src/utils/data/node-tree.ts
```diff
@@ -1,4 +1,30 @@
...
+export interface TreeNode {
+  id: string
+  label: string
+  type?: string
+  status?: string
+  children: TreeNode[]
+  level: number
+}
+
+export function buildNodeTree(wfSpec: WfSpec, threadSpecName: string): TreeNode[] {
```
Snarr, line 12: This just builds the top level of the tree, right ? Or is this more like a list at this point in the code?

GhostArrays: this is a util function we use for building the react flow diagram

PR #1678, opened 2025-06-25
feat(sdk-dotnet): adds native support for async Tasks in .NET

MateoRojas:


sdk-dotnet/LittleHorse.Sdk.Tests/Worker/LHTaskSignatureTest.cs
```diff
@@ -213,61 +214,62 @@ class TestWorker
...
-        public string Add([LHType(masked: TRUE_IS_MASKET)] string name)
+        public Task<string> Add([LHType(masked: TRUE_IS_MASKET)] string name)
         {
-            return $"Output value: {name}";
+            return Task.FromResult($"Output value: {name}");
         }
         
         [LHTaskMethod(TASK_DEF_NAME_INFORM)]
-        public void Inform([LHType(masked: TRUE_IS_MASKET)] string name)
+        public Task Inform([LHType(masked: TRUE_IS_MASKET)] string name)
```
eduwercamacaro, line 223: I think this method should be `public async Task` instead of returning `Task.CompletedTask`

eduwercamacaro: The same applies to any other task worker on this test class.

sdk-dotnet/LittleHorse.Sdk/LHConfig.cs
```diff
@@ -67,80 +67,44 @@ public LHConfig(Dictionary<string, string> configArgs, ILoggerFactory? loggerFac
             LHLoggerFactoryProvider.Initialize(loggerFactory);
             _logger = LHLoggerFactoryProvider.GetLogger<LHConfig>();
             _inputVariables = new LHInputVariables(configArgs);
-            _createdChannels = new Dictionary<string, LittleHorseClient>();
```
eduwercamacaro, line 70: oh! good catch.

sdk-dotnet/LittleHorse.Sdk/LHConfig.cs
```diff
@@ -211,20 +169,14 @@ public LittleHorseClient GetGrpcClientInstance(string host, int port)
...
-            _logger?.LogInformation("Establishing connection to: " + channelKey);
-
-            GrpcChannel channel = CreateChannel(host, port);
-            
-            var lhClient = new LittleHorseClient(channel);
-
-            _createdChannels.Add(channelKey, lhClient);
-
-            return lhClient;
+            return new LittleHorseClient(channel);
```
eduwercamacaro, line 179: I believe users should have the option to reuse client stubs. or perhaps we can say in the docs that reusing the client is a user responsibility?

PR #1679, opened 2025-06-25
feat(sdk-python): masked correlation id support in python

joleson15:


sdk-python/examples/correlated_event/example_correlated_event.py
```diff
@@ -0,0 +1,70 @@
...
+    def my_entrypoint(wf: WorkflowThread) -> None:
+        wf.execute(ASK_FOR_NAME)
+        ext_event_output = wf.wait_for_event(
+            EXT_EVENT,
+            correlation_id="some-correlation-id",
+            mask_correlation_id=True
+        )
+
+        name = wf.add_variable("name", VariableType.STR)
+        wf.mutate(name, VariableMutationType.ASSIGN, ext_event_output)
```
coltmcnealy-lh: Super small nit pick: we're trying to get people to use `name.assign(ext_event_output)` rather than `wf.mutate(...)`. The external event example was written before this so it wasn't available. I'd like our examples to start using the "recommended" patterns.

(Also note: you'll need to remove the unused import for `VariableMutationType` above or the linter will get grumpy again)

sdk-python/tests/test_workflow.py
```diff
@@ -3754,6 +3754,66 @@ def wf_func(wf: WorkflowThread) -> None:
             "my-var",
         )
 
+    def test_automatically_mask_correlation_id_from_masked_var(self):
```
coltmcnealy-lh, line 3757: Beautiful tests

coltmcnealy-lh (approved):

Issue #1008, opened 2024-09-18
Server shutdown process is not working properly


eduwercamacaro:
We expect to finish all the command waiters during a rolling restart or normal server shutdown before turning off the Kafka streams instance. If not, the command topic will receive the command record, but clients will receive a CANCEL status code.

eduwercamacaro:
@coltmcnealy-lh any thoughts?

coltmcnealy-lh:
This is worth investigating and/or fixing. Maybe @l4crito can take this?

coltmcnealy-lh:
This has been fixed by #1680

PR #1680, opened 2025-06-26
Improve Server 

l4crito:
#1008

### DIFF
--- a/server/src/main/java/io/littlehorse/server/LHServer.java
+++ b/server/src/main/java/io/littlehorse/server/LHServer.java
@@ -181,46 +181,59 @@ public class LHServer {
     }
 
     public void close() {
-        CountDownLatch latch = new CountDownLatch(4 + listeners.size());
 
-        new Thread(() -> {
-                    log.info("Closing timer Kafka Streams");
-                    timerStreams.close();
-                    latch.countDown();
-                    log.info("Done closing timer Kafka Streams");
-                })
-                .start();
+        CountDownLatch listenerLatch = new CountDownLatch(listeners.size());
 
-        new Thread(() -> {
-                    log.info("Closing core Kafka Streams");
-                    coreStreams.close();
-                    latch.countDown();
-                    log.info("Done closing core Kafka Streams");
-                })
-                .start();
-
-        new Thread(() -> {
-                    log.info("Closing internalComms");
-                    internalComms.close();
-                    latch.countDown();
-                })
-                .start();
-
-        new Thread(() -> {
-                    log.info("Closing health service");
-                    healthService.close();
-                    latch.countDown();
-                })
-                .start();
+        ExecutorService listenerExecutor = Executors.newFixedThreadPool(listeners.size());
+        
+        log.info("Closing {} listeners", listeners.size());
 
         for (LHServerListener listener : listeners) {
-            new Thread(() -> {
-                        log.info("Closing listener {}", listener);
-                        listener.close();
-                        latch.countDown();
-                    })
-                    .start();
+            listenerExecutor.submit(() -> {
+                log.info("Closing listener {}", listener);
+                listener.close();
+                listenerLatch.countDown();
+            });
         }
+        listenerExecutor.shutdown();
+        try {
+            listenerLatch.await();
+            log.info("All listeners closed successfully");
+        } catch (InterruptedException e) {
+            log.error("Listener close interrupted", e);
+        }
+
+        CountDownLatch latch = new CountDownLatch(4 );
+
+        ExecutorService shutdownExecutor = Executors.newFixedThreadPool(4);
+
+        shutdownExecutor.submit(() -> {
+            log.info("Closing timer Kafka Streams");
+            timerStreams.close();
+            latch.countDown();
+            log.info("Done closing timer Kafka Streams");
+        });
+
+        shutdownExecutor.submit(() -> {
+            log.info("Closing core Kafka Streams");
+            coreStreams.close();
+            latch.countDown();
+            log.info("Done closing core Kafka Streams");
+        });
+
+        shutdownExecutor.submit(() -> {
+            log.info("Closing internalComms");
+            internalComms.close();
+            latch.countDown();
+        });
+
+        shutdownExecutor.submit(() -> {
+            log.info("Closing health service");
+            healthService.close();
+            latch.countDown();
+        });
+
+        shutdownExecutor.shutdown();
 
         try {
             latch.await();
--- a/server/src/main/java/io/littlehorse/server/LHServer.java
+++ b/server/src/main/java/io/littlehorse/server/LHServer.java
@@ -58,6 +58,7 @@ public class LHServer {
     private final CommandSender commandSender;
     private final LHInternalClient lhInternalClient;
     private final AsyncWaiters asyncWaiters = new AsyncWaiters();
+    public static final int CORE_PROCESSES_COUNT = 4; // core , timer, internal and health
 
     private RequestExecutionContext requestContext() {
         return contextKey.get();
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/littlehorse/server/interceptors/ShutdownInterceptor.java
```diff
@@ -0,0 +1,37 @@
...
+
+import io.grpc.Metadata;
+import io.grpc.ServerCall;
+import io.grpc.ServerCallHandler;
+import io.grpc.ServerInterceptor;
+import io.grpc.Status;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class ShutdownInterceptor implements ServerInterceptor {
```
eduwercamacaro: `ShutdownInterceptor` has a public scope with static methods which can cause undesired behaviors if someone tries to call it outside of a shutdown process. Instead make either private or only accessible only in the scope of the `LHServer` class.

server/src/main/java/io/littlehorse/server/interceptors/ShutdownInterceptor.java
```diff
@@ -0,0 +1,37 @@
...
+
+    private static volatile boolean isShuttingDown = false;
```
eduwercamacaro: I think a AtomicBoolean fits better in this case as it provides a nice API to protect the consistency of this value

server/src/main/java/io/littlehorse/server/interceptors/ShutdownInterceptor.java
```diff
@@ -0,0 +1,37 @@
...
+
+    public static boolean isShuttingDown() {
+        return isShuttingDown;
+    }
+
+    @Override
+    public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
+            ServerCall<ReqT, RespT> call, Metadata headers, ServerCallHandler<ReqT, RespT> next) {
+
+        if (isShuttingDown) {
```
eduwercamacaro: This is a volatile variable reference which can be expensive to read during runtime 
I think we can remove this comparison if we use a `Strategy` pattern and changing the behavior of this class during runtime. 

The way I can see it is by defining two different strategies for this class. First strategy is just call `next.startCall` and the second one immediately closes the call with a `UNAVAILABLE` status code. 

What do you think?

server/src/main/java/io/littlehorse/server/LHServer.java
```diff
@@ -181,47 +186,68 @@ public void start() throws IOException {
...
-                })
-                .start();
-
-        new Thread(() -> {
-                    log.info("Closing health service");
-                    healthService.close();
-                    latch.countDown();
-                })
-                .start();
+        ShutdownInterceptor.setShuttingDown();
```
eduwercamacaro: Wondering if we should move this into its own shutdown hook.

server/src/main/java/io/littlehorse/server/LHServer.java
```diff
@@ -181,47 +186,68 @@ public void start() throws IOException {
...
+
+        log.info("Waiting {} seconds for active requests to complete...", GRACE_PERIOD_SECONDS);
+        try {
+            TimeUnit.SECONDS.sleep(GRACE_PERIOD_SECONDS);
```
eduwercamacaro: I'm not sure about this one.
Shutdowns should run as fast as possible because it could cause unclean shutdowns if it reaches any deadline provided by a container orchestrator. 

cc: @coltmcnealy-lh

server/src/main/java/io/littlehorse/server/LHServer.java
```diff
@@ -181,47 +186,68 @@ public void start() throws IOException {
...
+        } catch (InterruptedException e) {
+            log.warn("Grace period wait was interrupted", e);
+        }
+
+        CountDownLatch listenerLatch = new CountDownLatch(listeners.size());
+
+        ExecutorService listenerExecutor = Executors.newFixedThreadPool(listeners.size());
```
eduwercamacaro: I believe we can still sequentially close each listener. what's the benefit of closing them in parallel?

server/src/main/java/io/littlehorse/server/interceptors/ShutdownInterceptor.java
```diff
@@ -0,0 +1,37 @@
...
+import io.grpc.Status;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class ShutdownInterceptor implements ServerInterceptor {
+
+    private static volatile boolean isShuttingDown = false;
+
+    public static void setShuttingDown() {
+        log.info("Setting server shutdown state to TRUE, server is shutting down.");
```
eduwercamacaro: Do we really need this info log here?

eduwercamacaro (commented):
Thank you for the PR. I like the idea of closing the requests before reaching a listener when the server is shutting down.

server/src/main/java/io/littlehorse/server/LHServer.java
l4crito: Now when the shutdown hook is triggered the interceptors will reject all the new request, in the meanwhile the shutdown proccess will be, turn off streams first (core and timer) , then (coms, and health) finally the grpc

server/src/main/java/io/littlehorse/server/LHServer.java
```diff
@@ -181,23 +184,31 @@ public void start() throws IOException {
...
                     log.info("Closing core Kafka Streams");
                     coreStreams.close();
-                    latch.countDown();
+                    streamLatch.countDown();
                     log.info("Done closing core Kafka Streams");
                 })
                 .start();
+        try {
+            streamLatch.await();
+            log.info("Done shutting down all LHServer threads");
```
eduwercamacaro: I believe we can improve this log message.
Probably we should mention something like:
`Done shutting down all Kafka Stream threads`

server/src/main/java/io/littlehorse/server/LHServer.java
```diff
@@ -181,23 +184,31 @@ public void start() throws IOException {
...
+        } catch (Exception exn) {
+            throw new RuntimeException(exn);
```
eduwercamacaro: I'm curious about the exception that can be thrown at this point. maybe timeout exceptions?

eduwercamacaro (approved):
This looks good to me, but I think we should wait for @coltmcnealy-lh's and/or @sauljabin opinion as well.

server/src/main/java/io/littlehorse/server/LHServer.java
```diff
@@ -181,23 +184,31 @@ public void start() throws IOException {
     }
 
     public void close() {
-        CountDownLatch latch = new CountDownLatch(4 + listeners.size());
 
+        CountDownLatch streamLatch = new CountDownLatch(2);
```
coltmcnealy-lh, line 188: Hmm. I think the "shutdown hook" in the `RequestBlocker` is clever; however in my opinion it would be more explicit and easier to follow if we were to explicitly call `requestBlocker.blockRequests()` here.

Additionally, what if someone runs an "embedded" LH Server? They may want to call `close()` on the server without waiting for a shutdown hook.

coltmcnealy-lh (commented):
Small change requested.

coltmcnealy-lh:
With these same configs, but with direct I/O disabled, I was able to get the pods to OOM by doing:

* Update one of the Task Workers to send a ton of data (32KB output)
* Run hundred-tasks every 0.1 seconds for 3 minutes

server/src/main/java/io/littlehorse/server/LHServer.java
eduwercamacaro: I think it should be as private as possible and only call `requestBlocker.blockRequests()` inside the context of a ShutdownHook. 

"Embedded" lh server users can create an instance of the RequestBlocker interceptor and pass it to the GrpcListener if they want. Or, they can create an instance of a LHServerListener which uses the RequestBlocker interceptor by default.

PR #1683, opened 2025-06-27
feat(server): Support WfRunId variables in a workflow

eduwercamacaro:
This PR adds a new primitive variable type in Littlehorse: WfRunId

This aims to support inter-workflow interactions. 

Users will be able to declare and assign variables of type WfRunId

follow-up PR will be added for the others sdks: dotnet, go and python


coltmcnealy-lh:
This is also useful for things like when you want to pass a WfRunId into a task without the Task Worker code coupling itself to use the `WorkerContext`. For example, it can be a "create-user" task that accepts a user id. Certain workflows might want to use their own workflow id as the user id; others might use different id's.

coltmcnealy-lh (approved):
Looks good to me!

coltmcnealy-lh:
Follow-up features before closing the ticket:

- [ ] Allow `.declareWfRunId(...)`
- [ ] Allow a `WorkflowThread#getWfRunId()` thing that returns a `LHExpression`

```java
WfRunVariable wfRunId = wf.declareWfRunId("some-id");

wfRunId.assign(wf.getWfRunId());
```

PR #1684, opened 2025-06-27
docs(readme): fix golang badge

mijailr:
Previous badge was referencing the last published tag, which not necessarily is the `latest` release, this changes to point to the latest release instead.


Snarr:
image
The new badge does not display on my end :O

Snarr (approved):

Issue #1669, opened 2025-06-24
Restructure node data on right sidebar and its convention across the project


bryson-g:
- The `right-sidebar/nodes` and `NodeDefinition.tsx` implementations need to be refactored. More componentized, declarative, and less bad.

PR #1687, opened 2025-06-27
feat(dashboard): Proto representation on right-sidebar

bryson-g:
# Description
resolves [#1669](https://github.com/littlehorse-enterprises/littlehorse/issues/1669)

This PR focuses primarily on three aspects and is not a completion of the right side-bar as a whole:
## Foundational Styling
_How to represent Proto objects within the right-sidebar_
- It is composed of `Section` and `Label` components. Sections represent objects, Labels represent fields of primitive data types.
- Labels have ellipsis truncation when overflowing. This can be revisited in a later PR to optimize for a better user experience. Otherwise, we must rely on expanding the sidebar for observing truncated text. We have an issue already open to better the sidebar expansion experience: https://github.com/orgs/littlehorse-enterprises/projects/2?pane=issue&itemId=116448126&issue=littlehorse-enterprises%7Clittlehorse%7C1644
- The two available tabs to switch between within the right-sidebar are: "Spec" and "Run." `Spec` represents data that is associated with a `WfSpec`, while `Run` represents everything that exists once a `WfRun` occurs.
  - This PR focused only on the `Spec` tab. The `Run` tab will have its own PR.
- Sections are collapsible. 
- Label text and respective values both take up exactly half of the available space, split evenly.

## UI Architecture
_The project convention of components are declared and assigned to nodes_
- All proto components for the right-sidebar exist under `src/components/diagram/right-sidebar/proto-components`
  - There is a unique component defined for each proto object used, so that they can be reused anywhere, as Proto objects often have nested objects within them.
- The `oneof node` is determined by the new `getNodeType` utility function, which attempts to abstract away the convoluted type casting required to associate a `Node` object with its respective `oneof` value.
  - The `node-definition.tsx` file then conditionally renders the appropriate proto-component for the given `Node`.

## Proto Data
_Data shown for Spec tab_
- This PR is only focusing on the `Spec` tab. The `Run` tab will have a separate PR for its data to be included.
  > The two available tabs to switch between within the right-sidebar are: "Spec" and "Run." `Spec` represents data that is associated with a `WfSpec`, while `Run` represents everything that exists once a `WfRun` occurs.
- All implemented Proto data can be observed under the `proto-components` directory
  > - All proto components for the right-sidebar exist under `src/components/diagram/right-sidebar/proto-components`
  There is a unique component defined for each Proto object used, so that they can be reused anywhere, as Proto objects often have nested objects within them.

# Video
https://github.com/user-attachments/assets/333ada6d-fdd1-49fa-8ebb-d864b54491ac


hazimoarafa (approved):
Just make sure to not hard code node types with strings and use the proto but other than that LGTM!

bryson-g:
> Just make sure to not hard code node types with strings and use the proto but other than that LGTM!

image

dashboard/src/components/diagram/right-sidebar/proto-components/nodes/user-task.tsx
```diff
@@ -0,0 +1,36 @@
...
+      {userTask.notes && <Label label="Notes">{getVariable(userTask.notes)}</Label>}
+      {userTask.onCancellationExceptionName && <Label label="OnCancellationExceptionName">{getVariable(userTask.onCancellationExceptionName)}</Label>}
+      <Section title="Actions">
+        {userTask.actions.map((action, index) => (
+          <Section key={JSON.stringify(action) + index} title={`UTActionTrigger ${index + 1}`}>
+            {action.task && <Section title="UTATask">
+              <TaskNodeComponent {...(action.task.task as TaskNode)} />
+              <Section title="Mutations">
+                {action.task.mutations.map((mutation, index) => (
+                  <VariableMutationComponent key={JSON.stringify(mutation) + index} {...mutation} />
```
mijailr: Why `JSON.stringify` for the component key?

mijailr: Looks like what you're having is duplicated keys because you were assigning just the index as key which could return a warning regarding reusing the same key for multiple components in a stacked map.

Would be better if you prefix the id with something, not the full json fo the action, otherwise the component size would be big just because.

do something like `key={"actions-${actionIndex}"}` and `key={"mutation-${actionIndex}-${mutationIndex}"}`

Rename the `index` to `actionIndex` and `mutationIndex` so they don't get shadowed.

dashboard/src/components/diagram/right-sidebar/proto-components/variable-assignment.tsx
```diff
@@ -0,0 +1,11 @@
...
+import { VariableAssignment } from "littlehorse-client/proto";
+
+// todo: more complex handling for runs, where other variableAssignment fields will be populated
+export function VariableAssignmentComponent(variableAssignment: VariableAssignment) {
+    return (
+        <div>
+            {getVariable(variableAssignment)}
+        </div>
+    )
+}
```
mijailr: Looks like you're either not running a formatter or you have disabled the end-of-file format. this is important so the next person that runs the formatter doesn't have to push unrelated changes.

dashboard/src/utils/data/extract-nodes.ts
```diff
@@ -12,7 +12,7 @@ export function extractNodes(wfSpec: WfSpec, threadSpec: ThreadSpec): CustomNode
       type: 'node',
       data: {
         node,
-        type: getNodeType(node),
+        type: getNodeType(node).type,
```
mijailr: Why is this now `getNodeType(node).type` the purpose of the function is to get the nodeType, what is this returning if not the type directly?

dashboard/src/utils/data/node.ts
```diff
@@ -15,18 +29,71 @@ export type NodeType =
...
+  WAIT_FOR_THREADS: node => !!node.waitForThreads,
+  NOP: node => !!node.nop,
+  SLEEP: node => !!node.sleep,
+  USER_TASK: node => !!node.userTask,
+  START_MULTIPLE_THREADS: node => !!node.startMultipleThreads,
+  THROW_EVENT: node => !!node.throwEvent,
+  WAIT_FOR_CONDITION: node => !!node.waitForCondition,
+}
+
+export function getNodeType(
```
mijailr: If the purpose now is not to return the nodeType, then change the function name to something more accurate.

dashboard/src/utils/data/variables.ts
```diff
@@ -19,10 +19,12 @@ function getVariableValue(variable?: VariableValue) {
   }
 }
 
-function getValueFromVariableName({
-  variableName,
-  jsonPath,
-}: Pick<VariableAssignment, 'variableName' | 'jsonPath'>) {
+export function getVariableTypeFromLiteralValue(literalValue: VariableValue) {
+  const key = Object.keys(literalValue)[0] as keyof VariableValue
```
mijailr: This would break if another key is added and the index changes, also since you're casting it to `VariableValue` we won't catch issues later. This is almost as bad as using `any`.

dashboard/eslint.config.mjs
```diff
@@ -198,6 +198,7 @@ const eslintConfig = [
     rules: {
       'custom/enforce-props-interface-naming': 'error',
       'custom/no-client-pages': 'error',
+      '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
```
mijailr: Don't use this, if the argument doesn't have any use don't add it.

mijailr (changes requested):

dashboard/src/utils/data/variables.ts
bryson-g: Is it possible that a `VariableValue` would ever have more than one key in runtime? Based from the Proto, it looks like the only field is a `oneof`.
image

dashboard/src/components/diagram/right-sidebar/proto-components/nodes/user-task.tsx
bryson-g: @hazimoarafa thoughts?

dashboard/eslint.config.mjs
bryson-g: I agree. Any objections @hazimoarafa ?

dashboard/src/components/diagram/right-sidebar/proto-components/nodes/user-task.tsx
mijailr: I see new docs are okay with stringify

dashboard/src/utils/data/variables.ts
mijailr: It can grow, more fields could be added to the proto message and key order is not guaranteed.

mijailr (approved):

dashboard/eslint.config.mjs
GhostArrays: Sure it doesn't matter either way, i personally think there is cases where its good to keep `_` but whatever.

Snarr (approved):

PR #1694, opened 2025-06-30
feat(sdk-dotnet) Support for masked correlation id in C#

joleson15:
Implemented masked correlation id, please thoroughly review and let me know if there are more tests to be run. I got a bit hung up on the `CorrelatedEventConfig` logic in the `ExternalEventNodeOutput.cs` file.


sdk-dotnet/LittleHorse.Sdk.Tests/Workflow/Spec/CorrelatedEventTest.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        if (workflow.ThreadSpecs.TryGetValue(workflow.EntrypointThreadName, out var entrypoint))
+        {
+            if (entrypoint.Nodes.TryGetValue("1-identity-verified-EXTERNAL_EVENT", out var node))
+            {
+                Assert.True(node.ExternalEvent.MaskCorrelationKey);
+            }
+            else
+            {
+                throw new Exception("External Event not found");
```
KarlaCarvajal: It is not needed to throw exceptions in the tests, as it is not the subject under test to be evaluated. Same comment for all tests.

sdk-dotnet/Examples/CorrelatedEventExample/WaitForExternalEventWorker.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        Console.WriteLine("Executing ask-for-name");
+        return "Hi what's your name?";
+    }
+    
+    [LHTaskMethod("greet")]
+    public string Greet(string name) 
+    {
+        Console.WriteLine("Executing greet");
+        return "Hello there, " + name;
```
KarlaCarvajal: In .net, you can interpolate variables with `$` like this `$"Hello there, {name}"`

sdk-dotnet/LittleHorse.Sdk/Workflow/Spec/WorkflowThread.cs
```diff
@@ -790,15 +790,15 @@ internal void AddTimeoutToExtEvtNode(ExternalEventNodeOutput node, int timeoutSe
         newNode.ExternalEvent.TimeoutSeconds = timeoutValue;
     }
 
-    internal void SetCorrelationIdOnExternalEventNode(ExternalEventNodeOutput node, Object correlationId) 
+    internal void SetCorrelationIdOnExternalEventNode(ExternalEventNodeOutput node, Object correlationId, bool maskCorrelationKey)
```
KarlaCarvajal: I think you only need a primitive `object` instead of `Object` for your usage.

sdk-dotnet/LittleHorse.Sdk.Tests/Workflow/Spec/CorrelatedEventTest.cs
eduwercamacaro: From what I can see, it is important to throw exceptions at this point because it prevents the test to ignore the assertions inside the condition. am I missing something else?

sdk-dotnet/LittleHorse.Sdk/Workflow/Spec/WorkflowThread.cs
```diff
@@ -790,15 +790,15 @@ internal void AddTimeoutToExtEvtNode(ExternalEventNodeOutput node, int timeoutSe
...
     {
         CheckIfWorkflowThreadIsActive();
         Node evtNode = FindNode(node.NodeName);
-
         evtNode.ExternalEvent.CorrelationKey = AssignVariableHelper(correlationId);
+        evtNode.ExternalEvent.MaskCorrelationKey = maskCorrelationKey; //maybe this?
```
eduwercamacaro: I think we can remove the comment at the end of this line.

sdk-dotnet/LittleHorse.Sdk/Workflow/Spec/WorkflowThread.cs
eduwercamacaro: hmm why? I believe we want an object reference here instead of a value object.

sdk-dotnet/LittleHorse.Sdk.Tests/Workflow/Spec/CorrelatedEventTest.cs
jacksonoleson: Karla and Edu, yes, this is why I implemented exceptions, the potential for null values was causing some warnings to pop up when I tried to implement something like this:
``` ThreadSpec entry = workflow.ThreadSpecs.GetValueOrDefault("entrypoint");```
Got this warning:
```Converting null literal or possible null value to non-nullable type.[CS8600]```

sdk-dotnet/LittleHorse.Sdk/Workflow/Spec/WorkflowThread.cs
KarlaCarvajal: because it is not been used more features of the class Object

jacksonoleson: I don't quite understand, although I did not change anything related to `Object`, just added the boolean parameter here.

sdk-dotnet/LittleHorse.Sdk.Tests/Workflow/Spec/CorrelatedEventTest.cs
KarlaCarvajal: I think, you should prepare the test to be sure that you have the entrypoint and the EXTERNAL_EVENT nodes that you want to test.

jacksonoleson: Could you give me some guidance on that? How would you suggest I implement this?

coltmcnealy-lh (approved):
Tests look good to me I think

KarlaCarvajal (approved):

PR #1700, opened 2025-07-02
feat(canary): introduce .NET canary worker

MateoRojas:


.github/workflows/branch.yml
```diff
@@ -60,13 +60,23 @@ jobs:
           distribution: "corretto"
           java-version: 21
 
-      - name: Tests and Build
+      - name: Setup Dotnet
+        uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: '8'
+
+      - name: Tests and Build Core
```
coltmcnealy-lh, line 68: My one concern is that the `lh-canary` build step already takes a long time. Will this make it longer than the server tests? If it's still faster than the server tests, it's fine, since we're not making the dev experience any worse.

canary/dotnet-worker/LittleHorse.Canary.Worker/CanaryConfig.cs
```diff
@@ -0,0 +1,37 @@
...
+
+public class CanaryConfig(IConfiguration configs)
+{
+
+    private const string TopicNameProperty = "topic.name";
+    private const string MetronomeWorkerEnableProperty = "metronome.worker.enable";
+    private const string MetronomeBeatExtraTagsProperty = "metronome.beat.extra.tags";
+    private const string MetronomeBeatExtraTagsPrefixProperty = $"{MetronomeBeatExtraTagsProperty}.";
+
+    private readonly Dictionary<string, string> _kafkaConfigs = configs.GetSection("Kafka").ToDictionary(true);
```
eduwercamacaro, line 17: I wonder if these lines raise an exception if none of these sections are present in the configuration file.

MateoRojas:
> I'm sorry, but I don't like the idea of copying dotnet and java files in the same docker image. Instead of making one single image, I would suggest creating two different image tags. One for java and another for dotnet. I see some problems with this PR:
> 
> * It increases the size of the canary docker image.
> * A security vulnerability in one of the dependencies in either java or dotnet will impact dotnet and java workers at the same time.
> * A helthycheck is missing in the new dotnet worker (At least I cannot find it).
> * Most of the time, client's containers will have dotnet runtime unnecessary installed (in cases where only java is required).
> * Dotnet binaries are being copy from the pipeline making it difficult to know what's the compilation version of those binaries. Instead we should build them using a multi-stage docker image.

PR #1719, opened 2025-07-03
perf(server): remember start key in timer topology range scan

coltmcnealy-lh:
- See this rocksdb documentation: https://github.com/facebook/rocksdb/wiki/Implement-Queue-Service-Using-RocksDB


mijailr:
Would this be ported to 0.14?

coltmcnealy-lh:
> Would this be ported to 0.14?

Perhaps. It's no proto change, and quite safe. But it's also not a huge issue.

Snarr (approved):

eduwercamacaro (approved):

PR #1720, opened 2025-07-07
feat(dashboard): Display thread variables

### LINKED ISSUE #1671: Add workflow variables

bryson-g:
resolves #1671

### Description
- Extends `demo_wf` for example purposes
- Includes workflow variables, categorized by thread name, under the `workflow` tab on the `left-sidebar`.
- Formats unrelated `sdk-python` files (pre-commit did this).


hazimoarafa (approved):

Snarr (approved):
Looks good, some comments:
- Variables tab should be resizable or expandable, if there are a lot of variables or data within variables it might be hard to read
- Variable Data is represented well, I like the nesting structure

bryson-g:
Agreed, prior to draft I realized this as well. Resizable issue is next on my todo list

dashboard/src/components/diagram/left-sidebar/tabs.tsx
```diff
@@ -5,7 +5,7 @@ import { Tabs, TabsList, TabsTrigger } from '@littlehorse-enterprises/ui-library
 import { Dispatch, SetStateAction } from 'react'
 
 const tabs: { id: LeftSidebarTabId; label: string }[] = [
-  { id: 'WfSpec', label: 'WfSpec' },
+  { id: 'Workflow', label: 'Workflow' },
```
mijailr, line 8: There's a preference that we call WfSpec and WfRun by the server API name so it keeps homogeneous across all context. `Workflow` is not a server object.

dashboard/src/components/diagram/left-sidebar/tab-content/workflow-tab.tsx
```diff
@@ -0,0 +1,153 @@
...
+export default function WorkflowTab({ wfSpec }: WorkflowTabProps) {
+  const [searchTerm, setSearchTerm] = useState('')
+
+  const nodeTree = buildNodeTree(wfSpec, wfSpec.entrypointThreadName)
+  const sortedNodeTree = sortNodeTree(nodeTree)
+
+  const threadSpecsWithVariableDefs = useMemo(
+    () =>
+      Object.entries(wfSpec.threadSpecs)
+        .filter(threadSpec => threadSpec[1].variableDefs.length > 0)
```
mijailr, line 54: Getting by index is dangerous, the server doesn't work with indexes, so it may change without affecting any feature in the server side, you must use a different approach.

dashboard/src/components/diagram/left-sidebar/tab-content/workflow-tab.tsx
```diff
@@ -0,0 +1,153 @@
...
+        .sort(a => (a[0] === 'entrypoint' ? -1 : 1)),
```
mijailr, line 55: what this mean?

dashboard/src/components/diagram/left-sidebar/tab-content/workflow-tab.tsx
```diff
@@ -0,0 +1,153 @@
...
+                  <Section title="VariableDef">
+                    {threadVarDef.varDef?.type && <Label label="VariableType">{`${threadVarDef.varDef?.type}`}</Label>}
+                    <Label label="Name">{`${threadVarDef.varDef?.name}`}</Label>
+                    <Label label="DefaultValue">{`${getVariableValue(threadVarDef.varDef?.defaultValue)}`}</Label>
+                    {threadVarDef.varDef?.maskedValue && (
+                      <Label label="MaskedValue">{`${threadVarDef.varDef?.maskedValue}`}</Label>
+                    )}
+                    {threadVarDef.varDef?.typeDef && (
+                      <Section title="TypeDef">
+                        <Label label="Type">{`${threadVarDef.varDef?.typeDef?.type}`}</Label>
```
mijailr: Would be better to handle the null|undefined scenarios properly, I see `?` is overused without proper handling.

dashboard/src/components/diagram/left-sidebar/tabs.tsx
bryson-g: > There's a preference that we call WfSpec and WfRun by the server API name so it keeps homogeneous across all context. `Workflow` is not a server object.

It was renamed to Workflow because, with the addition of Workflow variables, those will be populated with data that only exists when a workflow run occurs as well, which no longer makes sense to be categorized by its "WfSpec" name. I think it would be too cumbersome to separate both into identical views, with the only difference being the populated data. Renaming it to a general tab makes more sense, because the name works for both contexts.

Additionally, we plan to add an icon representing the status of the node run of each node in the node-tree area under the Workflow tab. These statuses only exist within a WfRun, so again, the WfSpec name doesn't quite make sense in this context.

dashboard/src/components/diagram/left-sidebar/tab-content/workflow-tab.tsx
bryson-g: This filters the threadSpecs by those which have `variableDefs.length > 0`, and then sorts the resulting filter so that the threadSpec whose name is `entrypoint` is always first in the array, and is displayed at the top of the UI area.

dashboard/src/components/diagram/left-sidebar/tab-content/workflow-tab.tsx
bryson-g: I don't see how it is dangerous in this scenario. The `threadSpec` parameter is not a direct getable from the server: it's an array created by `Object.entries` which can be destructured as `[key, value]`. The server works in key-value pairs as per the proto:
image

This code is simply indexing the second value in the `[key, value]` array structure.

mijailr: I see, wouldn't be more readable if you simple do a `values` instead?


```ts
 Object.values(wfSpec.threadSpecs).filter(spec => spec.variableDefs.length > 0)
```

The thread name `entrypoint` can be changed by the SDK without affectation in the behavior on the server side, so you must use `$.entrypointThreadName` to support it fully.

dashboard/src/components/diagram/left-sidebar/tab-content/workflow-tab.tsx
```diff
@@ -0,0 +1,157 @@
...
+    const bFirstPart = b.id.split('-')[0]
+
+    // If the first part is a number, sort numerically in ascending order
+    if (!isNaN(Number(aFirstPart)) && !isNaN(Number(bFirstPart))) {
+      return Number(aFirstPart) - Number(bFirstPart)
+    }
+
+    // Fallback to alphabetical sorting if not numbers
+    return bFirstPart.localeCompare(aFirstPart)
+  })
```
mijailr, line 156: Node name is also coincidental to the SDK, it adds a number but it's not mandatory on the server side. 

Sorting should be calculated starting from the entrypoint following the outgoing edges as a directed graph.

PR #1728, opened 2025-07-09
revert(dashboard): old dashboard

bryson-g:
Checked-out from: 8d9e1658f0b0b2d1f765be714bc0d1046944c90d


dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Modals/NodeRun/WorkflowEventDefDetail.tsx
```diff
@@ -59,7 +59,7 @@ export const WorkflowEventDefDetail: FC<{ nodeRun: NodeRun }> = ({ nodeRun }) =>
 
       <div className={cn('mt-2 flex w-full flex-col overflow-auto rounded p-1', 'bg-zinc-500 text-white')}>
         <h3 className="font-bold">Content</h3>
-        <pre className="overflow-auto">{getVariableValue(data.content)}</pre>
+        <pre className="overflow-auto">{String(getVariableValue(data.content))}</pre>
```
Comment (mijailr) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Modals/NodeRun/WorkflowEventDefDetail.tsx]: Why is `String` used for all `getVarialbeValue` and `getVariable` functions?

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserAndGroupAssignmentInfo.tsx
```diff
@@ -1,13 +1,14 @@
+import { WfRunId } from 'littlehorse-client/proto'
 import { FC } from 'react'
 interface UserTaskRunDetailsProps {
-  userGroup: string | number | boolean | Buffer | undefined
-  userId: string | number | boolean | Buffer | undefined
+  userGroup: string | number | boolean | Buffer | undefined | WfRunId
```
Comment (mijailr) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserAndGroupAssignmentInfo.tsx]: Can a userGroup be a WfRunId?

Is "explicit" `undefined` required here? Can't it be solved with optional?

```suggestion
  userGroup?: string | number | boolean | Buffer | WfRunId
```

Comment (mijailr) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserAndGroupAssignmentInfo.tsx]: Please validate with @Snarr if WfRunId applies in this context.

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserTaskNotes.tsx
```diff
@@ -1,12 +1,13 @@
+import { WfRunId } from 'littlehorse-client/proto'
 import { FC } from 'react'
 interface UserTaskNotesProps {
-  notes: string | number | boolean | Buffer | undefined
+  notes: string | number | boolean | Buffer | undefined | WfRunId
```
Comment (mijailr) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserTaskNotes.tsx]: Same here, looks like this is a pattern.

local-dev/compile-proto.sh
```diff
@@ -114,7 +114,7 @@ echo "Compiling protobuf sdk-js"
 $docker_run protoc \
 	--plugin=/usr/local/lib/node_modules/ts-proto/protoc-gen-ts_proto \
 	--ts_proto_out /littlehorse/sdk-js/src/proto \
-	--ts_proto_opt=env=node,oneof=unions,outputServices=nice-grpc,outputServices=generic-definitions,outputJsonMethods=false,useExactTypes=false,eslint_disable,esModuleInterop=true,useDate=string,stringEnums=true,exportCommonSymbols=false \
+	--ts_proto_opt=env=node,outputServices=nice-grpc,outputServices=generic-definitions,outputJsonMethods=false,useExactTypes=false,eslint_disable,esModuleInterop=true,useDate=string,stringEnums=true,exportCommonSymbols=false \
```
mijailr, line 117: We may want to bring oneof unions anyway.

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Modals/NodeRun/WorkflowEventDefDetail.tsx
Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Modals/NodeRun/WorkflowEventDefDetail.tsx]: Because previously the return type of getVariableValue did not include WfRunId, but now it does since a server update changed the proto. 

WfRunId is not a data type that ReactNode supports out of the box as it is a custom type, so it must be converted to a string to display.

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserAndGroupAssignmentInfo.tsx
Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserAndGroupAssignmentInfo.tsx]: https://github.com/littlehorse-enterprises/littlehorse/blob/25966cd79793d865252ecb84c9052c9a5ff4c489/dashboard/src/app/(authenticated)/%5BtenantId%5D/(diagram)/components/NodeTypes/UserTask/UserTaskDefDetails.tsx#L11

the values passed into this prop are return types from `getVariable`, which does return this exact union. 

This is why the types must match here with undefined and WfRunId. 

userTask.userGroup and userTask.userId are both VariableAssignments.

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserTaskNotes.tsx
Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserTaskNotes.tsx]: I can do `ReturnType<typeof getValue> `instead of doing the union pattern, which is better. 

But either way, this type does need to have WfRunId and undefined since that's the return type of `getVariable`

We could forgo using the function all together, but it doesn't seem safe to access the VariableAssignment origin of the userGroup and userId props without it. Because then, we'd have to assume the fields that VariableAssignment is supposed to have, rather than using our guarded function that will handle all scenarios.

local-dev/compile-proto.sh
bryson-g: Sure we can, but I believe that should be a separate PR from reverting the changes. The scope of this is to just revert and make sure build is compatible, and to review any new changes that fix the build.

bryson-g: Adding oneof=unions will result in major refactoring

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserAndGroupAssignmentInfo.tsx
Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserAndGroupAssignmentInfo.tsx]: The props here use the getVariable function, since `...userGroup` and `...userId` are VariableAssignment, so WfRunId is included in the type here.

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Modals/NodeRun/WorkflowEventDefDetail.tsx
Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Modals/NodeRun/WorkflowEventDefDetail.tsx]: Making all returns of this function strings will fix this issue safely

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserAndGroupAssignmentInfo.tsx
Comment (Snarr) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserAndGroupAssignmentInfo.tsx]: UserId and UserGroup should always be a String I believe, I don't think there's any way for them to be stored in another format on the server. If a UserId or UserGroup is determined by a VariableAssignment, the server converts them to Strings internally.

Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/NodeTypes/UserTask/UserAndGroupAssignmentInfo.tsx]: So if it's always going to be a string, do we want to enforce that assumption on the frontend? We can restrict the prop to be a string, and if it's not a string, we'd have to return `null` and nothing will be rendered. 

Or is using the `getVariable` function preferable?

@mijailr @Snarr

local-dev/compile-proto.sh
mijailr: Understood

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Modals/NodeRun/WorkflowEventDefDetail.tsx
Comment (mijailr) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Modals/NodeRun/WorkflowEventDefDetail.tsx]: WfRunId should be returned as string. not as type.

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Modals/NodeRun/UserTaskDefDetail.tsx
```diff
@@ -0,0 +1,190 @@
...
+  const { data, isLoading } = useQuery({
+    queryKey: ['userTaskRun', wfRunId, taskId],
+    queryFn: async () => {
+      if (!wfRunId) return
+      if (!taskId) return
+      const taskRun = await getUserTaskRun({
+        wfRunId: {
+          id: wfRunId,
+          parentWfRunId: undefined,
+        },
```
Comment (mijailr) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Modals/NodeRun/UserTaskDefDetail.tsx:21]: This won't work for child workflows.

mijailr (approved):

Issue #1573, opened 2025-06-06
Register `ExternalEventDef` and `WorkflowEventDef` in Python, C#, GO


coltmcnealy-lh:
In Java, we can do:

```java
wf.waitForEvent("my-external-event").registeredAs(String.class);
```

We need to:

- [ ] Add support for `StructDef` types in the `ExternalEventDef`
- [x] Support Go
- [x] Support Python
- [x] Support C#

PR #1731, opened 2025-07-10
feat(sdk-go): Register ExternalEventDef and WorkflowEventDef

l4crito:
This PR enables workflows in the Go SDK to programmatically register external and workflow events together when the  WorkflowSpec is registeered, issue #1573 .
 Adds tests and an example demonstrating automatic event registration.


sdk-go/littlehorse/wf_lib_internal.go
```diff
@@ -1539,3 +1561,42 @@ func (t *WorkflowThread) checkIfIsActive() {
...
+func (n *ThrowEventNodeOutput) toPutWorkflowEventDefRequest() (*lhproto.PutWorkflowEventDefRequest, error) {
+	return &lhproto.PutWorkflowEventDefRequest{
+		Name:        n.eventNameDefName,
+		ContentType: &lhproto.ReturnType{ReturnType: &lhproto.TypeDefinition{Type: n.payloadType}},
+	}, nil
+}
+
+func (wf *LHWorkflow) registerWfSpec(client lhproto.LittleHorseClient) error {
+
+	putWf, _ := wf.Compile()
```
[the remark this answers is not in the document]
l4crito: The compile function is returning nill as error, think is not necesary to evaluate it

sdk-go/littlehorse/wf_lib_public.go
```diff
@@ -20,8 +20,18 @@ type LHWorkflow struct {
...
 
-	spec  lhproto.PutWfSpecRequest
-	funcs map[string]ThreadFunc
+	spec                     lhproto.PutWfSpecRequest
+	funcs                    map[string]ThreadFunc
+	workflowEventsToRegister []*ThrowEventNodeOutput
+	externalEventsToRegister []*ExternalEventNodeOutput
+}
+
+func (l *LHWorkflow) GetWorkflowEventsToRegister() []*ThrowEventNodeOutput {
```
coltmcnealy-lh: Why does this need to be public? I don't imagine that our users will need to call this.

sdk-go/littlehorse/wf_lib_public.go
```diff
@@ -20,8 +20,18 @@ type LHWorkflow struct {
...
+	return l.workflowEventsToRegister
+}
+
+func (l *LHWorkflow) GetExternalEventsToRegister() []*ExternalEventNodeOutput {
```
coltmcnealy-lh: Same comment here.

sdk-go/littlehorse/wf_lib_public.go
```diff
@@ -80,9 +90,11 @@ func (n *TaskNodeOutput) JsonPath(jsonPath string) NodeOutput {
...
 
 type ExternalEventNodeOutput struct {
-	nodeName string
-	jsonPath *string
-	thread   *WorkflowThread
+	nodeName             string
+	jsonPath             *string
+	thread               *WorkflowThread
+	externalEventDefName string
+	payloadType          lhproto.VariableType
```
coltmcnealy-lh: Hmm. Can we use a `TypeDefinition` here instead of `VariableType`?

sdk-go/littlehorse/wf_lib_public.go
```diff
@@ -93,6 +105,10 @@ func (n *ExternalEventNodeOutput) JsonPath(jsonPath string) NodeOutput {
 	}
 }
 
+func (n *ExternalEventNodeOutput) RegisteredAs(payloadType lhproto.VariableType) *ExternalEventNodeOutput {
```
coltmcnealy-lh, line 104: Same here, this should be a `TypeDefinition` not a `VariableType`.

@Snarr would you care to chime in about how this might work with `StructDef` registration?

coltmcnealy-lh (commented):
Getting closer.

sdk-go/littlehorse/wf_lib_internal.go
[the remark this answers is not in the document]
coltmcnealy-lh: That's because when I wrote this SDK I was very new to Go and didn't understand how error propagation actually works, so I threw a bunch of `panic()`'s in the code. That's not too great...eventually we do want to use proper error propagation....

@eduwercamacaro what do you think here?

coltmcnealy-lh:
Additionally, this PR is missing support for creating the `CorrelatedEventConfig` (supporting correlated events).

sdk-go/littlehorse/wf_lib_public.go
l4crito: Think there is no need for this to be public

l4crito: but need to figure out another way for unit testing this

sdk-go/littlehorse/wf_lib_public.go
l4crito: So go was weird about it , i tried to do it same that in the dotnet or python implementation
image or 
image
but in go we will end up with something like this 
image
which to me was weird, and i used the proto enum instead

sdk-go/littlehorse/wf_lib_internal.go
```diff
@@ -1182,6 +1184,25 @@ func (t *WorkflowThread) maskCorrelationId(n *ExternalEventNodeOutput, masked bo
 	return n
 }
 
+func (t *WorkflowThread) registerWorkflowEventdAs(n *ThrowEventNodeOutput, payloadType lhproto.VariableType) *ThrowEventNodeOutput {
```
coltmcnealy-lh: Can you please use `registerWorkflowEventDefAs` insstead of `WorkflowEventd`? Thanks!

sdk-go/littlehorse/wf_lib_internal.go
```diff
@@ -1182,6 +1184,25 @@ func (t *WorkflowThread) maskCorrelationId(n *ExternalEventNodeOutput, masked bo
...
+	return n
+}
+
+func (t *WorkflowThread) registerExternalEventdAs(n *ExternalEventNodeOutput, payloadType lhproto.VariableType) *ExternalEventNodeOutput {
+	t.checkIfIsActive()
+	n.payloadType = &payloadType
+	t.wf.externalEventsToRegister = append(t.wf.externalEventsToRegister, n)
+	return n
+}
+func (t *WorkflowThread) registerExternalEventAsEmpty(n *ExternalEventNodeOutput) *ExternalEventNodeOutput {
```
coltmcnealy-lh: Same comment here. Should be `registerExternalEventDefAsEmpty`

PR #1739, opened 2025-07-16
fix(server): properly configure flush threads

coltmcnealy-lh:
- Introduces LHS_ROCKSDB_FLUSH_THREADS which defaults to value of LHS_ROCKSDB_COMPACTION_THREADS
- Properly uses the RocksDB Env object to configure the value of the low (compaction) and high (flush) priority thread pools.
- Fixes the issue where options.setIncreaseParallelism() actually set the flush threads to only 1 whereas we thought it set flush AND compaction to the same value.


coltmcnealy-lh:
Closing in favor of other work that is being done in collaboration with @eduwercamacaro

Issue #1740, opened 2025-07-16
Command producer's network thread gets block when executing remote wait for commands


eduwercamacaro:
Thread dump:
```
Name: kafka-producer-network-thread | lh-0
State: WAITING on io.grpc.stub.ClientCalls$ThreadlessExecutor@3613b1b5
Total blocked: 40  Total waited: 283,022

Stack trace: 
java.base@21.0.7/jdk.internal.misc.Unsafe.park(Native Method)
java.base@21.0.7/java.util.concurrent.locks.LockSupport.park(LockSupport.java:221)
app//io.grpc.stub.ClientCalls$ThreadlessExecutor.waitAndDrain(ClientCalls.java:752)
app//io.grpc.stub.ClientCalls.blockingUnaryCall(ClientCalls.java:161)
app//io.littlehorse.common.proto.LHInternalsGrpc$LHInternalsBlockingStub.waitForCommand(LHInternalsGrpc.java:358)
app//io.littlehorse.server.streams.CommandSender.waitForCommand(CommandSender.java:166)
app//io.littlehorse.server.streams.CommandSender.lambda$doSend$0(CommandSender.java:85)
app//io.littlehorse.server.streams.CommandSender$$Lambda/0x00007f1bd7781410.apply(Unknown Source)
java.base@21.0.7/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1150)
java.base@21.0.7/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
java.base@21.0.7/java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2179)
app//io.littlehorse.common.util.LHProducer$CompletableFutureCallback.onCompletion(LHProducer.java:64)
app//org.apache.kafka.clients.producer.KafkaProducer$AppendCallbacks.onCompletion(KafkaProducer.java:1566)
app//org.apache.kafka.clients.producer.internals.ProducerBatch.completeFutureAndFireCallbacks(ProducerBatch.java:309)
app//org.apache.kafka.clients.producer.internals.ProducerBatch.done(ProducerBatch.java:273)
app//org.apache.kafka.clients.producer.internals.ProducerBatch.complete(ProducerBatch.java:219)
app//org.apache.kafka.clients.producer.internals.Sender.completeBatch(Sender.java:754)
app//org.apache.kafka.clients.producer.internals.Sender.completeBatch(Sender.java:725)
[...]

PR #1741, opened 2025-07-16
fix(server): Async remote wait for command

eduwercamacaro:
Solves #1740

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streams/CommandSender.java
+++ b/server/src/main/java/io/littlehorse/server/streams/CommandSender.java
@@ -5,6 +5,7 @@ import com.google.protobuf.ByteString;
 import com.google.protobuf.Empty;
 import com.google.protobuf.Message;
 import io.grpc.Status;
+import io.grpc.StatusRuntimeException;
 import io.grpc.stub.StreamObserver;
 import io.littlehorse.common.LHServerConfig;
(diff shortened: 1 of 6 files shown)


server/src/main/java/io/littlehorse/server/streams/CommandSender.java
```diff
@@ -160,10 +164,24 @@ private CompletableFuture<Message> waitForCommand(
...
+            ListenableFuture<WaitForCommandResponse> futureResponse = internalClient.waitForCommand(req);
+            CompletableFuture<Message> out = new CompletableFuture<>();
+            futureResponse.addListener(
+                    () -> {
+                        try {
+                            WaitForCommandResponse response = futureResponse.get();
+                            out.complete(buildRespFromBytes(response.getResult(), responseCls));
+                        } catch (ExecutionException e) {
+                            out.completeExceptionally(e.getCause());
+                        } catch (InterruptedException e) {
```
coltmcnealy-lh, line 178: Doesn't `InterruptedException` occur during shutdown?

eduwercamacaro: The listenableFuture is already done by the time this listener is called. So, I don't think it is possible to have an interrupted exception at this point. I covered the case because it is part of the `futureResponse.get()` API.

PR #1743, opened 2025-07-16
fix(dashboard): correct entrypoint accessing

bryson-g:


mijailr:
I know this is not part of the scope but we should take the chance to fix some things in this file:

- Verify the `?` checks, is possible to get rid of them if we ensure data is present.
- Rename `lhWorkflowSpec` to `wfSpec` to keep consistency.
- The push to path `router.push(`/${tenantId}/wfRun/${wfRun.id.id}`) ` won't work if the workflow is a child workflow. I'm not sure if this works for child workflows though, please check.

bryson-g:
> I know this is not part of the scope but we should take the chance to fix some things in this file:
> 
> * Verify the `?` checks, is possible to get rid of them if we ensure data is present.
> * Rename `lhWorkflowSpec` to `wfSpec` to keep consistency.
> * The push to path `/${tenantId}/wfRun/${wfRun.id.id} ` won't work if the workflow is a child workflow. I'm not sure if this works for child workflows though, please check. I suppose the correct way is `/${tenantId}/wfRun/${concatWfRunIds(wfRun.id)}`

I'll make two issues for refactoring those items you mentioned because they're applicable for the entire project. 
I'll fix those in the files that I touch in regards to this PR for now.

I'll test the workflow under some example circumstances and I'll get back to you on it.

mijailr (approved):

PR #1753, opened 2025-07-17
fix(server): Wait for task claim event when sending response back to the client observer

eduwercamacaro:
Currently, the task protocol only waits for the task claim command to be acknowledged in the kafka broker, but this approach causes some duplicated commands during exceptional cases like a rebalance or a rolling restart. 

This PR aims to wait for the Task Claim command to be processed before sending the ScheduledTask to the client observer. This approach adds an extra latency to the task protocol, but it improves correctness and reduces the possibility of sending the same task multiple times to the client observer. 

Latency depends on the current lag on the command topic.

During low traffic, it is expected to have at least 2 ms extra latency, but this latency increases with the process latency of each stream thread.

### DIFF
--- a/lhctl/internal/run.go
+++ b/lhctl/internal/run.go
@@ -56,9 +56,7 @@ lhctl run my_workflow_id foo '{"bar":"baz"}'
 
 		parentWfRunId, _ := cmd.Flags().GetString("parentWfRunId")
 		if parentWfRunId != "" {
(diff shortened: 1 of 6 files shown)


coltmcnealy-lh (approved):
Looks good, but I'm not a `CompletableFuture` expert. Thanks! This is important for correctness.

PR #1755, opened 2025-07-20
chore(server): Set ZGC as default java garbage collector

eduwercamacaro:
This PR aims to:
- Use ZGC as a default jvm argument for Lh-server. ZGC fits better in the context of littlehorse server as it extensively uses heap memory when processing gRPC requests.
- Reduce the size of docker image by using different layers for dependencies and application binaries.

### DIFF
--- a/docker/server/Dockerfile
+++ b/docker/server/Dockerfile
@@ -3,6 +3,7 @@ WORKDIR /lh
 RUN amazon-linux-extras install epel -y && yum install jemalloc -y
 ENV LD_PRELOAD=/usr/lib64/libjemalloc.so.1
 COPY ./docker/server/docker-entrypoint.sh /lh
-COPY ./server/build/libs/server-*-all.jar /lh/server.jar
+COPY ./server/build/install/server/server /lh/bin
+COPY ./server/build/install/server/lib /lh/lib
 ENTRYPOINT ["/lh/docker-entrypoint.sh"]
 CMD ["server"]
--- a/docker/server/docker-entrypoint.sh
+++ b/docker/server/docker-entrypoint.sh
@@ -4,7 +4,7 @@ set -e
 
 if [ "$1" = 'server' ]; then
     shift
-    exec java $JAVA_OPTS -jar /lh/server.jar "$@"
+    exec /lh/bin/server
 fi
 
 exec "$@"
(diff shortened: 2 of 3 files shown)


docker/server/docker-entrypoint.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 if [ "$1" = 'server' ]; then
     shift
-    exec java $JAVA_OPTS -jar /lh/server.jar "$@"
```
coltmcnealy-lh, line 7: Hmm. This change removes the ability of users to specify a config file with properties. It requries them to use env vars. Can we keep the `"$@"`?

coltmcnealy-lh:
```java
    public static void main(String[] args) throws IOException, InterruptedException, ExecutionException {
        LHServerConfig config;

        if (args.length > 0) {
            log.info("Loading configuration from file '{}'", args[0]);
            config = new LHServerConfig(args[0]);
        } else {
            log.info("Loading configuration from env variables");
            config = new LHServerConfig();
        }

```
That's the LH Server entrypoint.

docker/server/Dockerfile
```diff
@@ -3,6 +3,7 @@ WORKDIR /lh
 RUN amazon-linux-extras install epel -y && yum install jemalloc -y
 ENV LD_PRELOAD=/usr/lib64/libjemalloc.so.1
 COPY ./docker/server/docker-entrypoint.sh /lh
-COPY ./server/build/libs/server-*-all.jar /lh/server.jar
+COPY ./server/build/install/server/server /lh/bin
+COPY ./server/build/install/server/lib /lh/lib
```
coltmcnealy-lh, line 6: Suggestion: Can we reverse the order of these two lines? For local builds, that would make it faster if you just change the code without adding a new dependency, because then you can cache the `lib` layer.

docker/server/docker-entrypoint.sh
eduwercamacaro: Good catch. Thanks!

docker/server/Dockerfile
eduwercamacaro: Yes, thank you! 🧠

PR #1756, opened 2025-07-21
fix(pipeline): Publish docker pipeline

eduwercamacaro:


sauljabin (approved):

.github/workflows/publish-docker.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         run: sed -i "s/version=.*/version=${{ inputs.tag }}/g" gradle.properties
 
       - name: Tests and Build
-        run: ./gradlew server:test server:shadowJar
+        run: ./gradlew server:test server:shadowJar server:installDist
```
coltmcnealy-lh, line 31: Do we still need the shadow jar?

coltmcnealy-lh (approved):

.github/workflows/publish-docker.yml
eduwercamacaro: We publish a shadow jar as part of the release. I think we need it.

Issue #725, opened 2024-03-17
Add `parent_wf_run_id` to `rpc SearchWfRun`


coltmcnealy-lh:
We should add a field:
```
optional WfRunId parent_wf_run_id = ?;
```

to the `message SearchWfRunRequest`. This will allow users to restrict the results from an `rpc SearchWfRun` to only those that are children of a specific `WfRun`.

As for implementation, here are some notes:

* We currently don't have a query planner, because we do not have Counted Tags.
* It is _most likely_ that the most efficient way to do this would be:
  * Object Id Prefix Scan based on parent wfrun id
  * Execute other filters, such as status/variables/wfSpecName, while iterating over children.

In the long term, we need to optimize this. How?
* Implement `COUNTED` tags, which allow us to know:
  * How many children does a `WfRun` have?
  * How many `WfRun`'s of a certain `WfSpec` have a given status?
  * How many `WfRun`'s have a variable matching a certain value?
* Choose the dimension with the fewest elements, and iterate over that dimension first.

coltmcnealy-lh:
@diablouma had requested this for the control plane. We won't add it quite yet...but the short-term implementation is no more than a half day of work.

Maybe Nelson or Mateo could take this when time permits? Alternatively, it might be a good starter issue for our new hire.

coltmcnealy-lh:
This was closed by #1768

PR #1768, opened 2025-07-24
feat(server): Add support for searching WfRuns by parent WfRun ID 

l4crito:
### Summary

This PR extends the `SearchWfRun` RPC and CLI to support filtering by **parent WfRun ID**, enabling efficient queries for child workflows in parent-child workflow hierarchies. Issue #725

### Server-side Changes

- Enhanced `WfRunModel` with conditional indexing for parent-child relationships:
  - `wfSpecName + parentWfRunId` index for direct parent-based searches.
  - `wfSpecName + status + parentWfRunId` composite index for filtered searches.

### CLI Changes

- Updated `lhctl search wfRun` command to accept the `--parentWfRunId` flag.

### Example Usage

```bash
lhctl search wfRun child --parentWfRunId my-parent-wf 
lhctl search wfRun child --status COMPLETED --parentWfRunId my-parent-wf

### CODE REVIEW

File: examples/child-workflow/src/main/java/io/littlehorse/examples/ChildWorkflowExample.java [hunk 1]
```diff
@@ -17,6 +17,17 @@
 
 public class ChildWorkflowExample {
 
+    public static Workflow getGrandChild() {
```
eduwercamacaro, line 20: If this change is intended, we need to update the README file.

lhctl/internal/wf_run.go
```diff
@@ -37,25 +38,31 @@ var getScheduledWfRun = &cobra.Command{
 }
 
 var searchWfRunCmd = &cobra.Command{
-	Use:   "wfRun <wfSpecName> [<majorVersion>] [<revision>]",
+	Use:   "wfRun [<wfSpecName>] [<majorVersion>] [<revision>]",
 	Short: "Search for WfRuns",
 	Long: `
 Search for WfRuns. You may provide the optional arguments:
+- [<wfSpecName>] (required unless --parentWfRunId is specified)
```
eduwercamacaro: I'm wondering if Cobra has any rules that validate this for us. Also, I'm not sure if this is the correct approach to handle this type of queries. What if we create another subcommand under `wfRun`? 

I think we need @Snarr's opinion here :)

examples/child-workflow/src/main/java/io/littlehorse/examples/ChildWorkflowExample.java
l4crito: Yes this was intended , i will update the README

lhctl/internal/wf_run.go
l4crito: So you mean we leave previous flags as it where, and we do something like this ? 

`lhctl search wfRun byParent parent-id --show-full-tree`
`lhctl search wfRun byParent parent-id --wfSpecName child-workflow`

l4crito: I think it make a lot of sense since we have flags that will be only available only when you search by parent-id

coltmcnealy-lh (approved):

PR #1771, opened 2025-07-28
refactor(proto): rename NodeRun wait_threads to wait_for_threads

mijailr:
this is no match the contract on Node


coltmcnealy-lh (approved):
You're lucky to have changed the `NodeRun` not the `Node` since changing the `Node` would have required changing the code in all 4 sdk's :wink:

PR #1774, opened 2025-07-29
fix(server): Properly check type definitions

eduwercamacaro:


server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/TypeDefinitionModel.java
```diff
@@ -134,7 +140,7 @@ public boolean isCompatibleWith(TypeDefinitionModel other) {
         if (type == VariableType.INT || type == VariableType.DOUBLE) {
             return other.getType() == VariableType.INT || other.getType() == VariableType.DOUBLE;
         }
-        return this.equals(other);
+        return this.getType().equals(other.getType());
```
coltmcnealy-lh, line 143: Good catch.

coltmcnealy-lh (approved):

PR #1776, opened 2025-07-31
feat(canary): Canary metronome inflight wfruns

eduwercamacaro:
This PR proposes rejecting `runWf` requests when the number of **pending requests** is equal to the number of requests that the canary metronome sends on every cycle. The goal is to avoid sending more requests than the Littlehorse server can process. 

Canary metronome schedules `runWf` requests at fixed rate (controlled by `lh.canary.metronome.run.frequency.ms` config), and every single request might wait for long periods waiting for a response from the server. If the server is unhealthy (rebalancing or any other scenario that prevents the requests from being processed), the number of pending requests will be kept in memory until the server is able to process the requests. These scenarios are not ideal for both the canary and the server because the number of pending requests will uncontrollably grow, increasing the chances of reaching an OOM error.

So, ideally in these scenarios the canary metronome should wait for a response before sending a new request.


hs_err_pid16006.log
l4crito, line 1: May be get rid of the log

l4crito (approved):

hs_err_pid16006.log
sauljabin: I was going to say the same, why this log file?

canary/src/main/java/io/littlehorse/canary/metronome/MetronomeRunWfExecutor.java
```diff
@@ -90,7 +95,8 @@ private void executeRun(final boolean isSampleIteration) {
 
         log.debug("Executing run {}", wfId);
 
-        final ListenableFuture<WfRun> future = lhClient.runCanaryWf(wfId, start, isSampleIteration);
+        final Deadline deadline = !inflightRequests.tryAcquire() ? expiredDeadline : null;
```
sauljabin, line 98: so, we are not interested on blocking, just to set (or not)  the deadline

sauljabin (approved):

canary/src/main/java/io/littlehorse/canary/metronome/MetronomeRunWfExecutor.java
eduwercamacaro: Yes, I believe it is better to not block the current iteration. Instead, the deadline will cause an immediate deadline_exceeded status code.

coltmcnealy-lh (approved):

PR #1779, opened 2025-08-04
proposal: extending-external-event-validations

JakeRose03:


proposals/006-extending-external-event-validations.md
```diff
@@ -0,0 +1,81 @@
...
+  // WfSpec must reference the external event being posted
+  optional boolean  require_wfspec_reference = 5;
+
+}
+```
+### Compatibility
+
+Keep in mind that the current default functionality among all clients is that a premature external event can trigger advances and interrupts. Also we do not check to see if a `wfSpec` references an external event giving events the ability to hang.
+
+In order to provide backwards compatibility the default values must not require a `wfRun` to exist, as well as not require a `wfSpec` to hold reference to an external event.
```
coltmcnealy-lh, line 72: Well done

proposals/006-extending-external-event-validations.md
```diff
@@ -0,0 +1,81 @@
...
+
+I chose flags over enums to represent these two functionalities because I believe that they are two separate binary operations where extensibility is not necessary.
```
Snarr, line 74: Just a note: I think extensibility is still possible when using booleans, as you can just create an enum where `FALSE_STATE = 0` and `TRUE_TRUE = 1` and the protobuf will serialize it all the same. Just an interesting tidbit

PR #1783, opened 2025-08-07
docs(proposal): Add Arrays to Proposal #000

Snarr:


proposals/000-struct-and-structdef.md
```diff
@@ -57,6 +57,53 @@ message Struct {
...
+
+  message StructArray {
+    repeated Struct items = 1;
+  }
+
+  message ArrayArray {
+    repeated Array items = 1;
+  }
+
+  oneof value {
```
Snarr, line 94: So this is tricky; you can't have repeated fields in `oneof`s, so the next best thing is wrapping the repeated fields in their own respective objects. Let me know your thoughts on this!

coltmcnealy-lh (approved):

PR #1784, opened 2025-08-08
feat(server): Validation policies for external events

JakeRose03:


schemas/littlehorse/service.proto
```diff
@@ -490,6 +490,10 @@ message PutExternalEventDefRequest {
 
   // If set, then this `ExternalEventDef` will allow the `CorrelatedEvent` feature.
   optional CorrelatedEventConfig correlated_event_config = 4;
+
+  // Policies that represent when an external event can be posted
```
eduwercamacaro: I think this should be singular:

maybe like this
```
Policy that represents when an external event can be posted
```

server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/PutExternalEventRequestModel.java
```diff
@@ -91,6 +97,53 @@ public ExternalEvent process(CoreProcessorContext executionContext, LHServerConf
             }
         }
 
+        WfRunModel wfRun = getableManager.get(wfRunId);
```
eduwercamacaro, line 97: I believe these checks should be implemented in the wfspec, or at least we should get all the external events defined in the wfspec (via `wfSpec.getNodeExternalEventDefs()`) and use those strings to verify whether the external event is present in the wfspec.

Also, that will eliminate the need for labels in the logic, which usually introduces complexity to the code.

eduwercamacaro:
```java
ExternalEventDefModel interruptExternalEventDefModel = service.getExternalEventDef(
                                interrupt.getExternalEventDefId().getName());
```
Moreover, We don't actually need to query the store to compare two external event defs. You can use the external event def ids instead.

PR #1785, opened 2025-08-08
feat(dashboard): WfRun ids shown in fixed width font

anomalizer:
Run-ids are fixed width strings but the dashboard renders it in proportional fonts. Using fixed width fonts makes the table more readable


coltmcnealy-lh:
Thank you for the contribution! Could @mijailr or @hazimoarafa take a look here?

mijailr:
Thanks for your contribution @anomalizer. I've checked and unfortunately this isn't producing the expected result, as you can see in the attached image the font is still variable width causing different char widths. 

image


Although if you replace the class name `font-code` to `font-mono` will provide the expected results.


image


I see how this will benefit the WfRuns that have the same length `WfRunId` although since wfRunIds can be arbitrary, we should also leverage on `flex` to grow the wfRunId` section in a way that the `Status` and `start date` gets fixed at the end of each row.

anomalizer:
Review comment 

> See comment above.

This is taken care of @mijailr

PR #1786, opened 2025-08-08
fix(server): npe handling putwfspecrequest with invalid nodeoutputref…

coltmcnealy-lh:
…erence

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/VariableAssignmentModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/VariableAssignmentModel.java
@@ -9,6 +9,7 @@ import io.littlehorse.common.model.getable.core.variable.VariableValueModel;
 import io.littlehorse.common.model.getable.global.wfspec.ReturnTypeModel;
 import io.littlehorse.common.model.getable.global.wfspec.TypeDefinitionModel;
 import io.littlehorse.common.model.getable.global.wfspec.WfSpecModel;
+import io.littlehorse.common.model.getable.global.wfspec.node.NodeModel;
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/VariableAssignmentModel.java
```diff
@@ -143,9 +144,12 @@ public Optional<TypeDefinitionModel> resolveType(
                 return Optional.of(new TypeDefinitionModel(VariableType.STR));
             case NODE_OUTPUT:
                 // TODO: handle here if nodeOutputType is a STRUCT and we access a field on it.
-                Optional<ReturnTypeModel> returnTypeOption = wfSpec.fetchThreadSpec(threadSpecName)
-                        .getNode(nodeOutputReference.getNodeName())
-                        .getOutputType(manager);
+                NodeModel node = wfSpec.fetchThreadSpec(threadSpecName).getNode(nodeOutputReference.getNodeName());
+                if (node == null) {
+                    throw new InvalidExpressionException("Node " + nodeOutputReference.getNodeName()
```
Snarr, line 149: Did we agree to use InvalidExpressionException here?

coltmcnealy-lh: Yes we did. It ends up propagatiing to the user as showing the correct node is invalid:

```
INVALID_ARGUMENT: PutWfSpecRequest is invalid: ThreadSpec entrypoint invalid: Node 2-validations-accept-int-TASK invalid: Task input variable with name arg0 at position 0 could not resolve type: Node not-a-real-node not present in threadspec entrypoint
```

PR #1789, opened 2025-08-12
fix(server): Mark wfrun as halted when processing stop wfrun requests

eduwercamacaro:
This PR fixes a problem when stoping wfrun requests, and makes UserTaskNodeRun stoppable.

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
@@ -362,7 +362,6 @@ public class ThreadRunModel extends LHSerializable<ThreadRun> {
         pi.externalEventId = trigger.getObjectId();
         pi.interruptedThreadId = number;
         pi.handlerSpecName = idef.handlerSpecName;
-
         wfRun.pendingInterrupts.add(pi);
     }
 
@@ -457,7 +456,10 @@ public class ThreadRunModel extends LHSerializable<ThreadRun> {
         }
 
         getCurrentNodeRun().maybeHalt(processorContext);
-        maybeFinishHaltingProcess();
+        boolean halted = maybeFinishHaltingProcess();
+        if (number == 0 && halted && !reason.isTransitioningHaltState()) {
+            wfRun.transitionTo(LHStatus.HALTED);
+        }
     }
 
     public void setStatus(LHStatus status) {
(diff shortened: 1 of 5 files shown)


server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
```diff
@@ -457,7 +456,10 @@ public void halt(ThreadHaltReasonModel reason) {
         }
 
         getCurrentNodeRun().maybeHalt(processorContext);
-        maybeFinishHaltingProcess();
+        boolean halted = maybeFinishHaltingProcess();
+        if (number == 0 && halted && !reason.isTransitioningHaltState()) {
```
coltmcnealy-lh, line 467: After reading the `isTransitioningHaltState()` method, I think this line generally says:
> If the entrypoint `ThreadRun` is halted because it is handling a failure or an interrupt, then the `WfRun` remains `RUNNING`.

Note that the only other two halt reasons are:
* `MANUAL_HALT`
* `PARENT_HALTED`.

(I am guessing that you just forgot about `PENDING_FAILURE`...?).

Note that `PARENT_HALTED` is not possible for the entrypoint (no parents).

I think the only time we want to make the `WfRun` be `HALTED` is if the reason is `MANUAL_HALT`. Can we just change this condition to be:

```
    if (number == 0 && halted && reason.getType() == ReasonCase.MANUAL_HALT)
```
?

coltmcnealy-lh (approved):
One comment. Also, it would be good to have a test (unit or e2e) which verifies that the WfRun does indeed go to HALTED when needed

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
eduwercamacaro: > If the entrypoint ThreadRun is halted because it is handling a failure or an interrupt, then the WfRun remains RUNNING.

Yes, that's exactly right. IMO, it is clearer in that way instead of just checking `MANUAL_HALT` reason. In practice, they are the same, but I think this ways is better for clarity and maintainability. 


> (I am guessing that you just forgot about PENDING_FAILURE...?).

Did I? Ok, I'll add it thanks :D

eduwercamacaro:
> Regarding making a UserTaskRun stoppable, how do we un-stop it?

That's a really good question. I guess we will need to handle this in the rescue request.

PR #1796, opened 2025-08-19
feat(server): Add Structs

Snarr:


examples/java/struct-def/src/main/java/io/littlehorse/examples/Car.java
```diff
@@ -0,0 +1,36 @@
+package io.littlehorse.examples;
+
+import io.littlehorse.sdk.worker.LHStructDef;
+
+@LHStructDef(name = "car")
+public class Car {
+    public String name;
```
l4crito: Could this be private or is there an issue?, and related to this question, could we have annotations like @Data from Lombok for the getters and setters

examples/java/struct-def/src/main/java/io/littlehorse/examples/Person.java
```diff
@@ -0,0 +1,27 @@
+package io.littlehorse.examples;
+
+import io.littlehorse.sdk.worker.LHStructDef;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
```
l4crito, line 8: May be we could have 1  example class with @Getter @Setter and @AllArgsConstructor and another with getters setters and constructors done by hand

examples/java/struct-def/src/main/java/io/littlehorse/examples/StructDefExample.java
```diff
@@ -0,0 +1,128 @@
...
+        LHConfig config = new LHConfig(props);
+
+        Workflow workflow = getWorkflow();
+
+        // New worker
+        List<LHTaskWorker> workers = getTaskWorkers(config);
+
+        // Register tasks if they don't exist
+        for (LHTaskWorker worker : workers) {
+            worker.registerStructDefs(StructDefCompatibilityType.NO_SCHEMA_UPDATES);
```
l4crito, line 93: Maybe we should allow individual schema compatibility registration somehow, here we are applying the same compatibility to all the structDefs in this example, and also quick question, once the compatibility is set is it posible to change it ?

PR #1799, opened 2025-08-20
feat(examples): consolidate examples in the same dir

l4crito:
# Consolidate and Standardize Example Projects for All SDKs

## Overview
- Moved all example projects under a unified `examples` directory at the repository root.  
- Tested all examples in their new locations. Import paths and references were fixed as needed.  

### .NET Example Changes
- Added new solution file `Examples.sln` containing only the example projects.  
- The SDK solution (`Littlehorse.sln`) now contains only SDK and test code (no examples).  

---

### Python
| Example        | Error/Reason                                                                 |
|----------------|-------------------------------------------------------------------------------|
| child_workflow | Inherited variable `input-name` is defined as **PRIVATE** in parent `WfSpec`. |

---

### Java
-  None.  

---

### .NET (dotnet)
-  None.  

---

### Go
| Example              | Error/Reason                                   |
|-----------------------|-----------------------------------------------|
| childthread           | Deploy error: Cannot use a `JSON_OBJ` as an `INT` |
| conditionals/while    | Deploy error: Cannot use a `STR` as an `INT`  |
| jsonarray             | Deploy error: expects `STR`, got `JSON_ARR`   |
| jsonobj               | Deploy error: expects `STR`, got `JSON_OBJ`   |
| taskmetadata          | Deploy error: expects `STR`, got `JSON_OBJ`   |

These failures appear to be related to **strong typing** we should address them in a different pr.


examples/CLEANUP_README.md
```diff
@@ -0,0 +1,13 @@
+# Cleaning Up the `examples` Directory
```
coltmcnealy-lh: Not sure we need to commit this into the main repo. I think it's a useful thing to comment on the PR and to send in the Community Slack #random channel though.

PR #1806, opened 2025-08-28
feat(server): Add supporf for casting on primitive types

l4crito:
## Summary
This PR introduces comprehensive support for type casting in LittleHorse workflows, including both automatic and manual casting, improved validation, and enhanced SDK usability. It also adds a new Java casting example.

## Server-side Changes
- Added support for type casting in workflow variable assignments and task inputs.  
- Implemented automatic casting from any primitive type to string, and from int to double.  
- Manual casting is now required for other type conversions (e.g., string → int, double → int, string → bool).  

## SDK & Proto Changes
- The Java SDK now provides methods for explicit casting, such as:  
  - `.cast(VariableType)`  
  - `.castToInt()`  
  - `.castToDouble()`  
  - `.castToBool()`  
  on workflow variables and node outputs.  
**Example usage:**
```java
WfRunVariable stringInput = wf.declareStr("string-number").withDefault("3.14");
NodeOutput doubleResult = wf.execute("double-method", stringInput.cast(VariableType.DOUBLE));
NodeOutput intResult = wf.execute("int-method", doubleResult.castToInt());
```
- Updated `VariableAssignment` in `common_wfspec.proto` to add a `target_type` field, enabling explicit type casting in workflow definitions.  

## Reference
For design details and rationale, see the proposal: **008-type-casting.md**


examples/java/casting/src/main/java/io/littlehorse/examples/CastingExample.java
```diff
@@ -0,0 +1,81 @@
...
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+
+public class CastingExample {
+
+    public static Workflow getWorkflow() {
+        return new WorkflowImpl("casting-workflow", wf -> {
+            WfRunVariable stringInput = wf.declareStr("string-number").withDefault("3.14");
```
eduwercamacaro, line 21: Can we add an error handler that handles a cast error? I'd like to see how to handle this casting errors when it is not possible to cast a value into another type.

l4crito: Sure!!, right now casting a Double into an Bool is not possible, that will fail when we try to register the wfSpec, but for cases like casting a String into a Bool,  i will add examples of exception handlers since it will fail at runtime

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/NodeOutput.java
```diff
@@ -18,4 +20,71 @@ public interface NodeOutput extends LHExpression {
...
+     * This enables manual type conversions that are not automatic, such as:
+     * - STR → INT/DOUBLE/BOOL (manual casting required)
+     * - DOUBLE → INT (manual casting required)
+     *
+     * The cast operation is non-mutating: the original node output remains unchanged.
+     *
+     * @param targetType the type to cast this node output to
+     * @return a new LHExpression representing the cast value
+     */
+    LHExpression cast(VariableType targetType);
```
coltmcnealy-lh: Should this be added on `LHExpression` rather than on the `NodeOutput`? Or is all of the necessary casting possible to do _before_ we do the expression logic?

l4crito: If we put it int he lh expresion we could chain cast().cast().cast(), and i am not totally sure how that should behave, but for consistency it makes a lot of sense to put it there

coltmcnealy-lh: `.cast().cast().cast()` is supported in Java... for example:

```
int bar = (int) (((double) ((int) foo)));
```

TLDR: I don't think the `.cast().cast()` is _bad_...but it does allow other things which are needed

coltmcnealy-lh (approved):

PR #1807, opened 2025-08-29
fix(server): Proper handle task already claimed case

eduwercamacaro:
Properly handle cases where a task gets claimed twice during a rebalance.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streams/taskqueue/PollTaskRequestObserver.java
+++ b/server/src/main/java/io/littlehorse/server/streams/taskqueue/PollTaskRequestObserver.java
@@ -113,8 +113,16 @@ public class PollTaskRequestObserver implements StreamObserver<PollTaskRequest> 
     }
 
     public void sendResponse(ScheduledTaskModel toExecute) {
-        PollTaskResponse response =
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/littlehorse/server/streams/taskqueue/PollTaskRequestObserver.java
```diff
@@ -113,8 +113,16 @@ RequestExecutionContext getFreshExecutionContext() {
     }
 
     public void sendResponse(ScheduledTaskModel toExecute) {
-        PollTaskResponse response =
-                PollTaskResponse.newBuilder().setResult(toExecute.toProto()).build();
-        responseObserver.onNext(response);
+        if (toExecute == null) {
+            // Tasks may be already claimed by other workers during rebalancing.
+            // TODO: Add metrics to track the frequency of empty scheduled tasks to better monitor rebalancing impact
+            log.warn("Processing pollTaskRequest for task that was already claimed");
```
coltmcnealy-lh, line 119: This could get quite chatty. I think maybe `debug` and open a GH ticket to add a metric here?

PR #1808, opened 2025-08-29
feat(server): run child wf synchronously in workflow

coltmcnealy-lh:
- This is the first checkpoint, taken at the end of the youtube video


server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/RunChildWfNodeRunModel.java
```diff
@@ -0,0 +1,109 @@
...
+
+            throw new NodeFailureException(failure);
+        }
+
+        WfSpecModel childSpec = ctx.service().getWfSpec(runWfNode.getWfSpecName(), runWfNode.getMajorVersion(), null);
+
+        this.childWfRunId =
+                new WfRunIdModel(LHUtil.generateGuid(), nodeRun.getId().getWfRunId());
+
+        WfRunModel out = new WfRunModel(ctx);
```
eduwercamacaro, line 81: How crazy would be to create a `RunWfRequestModel` and execute it?

Just thinking in the possibility of reusing the code existing in that command.

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/RunChildWfNodeRunModel.java
```diff
@@ -0,0 +1,109 @@
...
+        out.transitionTo(LHStatus.RUNNING);
+
+        out.startThread(
+                childSpec.getEntrypointThreadName(),
+                ctx.currentCommand().getTime(),
+                null,
+                inputs,
+                ThreadType.ENTRYPOINT);
+        ctx.getableManager().put(out);
+        out.advance(time);
```
eduwercamacaro, line 96: `out.advance` can throw with a node failure when the first node after the entry point fails. am I right? If so, I think we should handle that.

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/RunChildWfNodeRunModel.java
```diff
@@ -0,0 +1,109 @@
...
+            failure.message = "Failed constructing input variables for child workflow: " + exn.getMessage();
+            failure.failureName = LHConstants.VAR_SUB_ERROR;
+
+            throw new NodeFailureException(failure);
+        }
+
+        WfSpecModel childSpec = ctx.service().getWfSpec(runWfNode.getWfSpecName(), runWfNode.getMajorVersion(), null);
+
+        this.childWfRunId =
+                new WfRunIdModel(LHUtil.generateGuid(), nodeRun.getId().getWfRunId());
```
eduwercamacaro, line 79: I don't think the proposal mentions this, but I think would be important to allow clients to specify a child wfrun id. Like for example, when creating data workflows that are logically related.

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/RunChildWfNodeRunModel.java
```diff
@@ -0,0 +1,109 @@
...
+
+        WfRunModel out = new WfRunModel(ctx);
+        out.setId(childWfRunId);
```
eduwercamacaro, line 82: I think we need to validate that the Id does not exist.

eduwercamacaro: In case we accept this comment: https://github.com/littlehorse-enterprises/littlehorse/pull/1808/files#r2433108643

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/RunChildWfNodeRunModel.java
coltmcnealy-lh: I think I moved some of this code to `WfSpecModel` and called that, which is much cleaner (and reduces the amount of code in `RunWfRequestModel`) but it died in my git stash. Going to try to resurrect it. Or maybe can that be a separate cleanup PR?

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/RunChildWfNodeRunModel.java
coltmcnealy-lh: If you care about the result of the child `WfRun` you can wait for it with a `waitForChildWf()`. What is special about node 1 compared to node 2 or 3 or 4?

We can add additional validations to the _creation_ of a `WfRun` but that would affect all types of `WfRun`'s so it should be a separate proposal. This can reduce the amount of cases where the first `NodeRun` fails.

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/RunChildWfNodeRunModel.java
coltmcnealy-lh: Oh. Yes, that can be a follow-up feature. I'll add a ticket to it.

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/RunChildWfNodeRunModel.java
coltmcnealy-lh: We can do that as part of the follow-up ticket for allowing to specify the `WfRunId`.

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/WfRunModel.java
```diff
@@ -690,6 +690,17 @@ public void handleThreadStatus(int threadRunNumber, Date time, LHStatus newStatu
...
         }
 
+        if (endTime != null) {
+            // wake up parent if parent exists
+            if (id.getParentWfRunId() != null) {
+                WfRunModel parent = executionContext
+                        .castOnSupport(CoreProcessorContext.class)
+                        .getableManager()
+                        .get(id.getParentWfRunId());
+                parent.advance(time);
```
eduwercamacaro: I think this query can return `null` if the parent was deleted by relation or other manual operations. (maybe bulk deletes?)

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/RunChildWfNodeRunModel.java
eduwercamacaro: NVM, I `advance` does not return any exception. So It is fine to call this method without a try/catch

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/WfRunModel.java
coltmcnealy-lh: Fixed :smile:

lhctl/internal/wf_spec.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 var getWfSpecCmd = &cobra.Command{
-	Use:   "wfSpec <name> [<majorVersion>] [<revision>]",
+	Use:   "wfSpec <name>",
```
Snarr, line 17: Good catch, these shouldn't use positional args until they are updated to. That can be a separate PR

examples/java/run-child-workflow/src/main/java/io/littlehorse/examples/ChildWorkflowExample.java
```diff
@@ -0,0 +1,84 @@
...
+import io.littlehorse.sdk.worker.LHTaskWorker;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.Map;
+import java.util.Properties;
+
+public class ChildWorkflowExample {
```
Snarr, line 17: Nit-picky but can we change the name of these new classes for the 2 examples?

lhctl/internal/wf_spec.go
coltmcnealy-lh: @mitchell-h this is the bug that we found last month.

Snarr (approved):
Looks good and runs on my machine, let's change the example class names as mentioned in my comment above ^^

We should also follow up on these comments with issues:

https://github.com/littlehorse-enterprises/littlehorse/pull/1808/files#r2433108643

https://github.com/littlehorse-enterprises/littlehorse/pull/1808/files#r2433169728

Issue #1800, opened 2025-08-21
Support `TIMESTAMP` type in `VariableValue` structure


coltmcnealy-lh:
Issue #1782 reports that in Java the `LHLibUtil#objToVarVal()` method fails when serializing an `Instant` type. It likely also fails with a `Date` type.

This is because that method first checks against all of the valid types in the `VariableValue` protobuf, and if none match then it treats the Java object as a `JSON_OBJ` (which is then attempted to be serialized into a Json String with Gson).

Obviously, the `Date` and `Instant` types don't play nicely with json serialization.

## Proposal

We should extend the `VariableValue` type to support `TIMESTAMP` (this will be a primitive type in `VariableType`).

### Proto
```proto

// Specifies a primitive type in LittleHorse.
enum VariableType {
 // ...
  // Represents a time normalized to the central time (GMT).
  TIMESTAMP = 8;
}

message VariableValue {
  reserved 1;

  // The value held in this VariableValue. If this is unset, treat it as
  // a NULL.
  oneof value {
    // ...

    // A timestamp
    google.protobuf.Timestamp utc_timestamp = 10;
  }
}
```

### SDK Examples
In java:

```
WfRunVariable myTimestamp = wf.declareTimestamp("my-timestamp");
wf.execute("my-task", myTimestamp);

myTimestamp.assign(wf.execute("task-that-returns-instant"));
```

```java
@LHTaskMethod("task-that-returns-instant")
public Instant foo() {
  return Instant.now();
}

@LHTaskMethod("my-task")
public void bar(Date time) { // will support Instant and Date and `com.google.protobuf.Timestamp`
  System.out.println(time);
}
```

## Future Work (out of scope)
* Casting `INT` (64-bit integer) to `TIMESTAMP`
* Casting `DOUBLE` (64-bit float) to `TIMESTAMP`
* Casting `STR` to `TIMESTAMP` using timezone formats. This is tricky.
* Storing timezones within the `TIMESTAMP`.

coltmcnealy-lh:
CC @clipinski 😄

Issue #1782, opened 2025-08-06
Bug Report: Serialization Failure for Java 8+ Time API (java.time.Instant, etc.) with GSON


clipinski:
### Environment
- **LittleHorse SDK Version**: 0.14.1
- **Java Version**: 17+
- **LittleHorse Server Version**: 0.14.1 (image: `littlehorse-server:0.14.1`)

---

### Issue Description

When a LittleHorse task returns an object containing `java.time.Instant` fields, the task fails during serialization with a `JsonIOException`. This affects all Java 8+ time types (such as `Instant`, `LocalDateTime`, etc.) due to GSON not supporting these types natively.

---

### Error Details

    com.google.gson.JsonIOException: Failed making field 'java.time.Instant#seconds' accessible; either increase its visibility or write a custom TypeAdapter for class java.time.Instant
        at com.google.gson.internal.reflect.ReflectionHelper.makeAccessible(ReflectionHelper.java:112)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$BoundField.<init>(ReflectiveTypeAdapterFactory.java:117)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$1.create(ReflectiveTypeAdapterFactory.java:99)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.getBoundFields(ReflectiveTypeAdapterFactory.java:162)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.create(ReflectiveTypeAdapterFactory.java:102)
        at com.google.gson.Gson.getAdapter(Gson.java:458)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.createBoundField(ReflectiveTypeAdapterFactory.java:117)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.getBoundFields(ReflectiveTypeAdapterFactory.java:166)
        at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory.create(ReflectiveTypeAdapterFactory.java:102)
        at com.google.gson.Gson.getAdapter(Gson.java:458)
        at com.google.gson.Gson.toJson(Gson.java:704)
        at com.google.gson.Gson.toJson(Gson.java:683)
        at com.google.gson.Gson.toJson(Gson.java:658)
        at io.littlehorse.sdk.common.util.LHLibUtil.serializeToJson(LHLibUtil.java:78)

---

[...]

Issue #1466, opened 2025-04-24
Timestamp Arithmetic


coltmcnealy-lh:
We should support some way to do timestamp arithmetic in the `WfSpec` DSL. This will be useful in a few places:

- An airline booking system that wants to notify customers 24 hours before their flight that they should check in.
- Send a reminder on the next business day at 8am (not just 24 hours later).

To do this, we probably need to introduce some new structures:

- a `Timestamp` variable type. We could just reuse the `google.protobuf.Timestamp` as the underlying representation, but it needs to somehow be entered into the `VariableValue`.
- Mutations relevant to the `Timestamp` in the `enum VariableMutationType` (hopefully renamed to `Operation` some day.

This proposal is very half-baked, and warrants significant discussion below and a proper API contract in the ticket before starting implementation.

PR #1828, opened 2025-09-10
feat(server, sdk-java) Add support for TIMESTAMP type

l4crito:
## Summary
This PR adds comprehensive TIMESTAMP support across the stack so workflow timestamp variables can be used directly by Java task methods in five common forms.  
Fixes: #1800, #1782  

- `com.google.protobuf.Timestamp`  
- `java.time.Instant`  
- `java.util.Date`  
- `java.time.LocalDateTime`  
- `java.sql.Timestamp`  

## Server-side Changes
- Added TIMESTAMP/utc_timestamp support.

## SDK Changes (Java)
- Added Gson adapters for time-related types under `io.littlehorse.sdk.common.LHLibUtil`:
  - `java.util.Date` → ISO-8601 string (serialize/deserialize).
  - `java.time.Instant` → ISO-8601 string (serialize/deserialize).
  - `java.time.LocalDateTime` → ISO-8601 string via system default timezone conversion.
  - `java.sql.Timestamp` → ISO-8601 string (serialize/deserialize).
  - `com.google.protobuf.Timestamp` → ISO-8601 string (serialize/deserialize).
- Added `declareTimestamp(String name)` in `WorkflowThread` for declaring timestamp workflow variables.

> **Note on `jsonPath` behavior**  
> If a JSON object contains a date field and we call something like:  
> ```java
> thread.execute("print-date", myVar.jsonPath("$.creationDate"));
> ```  
> the server will currently resolve this as a `STR`. If the target task input is declared as a `TIMESTAMP`, the server will reject it at runtime since the types do not match.  
> We avoid server-side parsing to prevent ambiguity (e.g., a task may genuinely expect a `String` containing a date literal).

### DIFF
--- a/schemas/littlehorse/common_enums.proto
+++ b/schemas/littlehorse/common_enums.proto
@@ -114,7 +114,7 @@ enum VariableType {
   // communication or tracking dependencies between workflows.
   WF_RUN_ID = 7;
 
-  // Represents a time normalized to the central time (GMT).
+  // Represents a time normalized to Coordinated Universal Time (UTC).
   TIMESTAMP = 8;
 }
 


examples/java/variables/src/main/java/io/littlehorse/examples/MyWorker.java
```diff
@@ -19,36 +28,56 @@ public Double greeting(@LHType(masked = true) String text) {
...
-                )
-        );
+        log.debug("Executing task sentiment-analysis vars (%s, %s, %s, %s, %s)".formatted(text,
+                sentimentScore, addLength, userId, instant));
         ProcessedText processedText = new ProcessedText();
         processedText.text = text;
         processedText.addLength = addLength;
         processedText.userId = userId;
         processedText.sentimentScore = sentimentScore;
+        processedText.creationDate = Date.from(instant);
```
Snarr: Should we put this in a separate example?

coltmcnealy-lh: It might make sense. Don't want to block a whole PR over this, however.

schemas/littlehorse/common_enums.proto
```diff
@@ -113,6 +113,9 @@ enum VariableType {
   // referencing specific workflow executions in various use cases, such as inter-workflow
   // communication or tracking dependencies between workflows.
   WF_RUN_ID = 7;
+
+  // Represents a time normalized to the central time (GMT).
```
coltmcnealy-lh: Nit pick: can we decide whether to call it UTC or GMT? I'd prefer UTC.

server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/expression/TimestampReturnTypeStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public TypeDefinitionModel getIdentity() {
+        return new TypeDefinitionModel(VariableType.TIMESTAMP);
+    }
+
+    @Override
+    public Optional<TypeDefinitionModel> add(ReadOnlyMetadataManager manager, LHTypeStrategy other)
+            throws InvalidExpressionException {
+        throw new InvalidExpressionException("Cannot add to a TIMESTAMP");
```
coltmcnealy-lh, line 19: This is worth an issue in the backlog.

schemas/littlehorse/common_enums.proto
Snarr: UTC as well

server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/expression/TimestampReturnTypeStrategy.java
Snarr: Looks like you made one well before timestamps were added :O Visionary https://github.com/littlehorse-enterprises/littlehorse/issues/1466

schemas/littlehorse/common_enums.proto
l4crito: Agreed

PR #1838, opened 2025-09-18
fix(server): uuid in timer key

coltmcnealy-lh:
Previously the key for a timer in the timer store was {maturation-time}/{topic}/{partitionkey}/{tenant}.

The partitionKey is always the WfRunId. Therefore if a single WfRun had two timers maturing at the same time, the second one would override the first (same key). This backwards-compatible PR changes the timer keys to be {maturation-time}/{uuid}. This will also have the benefit of shortening the keys which improves rocksdb performance.


coltmcnealy-lh:
To test this PR, what I did was run a `WfRun` with this logic with the server on the `master` branch:

```
wf.sleepSeconds(120);
wf.execute("some-task");
```

I then restarted the server on my current branch, and the `WfRun` moved to completed at the correct time, verifying backwards compatibiltiy

PR #1840, opened 2025-09-18
perf(server): shorten guid and allow capital letters in ids

coltmcnealy-lh:
- Allows capital letters in ID's
- Encodes the UUID's for wfRunId's to make them shorter, taking advantage of capital letters
- Continues to preserve 128 bits of entropy for uniqueness.

I thought about ULID's; however, ULID's are not supported natively in Java.


coltmcnealy-lh:
Closing this PR after internal discussions. There are competing proposals:

1. Wait for results of tests with current optimizations, which might already be sufficient
2. Make the WfRunId a ULID which makes inserts sorted lexicographically which should really help compaction

PR #1842, opened 2025-09-20
perf(server): put getables for a single wfrun under same prefix

coltmcnealy-lh:
This backwards-compatible change puts all getable's such as TaskRun's and NodeRun's for a single WfRun onto the same prefix. Doing this will put most of the keys for a WfRun onto the same block in RocksDB (or in adjacent blocks), which optimizes I/O both in compaction and in cold reads.


coltmcnealy-lh:
Canary tail latencies also _seem to_ improve a little bit but it's slight. Here's the before:
image

And here's the after:
image

examples/java/external-event/src/main/java/io/littlehorse/examples/ExternalEventExample.java
coltmcnealy-lh, line 1: I'd recommend reverting these changes. Useful for testing this specific change, but it does make the example more complex than it needs to be.

server/src/main/java/io/littlehorse/common/Storeable.java
```diff
@@ -1,13 +1,29 @@
...
 
 public abstract class Storeable<T extends Message> extends LHSerializable<T> {
 
+    public static final String GROUPED_WF_RUN_PREFIX = "wrg";
+
     public String getFullStoreKey() {
-        return getFullStoreKey(getType(), getStoreKey());
+        if (getGroupingWfRunId().isPresent()) {
+            if (!(this instanceof StoredGetable)) {
+                // TODO: We'lll have to be fancy for the WfRunStoredInventory
```
coltmcnealy-lh: :porg: This code or this comment looks wrong.

server/src/main/java/io/littlehorse/common/model/getable/GroupedObjectId.java
coltmcnealy-lh, line 1: This interface looks like it's not used anywhere. Can we clean it up and remove?

server/src/main/java/io/littlehorse/common/model/getable/objectId/TaskRunIdModel.java
coltmcnealy-lh, line 1: Not sure where to put this comment but we need to override `UserTaskRunModel#getGroupingWfRunId()` as well.

server/src/main/java/io/littlehorse/server/monitoring/HealthService.java
```diff
@@ -76,6 +81,7 @@ public HealthService(
         statusServer.handle(config.getStatusPath(), ContentType.JSON, this::getStatus);
         statusServer.handle(config.getDiskUsagePath(), ContentType.JSON, this::getDiskUsage);
         statusServer.handle(config.getStandbyStatusPath(), ContentType.JSON, this::getStandbyStatus);
+        statusServer.handle("/dump-store", ContentType.TEXT, this::dumpStore);
```
coltmcnealy-lh: Reminder to revert this :smile: I know you probably already did.

server/src/main/java/io/littlehorse/server/streams/BackendInternalComms.java
```diff
@@ -710,7 +710,8 @@ private InternalScanResponse objectIdPrefixScanOnThisHost(InternalScan req) {
         ReadOnlyTenantScopedStore store = getStore(partition, req.storeName);
         PartitionBookmarkPb partBookmark = reqBookmark.getInProgressPartitionsOrDefault(partition, null);
 
-        String endKey = req.boundedObjectIdScan.getEndObjectId() + "~";
+        String endKey =
+                StoredGetable.getRocksDBKey(req.boundedObjectIdScan.getStartObjectId(), req.getObjectType()) + "~";
```
coltmcnealy-lh, line 714: Is there a reason why we are using `getStartObjectId()` instead of `getEndObjectId()`? It might work but still it seems not so kosher.

server/src/main/java/io/littlehorse/server/streams/store/StoredGetable.java
```diff
@@ -78,6 +90,14 @@ public StoreableType getType() {
     }
 
     public static String getRocksDBKey(String key, GetableClassEnum objType) {
-        return objType.getNumber() + "/" + key;
+        // Weird but needed for the new format
+        var parts = key.split("/");
+        String wfRun = parts[0];
```
coltmcnealy-lh, line 95: This is unfortunate, but I see that it is necessary. We'll have to codify the fact that `_` and `/` are special characters for anyone who creates another thing that extends `ObjectId`

server/src/test/java/e2e/WfRunSearchTest.java
coltmcnealy-lh, line 1: A necessary sacrifice :saluting_face:

coltmcnealy-lh (commented):
Approved. But technically I opened this PR so I can't approve it myself.

server/src/main/java/io/littlehorse/common/Storeable.java
l4crito: ill get ride of it

server/src/main/java/io/littlehorse/server/streams/BackendInternalComms.java
l4crito: there was a reason but i don't remember exactly what it was, let me debug really quick

l4crito: I think  that getEndObjectId() will return something like wfrunId~ and when we build the prefix from it, it is gonna be /tenant/storable/wrg/wfrunId~/storable/gettable

server/src/main/java/io/littlehorse/common/Storeable.java
```diff
@@ -1,13 +1,28 @@
...
 
 public abstract class Storeable<T extends Message> extends LHSerializable<T> {
 
+    public static final String GROUPED_WF_RUN_PREFIX = "wrg";
+
     public String getFullStoreKey() {
-        return getFullStoreKey(getType(), getStoreKey());
+        if (getGroupingWfRunId().isPresent()) {
+            if (!(this instanceof StoredGetable)) {
+                throw new NotImplementedException();
```
coltmcnealy-lh: I'm still confused, how does this work with `WfRunStoredInventoryModel`? Are those objects stored in a different key prefix?

server/src/main/java/io/littlehorse/server/streams/BackendInternalComms.java
coltmcnealy-lh: Can you create a follow-up ticket with type `Refactor` to address this? We should properly refactor the Internal Scan protobuf to understand the concept of grouping WfRun's. I think that would be a cleaner implementation; but it can wait for a while longer I think.

PR #1849, opened 2025-09-24
fix(canary): put back the mergeServiceFiles for shadowJar

coltmcnealy-lh:


coltmcnealy-lh:
Oh, I think we can close this— @mijailr fixed this problem in a different PR. The upgrade to java 24 broke the canary because it broke the shadow jar

PR #1851, opened 2025-09-25
fix(server): Limit the number of io threads for internal grpc clients

eduwercamacaro:


server/build.gradle
```diff
@@ -19,7 +19,13 @@ task javadocJar(type: Jar) {
 
 def defaultJvmArgs = [
     '-XX:+UseZGC',
-    '-Dlog4j.shutdownHookEnabled=false'
+    '-Dlog4j.shutdownHookEnabled=false',
```
coltmcnealy-lh, line 22: Should we figure out how to allow users to specify their own jvm args? Some of these look like thye're intended for local host.

server/src/main/java/io/littlehorse/server/streams/InternalNettyChannel.java
```diff
@@ -0,0 +1,53 @@
...
+public class InternalNettyChannel implements AutoCloseable {
+
+    private static final int MAX_NUMBER_IO_THREADS = 3;
+    private static final String THREAD_NAME_PREFIX = "lh-grpc-internal";
+
+    private final ThreadFactory namedFactory = new DefaultThreadFactory(THREAD_NAME_PREFIX);
+    private final EventLoopGroup nioEventLoopGroup = new NioEventLoopGroup(MAX_NUMBER_IO_THREADS, namedFactory);
+    private final ManagedChannel grpcChannel;
+
+    public InternalNettyChannel(String host, int port) {
```
coltmcnealy-lh, line 23: Is there a reason why we don't pass in the `EventLoopGroup` into the constructor? This creates a new group for every single `Channel` (and we create one for every different server we need to talk to).

server/src/main/java/io/littlehorse/server/streams/InternalNettyChannel.java
```diff
@@ -0,0 +1,53 @@
...
+                .build();
+    }
+
+    public InternalNettyChannel(String host, int port, ChannelCredentials credentials) {
+        Objects.requireNonNull(host);
+        Objects.requireNonNull(credentials);
+        this.grpcChannel = NettyChannelBuilder.forAddress(host, port, credentials)
+                .eventLoopGroup(nioEventLoopGroup)
+                .channelType(NioSocketChannel.class)
+                .usePlaintext()
```
coltmcnealy-lh: This looks like a bug.

server/src/main/java/io/littlehorse/server/streams/InternalNettyChannel.java
eduwercamacaro: This is configuring an independent event loop group for each channel. Using the same event loop group might cause bottlenecks when having multiple connections to other servers.

PR #1852, opened 2025-09-28
fix(server): Solve memory leak in async waiter class

eduwercamacaro:
This PR solves a memory leak reported during soak tests.

AsyncWaiters clients were completing the incorrect stage of the Async chain.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streams/util/AsyncWaiters.java
+++ b/server/src/main/java/io/littlehorse/server/streams/util/AsyncWaiters.java
@@ -81,11 +81,15 @@ public class AsyncWaiters {
     private FutureResponseAndWaitingClient createFuture(String commandId, CompletableFuture<Message> completable) {
         Objects.requireNonNull(commandId);
         Objects.requireNonNull(completable);
-        CompletableFuture<FutureResponseAndWaitingClient> waitingClient = new CompletableFuture<>()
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/littlehorse/server/streams/util/AsyncWaiters.java
```diff
@@ -81,11 +81,15 @@ public void handleRebalance(Set<TaskId> taskIds) {
...
         Objects.requireNonNull(commandId);
         Objects.requireNonNull(completable);
-        CompletableFuture<FutureResponseAndWaitingClient> waitingClient = new CompletableFuture<>()
-                .orTimeout(removeCompletedAfter, timeUnit)
-                .handle((o, throwable) -> {
-                    return responses.remove(commandId);
-                });
+        CompletableFuture<FutureResponseAndWaitingClient> waitingClient =
+                new CompletableFuture<FutureResponseAndWaitingClient>().orTimeout(removeCompletedAfter, timeUnit);
+        waitingClient.whenComplete((o, throwable) -> {
```
eduwercamacaro, line 86: This is the key part of these changes. Now, this function returns the root stage.

Snarr (approved):
So now we will catch a scenario where `commandId` was never apart of the responses and throw an IllegalStateException? Whereas before it would not catch that scenario and the code would silently continue 👍

PR #1857, opened 2025-09-30
fix(standalone): use server dist binary

mijailr:
ShadowJar is broken since the Java upgrade. This makes the standalone image to use the built binary.


coltmcnealy-lh (approved):
Can we also do this with the Canary too?

PR #1863, opened 2025-10-01
Automatically build and publish snapshot versions into maven repositories. 

eduwercamacaro:
This PR adds the ability to build and publish snapshots versions every week on Fridays. Also, this PR modifies the current version management in the repository, putting a static version in the `gradle.properties` that can be associated with every commit. 

This PR opens the possibility to optimize the current internal release process. This process is going to be revised accordingly.


gradle.properties
```diff
@@ -1,4 +1,4 @@
-version=0.0.0-development
+version=0.15.0-SNAPSHOT
```
mijailr, line 1: This means we can't forget to change the number right after a release. What's the plan for changing the value? For every patch or for major/minor versions only?

eduwercamacaro: Yes, we should always put a version for `master` branch, so every commits is tied to a specific version. The plan is to bump this version immediately after every release. 
Something similar for release branches.

PR #1871, opened 2025-10-04
DO NOT MERGE: experiment to reduce fetch requests

coltmcnealy-lh:
This branch has a weird git history, but I need it so that I have something compatible with the other experiments running in my soak cluster. And a PR is needed to get the image build pipeline going


coltmcnealy-lh:
NVM, ther'es too much changed here, I'm just going to publish to my personal GHCR

PR #1873, opened 2025-10-06
docs(proposals): Add proposal for type safe field access in `WfSpecs`

Snarr:
This PR introduces the proposal for type safe `LHPath`s in `WfSpec`s using a new `get()` method.

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/getable/core/noderun/NodeRunModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/getable/core/noderun/NodeRunModel.java
@@ -483,10 +483,14 @@ public class NodeRunModel extends CoreGetable<NodeRun> {
             getSubNodeRun().arrive(time, processorContext);
             setStatus(LHStatus.RUNNING);
         } catch (NodeFailureException exn) {
-            failures.add(exn.getFailure());
-            setStatus(exn.getFailure().getStatus());
+            fail(exn);
             throw exn;
         }
+    }
+
+    public void fail(NodeFailureException exn) {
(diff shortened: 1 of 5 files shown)


proposals/009-type-safe-structpaths.md
```diff
@@ -0,0 +1,249 @@
...
+## Protobuf Changes
+
+### `StructPath` Type
+
+To enable this new functionality, we will introduce a new `StructPath` type for extracting data from `Struct`s, `Arrays`, `JSON_OBJ`, and `JSON_ARR`s:
+
+```proto
+// A path of repeated Strings resolving to a nested field in an object.
+message StructPath {
+  repeated string path = 1;
```
mijailr: I prefer `jsonPath` as query language instead of a list of strings, reason is that you can do things like:

Imagine this object:

```json
{
  "brand": "Something",
  "alternativeNames": [ "smthg", "smg" ]
}
```

With JSONPath and this example, you can query the last item in `alternativeNames`. Like:

```
.alternativeNames[-1]
```

JSONPath is a powerful query language, there are many operations you could do and the implementation is straightforward, plus it's a common language and you could validate against the schema. With an array of strings you can only navigate through keys.

Snarr: This is a great point thank you for your feedback.

I think your example could also be replicated with StructPath, say we support repeated fields of Integers and Strings, we could do the following:

```proto
alternativeNames.get(-1)
```

There are certainly more examples though for JSONPath that can't be replicated with just `get()`. I will need to consider those.

PR #1879, opened 2025-10-08
fix(server): WaitForThreads checks interrupt handlers on childs

eduwercamacaro:
This PR proposes a mechanism to check failures on interrupts when updating WaitForThreads nodes.


eduwercamacaro:
I've updated this PR to put the interrupt thread failure into the currentNodeRun for the parent thread

coltmcnealy-lh (approved):

PR #1881, opened 2025-10-09
feat(server): introduce durable execution through `Checkpoint` in `TaskRun`

coltmcnealy-lh:


local-dev/compile-proto.sh
mijailr, line 1: Why is this disabling the compile proto image?

coltmcnealy-lh: Mistake. That was because I wrote most of this on a plane. Everything except the tests.

examples/java/checkpoint-tasks/src/main/java/io/littlehorse/examples/CheckpointTaskWorker.java
```diff
@@ -0,0 +1,39 @@
...
+public class CheckpointTaskWorker {
+
+    private static final Logger log = LoggerFactory.getLogger(CheckpointTaskWorker.class);
+
+    @LHTaskMethod("greet")
+    public String greeting(String name, WorkerContext context) { // WorkerContext enables checkpointing
+        int attemptNumber = context.getAttemptNumber();
+        System.out.println("Hello from task worker on attempt " + attemptNumber + " before the checkpoint");
+
+        String result = context.executeAndCheckpoint(() -> {
```
Snarr, line 18: I'm interested in the semantics of this checkpoint syntax. Why `context.executeAndCheckpoint(Executable executable)` instead of `context.checkpoint(VariableValue value)`?

coltmcnealy-lh: The point of the checkpoint is to prevent having to do the computation twice.

Snarr: Ohh makes perfect sense now that you say that, I was thinking of checkpoints as logs and not retrievable units of data

PR #1902, opened 2025-10-19
fix(server): dodge tombstones on rehydration

coltmcnealy-lh:
See this documentation in RocksDB: https://github.com/facebook/rocksdb/wiki/Implement-Queue-Service-Using-RocksDB

When doing a range scan in rocksdb, if there were keys recently deleted from that range, then you end up iterating over tombstones for all of those keys. In the case of a queue-like service, you may have many (potentially 100k to 1M+) tombstones.

Assuming it takes 2 hours for a tombstone to be fully compacted down to the bottom level (which is totally reasonable in the case of significant state > 20GB), if you have 50 tasks per second per partition then you can end up with 50 * 3600 * 2 == 360,000 tombstones in the range. This is because for every `TaskRun` we end up with a new `Schedule

In our Kafka Streams Processor `init()` method, we do a full prefix scan over `ScheduledTaskModel` store. In case of high per-partition load, this can take a very long time, causing the producer to get fenced which is a big orzdash.

This PR solves the problem with two changes:
* Storing `ScheduledTask`s in chronological order and processing them in chronological order.
* Remembering the timestamp of the last-processed ScheduledTask
* In the `init()` method, starting the iteration from the timestamp of the last processed `ScheduledTask` so as to skip over the tombstones in the range.


server/src/main/java/io/littlehorse/server/streams/storeinternals/ReadOnlyGetableManager.java
```diff
@@ -125,14 +126,16 @@ public LHKeyValueIterator<Tag> tagScan(String start, String end) {
...
-     */
-    public ScheduledTaskModel getScheduledTask(TaskRunIdModel scheduledTaskId) {
-        return getScheduledTask(scheduledTaskId.toString());
+    public ScheduledTaskModel getScheduledTask(TaskRunIdModel taskRunId) {
+        TaskRunModel taskRun = get(taskRunId);
+        if (taskRun == null) return null;
+        return getScheduledTask(taskRun);
+    }
+
+    public ScheduledTaskModel getScheduledTask(TaskRunModel taskRun) {
```
eduwercamacaro, line 135: We should either add a java doc or rename this method in order to make it clear that the implementation will use the legacy key format.

server/src/main/java/io/littlehorse/server/streams/storeinternals/TaskQueueHintModel.java
```diff
@@ -0,0 +1,65 @@
...
+        return StoreableType.TASK_QUEUE_HINT;
+    }
+
+    @Override
+    public String getStoreKey() {
+        return TaskQueueHintModel.TASK_QUEUE_HINT_KEY;
+    }
+
+    public String getKeyToResumeFrom() {
+        return "a/" + LHUtil.toLhDbFormat(LHUtil.fromProtoTs(lastProcessedTimestamp));
```
eduwercamacaro: Consider making this prefix a constant.

server/src/main/java/io/littlehorse/common/model/ScheduledTaskModel.java
```diff
@@ -115,4 +103,33 @@ public void initFrom(Message proto, ExecutionContext context) {
...
+    }
+
+    public static String getScheduledTaskKey(TaskRunIdModel taskRunId, Date taskAttemptCreatedAt) {
+        // Note: only one ScheduledTask can be active at once for a
+        // TaskRun, so we don't need to worry about the attemptNumber.
+        //
+        // For compatibility we use "a" as a prefix since it is the first ASCII
+        // character. This guarantees that the start of the iteration doesn't ignore
+        // any previous keys.
+        return "a/" + LHUtil.toLhDbFormat(taskAttemptCreatedAt) + "/" + taskRunId.toString();
```
l4crito: That "a/" looks so weird

coltmcnealy-lh: Doesn't matter if it looks weird, users don't see it :wink:

The reason for this is that there could be thousands of old scheduled tasks from the old server, then we add a few more scheduled tasks before we restart the task worker.

If we did that, the Task Queue Hint would start from a timestamp (which has a `0` at first) and would skip over all of the old tasks, and we would end up dropping tasks. Putting `a/` guarantees that everything put from the new format comes _before_ anything with the old format in the iteration, which allows us to know that we won't drop the old tasks.

server/src/main/java/io/littlehorse/server/streams/storeinternals/ReadOnlyGetableManager.java
eduwercamacaro: I'm mentioning this because I see two overloaded methods. One uses the legacy key and the other doesn't which is confusing.

coltmcnealy-lh: The first method uses the second method. I am a little confused.

server/src/main/java/io/littlehorse/server/streams/storeinternals/ReadOnlyGetableManager.java
```diff
@@ -125,14 +126,16 @@ public LHKeyValueIterator<Tag> tagScan(String start, String end) {
...
+        ScheduledTaskModel result = getScheduledTask(ScheduledTaskModel.getScheduledTaskKey(taskRun));
+        if (result != null) return result;
+        return getScheduledTask(ScheduledTaskModel.getLegacyKey(taskRun));
     }
 
     public ScheduledTaskModel getScheduledTask(String scheduledTaskId) {
```
coltmcnealy-lh, line 141: Oh, @eduwercamacaro is this the confusing one? I think we can delete it.

server/src/main/java/io/littlehorse/server/streams/storeinternals/TaskQueueHintModel.java
```diff
@@ -0,0 +1,67 @@
...
+import io.littlehorse.common.Storeable;
+import io.littlehorse.common.model.ScheduledTaskModel;
+import io.littlehorse.common.proto.StoreableType;
+import io.littlehorse.common.proto.TaskQueueHint;
+import io.littlehorse.common.util.LHUtil;
+import io.littlehorse.server.streams.topology.core.ExecutionContext;
+import java.util.Date;
+
+/**
+ * See the description in the PR (TODO put PR number).
```
Snarr, line 14: You can update this now that the PR is open (#1902)

server/src/main/java/io/littlehorse/server/streams/storeinternals/ReadOnlyGetableManager.java
coltmcnealy-lh: This will move to a follow-up PR because it will end up touching more systems.

PR #1903, opened 2025-10-22
fix(server): properly handle case when task is already claimed

coltmcnealy-lh:


coltmcnealy-lh:
I want to keep this PR minimal and small. However, I believe that after we fix this, there is room for significant cleanup and complexity reduction from the following principle:

> We do not need to fetch the `ScheduledTask` in the `RequestExecutionContext`; we only need to use the `TaskRunId`. The result of processing the `Command` will have the `ScheduledTask` in it.

I had another branch which changed 12 files and reduced a lot of code; however, I believe it is best to do that separately after the 0.15 release.

server/src/main/java/io/littlehorse/server/streams/CommandSender.java
```diff
@@ -110,13 +111,14 @@ private <U extends Message> U buildRespFromBytes(ByteString bytes, Class<?> resp
...
+        taskClaim.setCommandId(LHUtil.generateGuid());
+        BiFunction<Message, Throwable, PollTaskResponse> completeTaskClaim = (taskClaimResponse, exception) -> {
             if (exception != null) {
                 client.onError(new LHApiException(Status.UNAVAILABLE, "Failed recording task claim to Kafka"));
-            } else {
-                client.sendResponse(scheduledTask);
             }
-            return (TaskClaimEvent) taskClaimResponse;
+            PollTaskResponse result = (PollTaskResponse) taskClaimResponse;
+            client.getResponseObserver().onNext(result);
```
eduwercamacaro, line 120: I assumed that the `TaskClaimEvent` responses a scheduled task that is equals to the one we received as an argument in this method. Wondering how this makes any difference.

server/src/main/java/io/littlehorse/server/streams/CommandSender.java
```diff
@@ -110,13 +111,14 @@ private <U extends Message> U buildRespFromBytes(ByteString bytes, Class<?> resp
 
     public void doSend(ScheduledTaskModel scheduledTask, PollTaskRequestObserver client) {
         CommandModel taskClaim = new CommandModel(new TaskClaimEvent(scheduledTask, client));
-        BiFunction<Message, Throwable, TaskClaimEvent> completeTaskClaim = (taskClaimResponse, exception) -> {
+        taskClaim.setCommandId(LHUtil.generateGuid());
```
eduwercamacaro, line 114: Good catch. This was missing. Thanks!

server/src/main/java/io/littlehorse/server/streams/CommandSender.java
coltmcnealy-lh: It doesn't. Look at the `process()` method of the `TaskClaimEvent.java`. THis code that was there before was silently failing with a casting exception.

eduwercamacaro (approved):

server/src/main/java/io/littlehorse/server/streams/CommandSender.java
eduwercamacaro: Yes, thanks for the explanation.
TaskClaimEvent can return an empty response, and in that case we return the received scheduled task.

PR #1905, opened 2025-10-22
fix(server): Support old clients that don't send a primitive type value

Snarr:
This adds support for old clients (< 0.14.1) that send a `TypeDefinition` without a `type`/`primitive_type` value. If a client sends a `TypeDefinition` without a `defined_type` case, it defaults to a `JSON_OBJ` now.

### Compatibility with 0.14.1

This fixes a number of broken tests when running the v0.14.1 E2E tests against the latest version of the server.

However, 4 tests still fail and those are intended to fail for the following reasons:

| Test Name                                                                                              | Reason                                                                                                                                  | When will this break old clients? |
|--------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------|
| FullyCompatibleSchemaUpdatesEvolutionTest shouldThrowErrorWhenPuttingIncompatibleStructDefEvolutions() | The latest version adds rules for how you can define StructDef Field Names. The old test had an unsupported character (`-`) in a field. | At StructDef compile time.        |
| VarMutationsNumbersTest shouldApplyMathMutationsOverNumberVariables()                                   | We no longer support assigning DOUBLE to INT without explicit casting. On old clients, this would have failed at runtime.               | At WfSpec compile time.           |
| VarMutationsNumbersTest shouldRollbackVariablesWhenMutationsFailed()                                   | We no longer support assigning DOUBLE to INT without explicit casting. On old clients, this would have failed at runtime.               | At WfSpec compile time.           |
| VarSubErrorTest shouldNotCastStuffToString()                                                           | We now support casting any type of primitive to a String.                                                                               | It won't break old clients.       |


server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/TypeDefinitionModel.java
```diff
@@ -116,7 +117,8 @@ public void initFrom(Message proto, ExecutionContext ctx) {
                 this.structDefId = StructDefIdModel.fromProto(p.getStructDefId(), ctx);
                 break;
             case DEFINEDTYPE_NOT_SET:
-            default:
+                this.definedTypeCase = DefinedTypeCase.PRIMITIVE_TYPE;
+                this.primitiveType = VariableType.JSON_OBJ;
                 break;
         }
```
coltmcnealy-lh, line 123: This seems like it will work. It will mean that old servers are not compatible with new clients, but that is our contract.

PR #1909, opened 2025-10-24
feat(server): upgrades to JDK 25

MateoRojas:


examples/java/struct-def/build.gradle
```diff
@@ -1,20 +1,17 @@
 plugins {
     id 'application'
-    id 'com.diffplug.spotless' version '6.25.0'
```
mijailr, line 3: We should have spotless on every example

mijailr: You can remove the version and keep the plugin

examples/java/struct-def/build.gradle
```diff
@@ -1,20 +1,17 @@
...
     mavenLocal()
 }
 
-spotless {
-    java {
-        target('**/*.java')
-        palantirJavaFormat()
-    }
-}
-
```
mijailr, line 17: Keep spotless

examples/java/struct-def/build.gradle
MateoRojas, line 3: We should put spotless in all examples, not just this one, in a separate commit

mijailr (changes requested):

coltmcnealy-lh (approved):

PR #1912, opened 2025-10-26
refactor(server): clean up taskqueue infrastructure

coltmcnealy-lh:
This refactor cleans up the logic for handling scheduled tasks in OneTaskQueue, TaskManager, and ReadOnlyGetableManager. It removes old code that was leftover from times when the server was intended to behave differently.

### DIFF
--- a/server/src/main/java/io/littlehorse/common/util/RocksConfigSetter.java
+++ b/server/src/main/java/io/littlehorse/common/util/RocksConfigSetter.java
@@ -108,6 +108,10 @@ public class RocksConfigSetter implements RocksDBConfigSetter {
             options.setRateLimiter(serverConfig.getGlobalRocksdbRateLimiter());
         }
 
+        // Open the DB faster
+        options.setSkipCheckingSstFileSizesOnDbOpen(true);
+        options.setSkipStatsUpdateOnDbOpen(true);
+
         options.setTableFormatConfig(tableConfig);
     }
 


server/src/main/java/io/littlehorse/server/streams/taskqueue/OneTaskQueue.java
```diff
@@ -35,12 +31,10 @@ public class OneTaskQueue {
     private final Queue<PollTaskRequestObserver> hungryClients = new LinkedList<>();
     private final String instanceName;
     private final LinkedHashSet<QueueItem> pendingTasks;
-    private final AtomicLong numberOfInMemoryTasks = new AtomicLong(0);
 
-    public OneTaskQueue(String taskDefName, TaskQueueManager parent, int capacity, TenantIdModel tenantId) {
+    public OneTaskQueue(String taskDefName, TaskQueueManager parent, TenantIdModel tenantId) {
```
eduwercamacaro, line 35: I like the changes proposed for this class because it significantly reduces the complexity while keeping the same behavior and/or outputs. 

`ScheduledTask` gave the false impression that a task could be completely resolved inside the `RequestContext`, but this was untrue.

I'm happy to see that the entire `OneTaskQueue` and `TaskManager` relies only on `TaskRunIds

server/src/main/java/io/littlehorse/server/streams/taskqueue/TaskQueueManager.java
```diff
@@ -38,23 +35,22 @@ public void onRequestDisconnected(PollTaskRequestObserver observer, TenantIdMode
     }
 
     public void onTaskScheduled(
-            TaskId streamsTaskId, TaskDefIdModel taskDef, ScheduledTaskModel scheduledTask, TenantIdModel tenantId) {
+            TaskId streamsTaskId, TaskDefIdModel taskDef, TaskRunIdModel scheduledTask, TenantIdModel tenantId) {
```
eduwercamacaro, line 38: suggestion: rename the variable to `taskRunId` or any other name that refers to the TaskRunId

server/src/main/java/io/littlehorse/server/streams/CommandSender.java
```diff
@@ -130,7 +130,7 @@ public void doSend(ScheduledTaskModel scheduledTask, PollTaskRequestObserver cli
                 .handleAsync(completeTaskClaim, networkThreadpool);
     }
 
-    public CompletableFuture<RecordMetadata> doSend(
+    public CompletableFuture<RecordMetadata> reportTaskAndDontWaitForResponse(
```
eduwercamacaro, line 133: Honestly,
It is annoying how we are using so many terms in this class that do not belong to `CommandSender` context.

I would expect this class to take commands and return a future that represents the result. Fancy names are an indicator that we are doing stuff in this class that we are not supposed to.

I know that this class was already using these terms in the methods implementation, but I think we need to solve that instead of changing the names.

For example:
Making this method to take a CommandModel a return a future with the result will keep this class simpler, and also it will force us to do it in the client code. It seems to me that the class that is calling this method is a better place to send a response back to the Grpc client.

coltmcnealy-lh: I agree with you, but does this need to be done in this PR? I'd like to keep things clean and simple. If you want I can return this to the original one which has four overloaded `doSend()` methods (which I thought was confusing) and then make this change in a follow up.

server/src/main/java/io/littlehorse/server/streams/taskqueue/TaskQueueManager.java
coltmcnealy-lh: In my opinion it's best to leave it as-is: VSCode/IntelliJ/a reader knows it's a `TaskRunIdModel`, but the fact that it's named `scheduledTask` tells us that the Task Run ID refers to a TaskRun which was just scheduled.

server/src/main/java/io/littlehorse/server/streams/CommandSender.java
eduwercamacaro: I see no reason to not do it in this PR. This is just a method rename and sending an extra argument. 
This is just a suggestion. Feel free to ignore it.

coltmcnealy-lh: Got it, I will file a `Task` ticket to clean this up. I don't personally have bandwidth this week to do it in this PR and properly test it unfortunately. But I do agree with you.

It's hard for me to write code during the week; I need a weekend in order to make real progress.

eduwercamacaro (approved):

PR #1925, opened 2025-11-08
refactor(server): clean up CommandSender.java internal api

coltmcnealy-lh:


coltmcnealy-lh:
@eduwercamacaro this is a start at the work you requested on #1912 . It is actually quite involved, and I'm not sure how to edit the tests to work with this new API. We _could_ delete the tests in `CommandSenderTest.java` which rely on the removed methods; not sure if you want to do that. Can you please advise?

coltmcnealy-lh:
Closing this. Was an intended helpful refactor but I don't think I can complete it

PR #1926, opened 2025-11-10
fix(server): properly map rocksdb log level

coltmcnealy-lh:


coltmcnealy-lh:
The `WARN` level does not log write stalls, which I believe we should have logged even with `INFO` level metrics.

eduwercamacaro (approved):

mijailr (approved):

PR #1930, opened 2025-11-10
Introduce production exception handler that handles RecordTooLargeExc…

eduwercamacaro:
…eption

### DIFF
--- a/server/src/main/java/io/littlehorse/common/LHServerConfig.java
+++ b/server/src/main/java/io/littlehorse/common/LHServerConfig.java
@@ -53,6 +53,7 @@ import org.apache.kafka.streams.errors.LogAndContinueExceptionHandler;
 import org.rocksdb.Cache;
 import org.rocksdb.LRUCache;
 import org.rocksdb.RateLimiter;
+import org.rocksdb.RateLimiterMode;
 import org.rocksdb.RocksDB;
 import org.rocksdb.WriteBufferManager;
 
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/littlehorse/server/streams/topology/core/LHProductionExceptionHandler.java
```diff
@@ -0,0 +1,32 @@
...
+public class LHProductionExceptionHandler extends DefaultProductionExceptionHandler {
+
+    /**
+     * Handles exceptions that occur during record production in Kafka Streams.
+     */
+    @Override
+    public ProductionExceptionHandlerResponse handle(
+            ErrorHandlerContext context, ProducerRecord<byte[], byte[]> record, Exception exception) {
+        if (exception instanceof RecordTooLargeException) {
+            log.debug("Dropping command due to record size exceeding Kafka's maximum message size limit: ", exception);
```
eduwercamacaro, line 27: I'm suggesting `debug` level as `INFO` level might be noisy in the server's logs. Any thoughts?

coltmcnealy-lh: How often do we expect to see this log? I think it only happens in a "really bad" case when we should alert the operator of a cluster, so it might actually make sense to be a `WARN`.

coltmcnealy-lh (approved):

PR #1934, opened 2025-11-10
fix(server): Limit the amount of bytes for each record in the core store

eduwercamacaro:
When we try to store a record that is bigger than the request maximum size in the streams producer, LH server fails with fatal errors, sending the server into an endless loop and crashing without a way to recover.

This PR introduces a mechanism to control the maximum request size for each producer by adding a new server configuration and a new serialization/deserialization (serdes) method that validates the size of a record prior to storage attempts.


examples/java/spawn-thread-foreach/src/main/java/io/littlehorse/examples/SpawnThreadForEachExample.java
coltmcnealy-lh, line 1: Can we revert these changes? :smile:

PR #1950, opened 2025-11-14
feat(dashboard): Add task node run information to the sidebard

Magol28:
The task node information is available on the sidebard
image

### DIFF
--- a/dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Sidebar/NodeRunInfo/NodeStatus.tsx
+++ b/dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Sidebar/NodeRunInfo/NodeStatus.tsx
@@ -1,10 +1,10 @@
 import { LHStatus, TaskStatus } from 'littlehorse-client/proto'
 import '../Node/node.css'
-import { taskStatus, wfRunStatus } from '../../../StatusColor'
+import { TASK_STATUS, WF_RUN_STATUS  } from '../../../StatusColor'
 
 export const NodeStatus = ({ status, type="workflow" }: { status: LHStatus | TaskStatus; type?: 'workflow' | 'task' }) => {
   const { color, Icon, textColor } =
-    type === 'workflow' ? wfRunStatus[status as LHStatus] : taskStatus[status as TaskStatus]
+    type === 'workflow' ? WF_RUN_STATUS[status as LHStatus] : TASK_STATUS[status as TaskStatus]
 
   return (
     <div className="ml-1 mt-2 flex items-center gap-2">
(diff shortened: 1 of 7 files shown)


dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Sidebar/Node/node.css
Comment (mijailr) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Sidebar/Node/node.css:1]: I see how cool is to use this magic classes to process colors for each state, although there's something you are missing handling it in this way.

You won't be able to catch when a new status is added in build time. I would suggest to change this to a typed data and not use `css`

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Sidebar/NodeRunInfo/NodeStatus.tsx
```diff
@@ -1,11 +1,13 @@
-import { LHStatus } from 'littlehorse-client/proto'
+import { LHStatus, TaskStatus } from 'littlehorse-client/proto'
 import { CircleCheck } from 'lucide-react'
 import '../Node/node.css'
-export const NodeStatus = ({ status }: { status: LHStatus }) => {
+
+export const NodeStatus = ({ status, type }: { status: LHStatus | TaskStatus; type?: string }) => {
+  const statusType = type === 'task' ? `task-status--${status.toUpperCase()}` : `node-status--${status.toLowerCase()}`
```
Comment (mijailr) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Sidebar/NodeRunInfo/NodeStatus.tsx]: why some are lowercase and others uppercase? 

This is where if you don't use a typed data structure, you won't catch when a new status is added during build time. Use typed data instead of classNames as strings.

dashboard/src/app/utils/struct.ts
```diff
@@ -24,3 +24,26 @@ const structToJSONObject = (struct: Struct): Object => {
...
+  if (!result) return 'No Output'
+
+  if (result.$case === 'output') {
+    return getAttemptOutput(result.value)
+  }
+
+  // For error/exception, prefer a message property if present, otherwise fallback to JSON
+  const val = result.value
+  return (val && (val.message ?? val.msg ?? val.toString())) || JSON.stringify(val) || 'No Output'
+}
```
mijailr, line 42: What is this? you should use the types from the littlehorse-client library. Also avoid using `any`

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Sidebar/NodeRunInfo/TaskRunNode.tsx
```diff
@@ -1,12 +1,58 @@
...
+import { InputVariables } from '../Components'
+import { Attempts } from '../Components/Attempts'
+import { NodeStatus } from './NodeStatus'
 
 export const TaskRunNode: FC<{ node: TaskNodeRun }> = ({ node }) => {
+  const taskRunId = node.taskRunId
+  const { tenantId } = useWhoAmI()
+  const [nodeTask, setNodeTask] = useState<TaskRun>()
+  const [attemptIndex, setAttemptIndex] = useState(0)
+  useLayoutEffect(() => {
```
Comment (mijailr) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Sidebar/NodeRunInfo/TaskRunNode.tsx]: You shouldn't use `useLayoutEffect` this must be handled by SWR.

mijailr (changes requested):

PR #1952, opened 2025-11-17
chore(server): upgrade to kafka 4.1.1

coltmcnealy-lh:


coltmcnealy-lh:
Turns out that we are now able to run with more than one stream thread. It works! I did:
* Soak for a few days.
* Test a few rolling bounces.
* Run with multiple stream threads ( :scream: )
* Kill one server and wait for restoration of standby's on the surviving servers.

Everything worked as expected.

eduwercamacaro (approved):

PR #1968, opened 2025-11-26
feat(dashboard): fix attempt selection and show result on a modal

Magol28:
When you change the task it wont reflect the one you are selecting.
When there is an error the result is an exception, and it is too large to present as a text, we are using a modal to show it
image

QA steps:
You can run the following example: ./gradlew example-checkpoint-task:run


steinsleger (approved):

Snarr (approved):
Looks great!

PR #1975, opened 2025-12-01
fix(server): Improve how node outputs are retrieved 

l4crito:
Add a new Storeable (NODE_OUTPUT) to store node outputs based on the node name, improving output retrieval performance by enabling direct lookup instead of backward iteration through node runs. Only outputs from nodes that are referenced by other nodes in the thread spec are stored to optimize storage.

Previously, to get a node's output, we performed a backward lookup through node runs to find the one matching the node name in the spec, which involved multiple lookups until a match was found.

Now, node outputs are stored separately when they are referenced by other nodes in the thread spec, allowing direct retrieval by thread run number and node name.

The logic maintains backward compatibility: it first attempts to find the output in the new format, and if not present, falls back to the old method for existing workflows.


examples/java/external-event/src/main/java/io/littlehorse/examples/ExternalEventExample.java
```diff
@@ -31,18 +36,32 @@ public static Workflow getWorkflow() {
...
             wf.mutate(name, VariableMutationType.ASSIGN, wf.waitForEvent("name-event"));
-
-            wf.execute("greet", name);
+            var output = wf.execute("greet", name);
+            var b = wf.waitForEvent("name-event");
+            wf.execute("greet", a);
+            wf.execute("greet", b);
+            SpawnedThread childThread = wf.spawnThread(child -> { // this is the child workflow
+                                                                  // thread
+                WfRunVariable childVar = child.addVariable("child-var", VariableType.STR);
```
eduwercamacaro: We should revert these changes and replace it with a separate example or an e2e

PR #1979, opened 2025-12-05
refactor(server): Remove timer topology

eduwercamacaro:
## Overview

This PR introduces **ServerTopologyV2**, a new topology that moves timer processing into the main Kafka Streams topology.  
This removes the need for a separate timer topology and keeps backward compatibility using a feature flag:

- `LHS_X_ENABLE_TIMER_STREAMS`

## Motivation

The current design uses two Kafka Streams instances (core + timer), which causes:

- More operational complexity  
- Higher resource usage  
- Extra communication through Kafka topics  

By merging timer processing into one topology, the system becomes simpler and more efficient.

## Solution: Unified Topology with Smart Routing

This new approach runs **core logic and timers in a single topology** and uses **one shared state store**.  

### Architecture Overview
image

## Key Components

The new topology runs both core logic and timers in a single pipeline using one shared state store. Timers are stored, checked, and executed inside the core topology, with wall-clock checks to detect when they are ready. Records are routed by type so timer events, output records, and normal commands each go to the right place. When a timer is on the wrong partition, it is sent through a repartition topic, and when it is already on the correct partition, it is executed directly. All timer commands are processed using the same store as core commands, reducing memory usage and avoiding duplicate processing.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,7 +100,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: "corretto"
-          java-version: "24"
+          java-version: "25"
 
       - name: Dump version
         env:
--- a/dashboard/jest.config.ts
+++ b/dashboard/jest.config.ts
@@ -16,6 +16,10 @@ const config: Config = {
   moduleNameMapper: {
     '^@/(.*)$': '<rootDir>/src/$1',
   },
+  // Transform ES modules from node_modules (like jose)
+  transformIgnorePatterns: [
+    'node_modules/(?!(jose|@panva|oidc-token-hash|oauth4webapi|preact-render-to-string|nanoid)/)',
+  ],
 }
 
 // createJestConfig is exported this way to ensure that next/jest can load the Next.js config which is async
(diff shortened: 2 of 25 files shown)


local-dev/configs/server-1.config
```diff
@@ -13,6 +13,7 @@ LHS_STREAMS_METRICS_LEVEL=debug
 LHS_METRICS_LEVEL=info
 
 LHS_X_MAX_DELETES_PER_COMMAND=5
+LHS_X_ENABLE_TIMER_STREAMS=true
```
coltmcnealy-lh: Propose commenting this out. I think having one server use timer topology and the other not use it would cause a broken configuration if we have `server-1.config` and `server-2.config` in use for local clustered testiing.

server/src/test/java/e2e/StandaloneTestBootstrapper.java
```diff
@@ -128,6 +128,7 @@ private Properties getBaseServerConfig() {
         serverProperties.put(LHServerConfig.ROCKSDB_TOTAL_MEMTABLE_BYTES_KEY, String.valueOf(1024L * 1024L * 100));
         serverProperties.put(LHServerConfig.ROCKSDB_TOTAL_BLOCK_CACHE_BYTES_KEY, String.valueOf(1024L * 1024L * 100));
         serverProperties.put(LHServerConfig.X_MAX_DELETES_PER_COMMAND_KEY, "10"); // To test wfrun deletion iteration
+        serverProperties.put(LHServerConfig.X_ENABLE_TIMER_STREAMS_KEY, "true");
```
coltmcnealy-lh: I'd prefer leaving this as `false` here so that we can verify that everything works without the timer topology.

PR #1980, opened 2025-12-08
feat(fix): Improve how to show do while conditions

Magol28:
Do while conditions where no presenting an clear diagram.
On this PR we change dagre to ELK so it is easier to draw, 
Dagree has 
❌ Poor cycle handling - Dagre doesn't have sophisticated cycle-breaking strategies 
❌ Inconsistent with cycles - Will struggle with back-edges like your 4→1 cycle

You can use: example-conditionals-while

### DIFF
--- a/dashboard/package.json
+++ b/dashboard/package.json
@@ -70,8 +70,6 @@
     "@types/react": "^18",
     "@types/react-dom": "^18",
     "@types/testing-library__jest-dom": "^5.14.9",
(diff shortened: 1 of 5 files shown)


bryson-g:
This is from `lh-examples/wealth-management`
Comparing both branches.

`fix-cycle-nodes`
image

`master`
image

Is it possible to make the edges straight as before?

Magol28:
initiate-transfer
image
image
image

Magol28:
image
image
image

PR #1981, opened 2025-12-09
feat(sdk-java): add checkpoint context log method

bryson-g:


Snarr (approved):
Looks great!

examples/java/checkpoint-tasks/src/main/java/io/littlehorse/examples/CheckpointTaskWorker.java
```diff
@@ -15,7 +16,8 @@ public String greeting(String name, WorkerContext context) { // WorkerContext en
         System.out.println("Hello from task worker on attempt " + attemptNumber + " before the checkpoint");
 
         String result = context.executeAndCheckpoint(
-                () -> {
+                (checkpointContext) -> {
+                    checkpointContext.log("test log");
```
Snarr: Can we change this to something more example friendly, like "This is a checkpoint log" or something ?

sdk-java/src/main/java/io/littlehorse/sdk/worker/CheckpointContext.java
```diff
@@ -0,0 +1,21 @@
...
+public class CheckpointContext {
+    private String logOutput;
+
+    public CheckpointContext() {
+        this.logOutput = "";
+    }
+
+    public void log(Object thing) {
+        if (thing != null) {
+            logOutput += thing.toString();
```
l4crito, line 12: may be we could add a `\n` to each entry or a separator like `;`

Snarr: Right now we just mirrored the behavior of WorkerContext log which also does this

Snarr: I agree maybe there is a use case for adding a `\n` character between entries though

PR #1984, opened 2025-12-12
chore(sdk-python,sdk-java,sdk-dotnet): Increase heartbeat interval

eduwercamacaro:
This PR proposes reducing the number of requests that clients send on the task worker protocol.

### DIFF
--- a/sdk-go/littlehorse/task_worker_internal.go
+++ b/sdk-go/littlehorse/task_worker_internal.go
@@ -15,6 +15,7 @@ import (
 )
 
 const TOTAL_RETRIES = 5
+const HEARTBEAT_INTERVAL = time.Minute
 
 func (tw *LHTaskWorker) registerTaskDef() error {
 	ptd := &lhproto.PutTaskDefRequest{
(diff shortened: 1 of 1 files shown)


sdk-go/littlehorse/task_worker_internal.go
```diff
@@ -256,7 +256,7 @@ func (m *serverConnectionManager) start() {
 			}
 		}
 
-		time.Sleep(time.Duration(time.Second * 8))
+		time.Sleep(time.Duration(time.Second * 60))
```
Snarr: Can we organize this into a constant somewhere so it is labeled as HeartbeatIntervalSeconds like the other SDKs?

eduwercamacaro: Thanks for the suggestion. Applied on this commit: https://github.com/littlehorse-enterprises/littlehorse/pull/1984/commits/0199d198c20790c973e990cf7dda9aa67376471b

Snarr: Looks great 👍

PR #1988, opened 2025-12-12
fix(dashboard): tenant selection

Magol28:
The tenant it is not being change when you select a new tenant

### DIFF
--- a/dashboard/src/contexts/WhoAmIContext.tsx
+++ b/dashboard/src/contexts/WhoAmIContext.tsx
@@ -16,12 +16,11 @@ const Context = createContext<ContextProps>({
   tenants: [],
 })
 
-type WhoAmIContextProps = Omit<ContextProps, 'tenantId'> & { tenantId?: string }
+type WhoAmIContextProps = ContextProps
 
-export const WhoAmIContext: FC<PropsWithChildren<WhoAmIContextProps>> = ({ children, user, tenants, tenantId: initialTenantId }) => {
+export const WhoAmIContext: FC<PropsWithChildren<WhoAmIContextProps>> = ({ children, user, tenants, tenantId }) => {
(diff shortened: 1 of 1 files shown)


dashboard/src/contexts/WhoAmIContext.tsx
```diff
@@ -15,10 +16,17 @@ const Context = createContext<ContextProps>({
   tenants: [],
 })
 
-type WhoAmIContextProps = ContextProps
+type WhoAmIContextProps = Omit<ContextProps, 'tenantId'> & { tenantId?: string }
```
mijailr: Why are you omitting it from context to then add it again? Can't you just change the ContextProp type?

Also the tenant will always be present. It should be optional.

dashboard/src/contexts/WhoAmIContext.tsx
```diff
@@ -15,10 +16,17 @@ const Context = createContext<ContextProps>({
...
 
-export const WhoAmIContext: FC<PropsWithChildren<WhoAmIContextProps>> = ({ children, user, tenants, tenantId }) => {
-  return <Context.Provider value={{ user, tenants, tenantId: tenantId || tenants[0] }}>{children}</Context.Provider>
+export const WhoAmIContext: FC<PropsWithChildren<WhoAmIContextProps>> = ({ children, user, tenants, tenantId: initialTenantId }) => {
+  const params = useParams()
+  const currentTenantId = (params?.tenantId as string) || initialTenantId || 'default'
```
mijailr: I don't understand why there is an "initialTenantId" and a currentTenantId. Both should come from params. 

Perhaps after the change to tenants in params makes sense to use only the one from params.

It will never be undefined.

mijailr (changes requested):

PR #1992, opened 2025-12-17
Workflow migrations proposal

JakeRose03:
This proposal introduces Workflow Migrations

I will also post a draft PR that has implementation for base case migrations.


proposals/010-workflow-migrations.md
```diff
@@ -0,0 +1,303 @@
...
+```
+message WfRunMigrationPlan {
+    // Name
+    MigrationPlanId id = 1;
+
+    // When the MigrationPlan was created.
+    google.protobuf.Timestamp created_at = 2; 
+
+    // Wrapper object that stores how to migrate every thread
+    MigrationPlan migration_plan = 3 ;
```
coltmcnealy-lh: Do you think we could just put the `map<string, ThreadMigrationPlan>` right here?

proposals/010-workflow-migrations.md
```diff
@@ -0,0 +1,303 @@
...
+
+The `MigrationPlan` consists of the following proto:
+
+```
+message MigrationPlan { 
+    // {Threadname: {How to migrate that thread}}
+    map<string, ThreadMigrationPlan> thread_migrations = 1;
+}
+```
+Migration plans are broken down to a thread by thread basis.
```
coltmcnealy-lh: It would be helpful in this proposal if you specify between `ThreadSpec` vs `ThreadRun` :slightly_smiling_face: 

There are some nuances because a single `ThreadSpec` could have zero or one or ten corresponding `ThreadRun`s.

proposals/010-workflow-migrations.md
```diff
@@ -0,0 +1,303 @@
...
+```
+// How to migrate a thread 
+message ThreadMigrationPlan {
+    // Name of the new thread we are migrating to
+    string new_thread_name = 1;
+    // name of curNode - > how to migrate curNode in new wfSpec
+    map<string, NodeMigrationPlan> node_migrations  = 2;
+    // Any variable that will be derived before migration node can be provided here
+    // if not provided then possible errors will be thrown
+    map<string, VariableValue> migration_variables = 3;
```
coltmcnealy-lh: Could we use a `VariableAssignment` here rather than a `VariableValue`? That would allow different `WfRun`'s to have different values rather than hard-coding a single value for all `WfRun`'s.

proposals/010-workflow-migrations.md
```diff
@@ -0,0 +1,303 @@
...
+Getting Object:
+
+```
+  // Get Migration Plan
+  rpc getMigrationPlan(MigrationPlanId) returns (WfRunMigrationPlan) {}
+```
+
+```
+message MigrationPlanId {
+    string name = 1;
```
coltmcnealy-lh: I think I can see why you're not using the `wf_spec_name` or `wf_spec_id` here...is that because this proposal supports defining multiple `MigrationPlan`s that would all move a `WfRun` from `my-workflow` version 0.0 to `my-workflow` version 0.1?

proposals/010-workflow-migrations.md
```diff
@@ -0,0 +1,303 @@
...
+In the first impl you must provide the migration variables for every thread in the `wfRunMigrationPlan`.
+Another approach to enable more flexibility would be to define the variables in the `migrateWfRunRequest`.
+
+### Proto
+
+We define the set of variables that are derived before the migration node but needed after the migration node
+
+```
+// How to migrate a thread 
+message ThreadMigrationPlan {
```
coltmcnealy-lh: Same comment here. Are we dealing with `ThreadSpec` or `ThreadRun`? To be obnoxious, I might say

> This is a plan to migrate a `ThreadRun` from one `ThreadSpec` to another `ThreadSpec` in a newer `WfSpec`.

proposals/010-workflow-migrations.md
```diff
@@ -0,0 +1,303 @@
...
+  map<string, MigrationVariables> = 5;
+
+}
+
+
+```
+
+```
+message MigrationVariables {
+    map<string, VariableValue> migrations_vars = 1
```
coltmcnealy-lh: Same comment applies here: it would be great to use a `VariableAssignment` so that it can be dynamic.

coltmcnealy-lh (commented):
Great progress!

proposals/010-workflow-migrations.md
JakeRose03: Exactly

PR #2006, opened 2025-12-26
ci(release): skip sdk-js publishing when version already exists

mijailr:
Skip sdk-js publishing when the version already exits, this will prevent failures when retrying the release pipeline after the publishing succeeds.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,6 +14,7 @@ jobs:
     outputs:
       tag: ${{ steps.set_tag.outputs.tag }}
       prerelease: ${{ steps.is_prerelease.outputs.prerelease }}
+      publish_js: ${{ steps.check_js.outputs.publish_js }}
     steps:
       - name: Set Tag
         id: set_tag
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -182,10 +182,22 @@ jobs:
...
+        env:
+          TAG: ${{ needs.prepare.outputs.tag }}
+        run: |
+          if npm view "littlehorse-client@$TAG" version >/dev/null 2>&1; then
+            echo "Version $TAG already exists on npm, skipping publish."
+          else
+            echo "version_exists=false" >> $GITHUB_OUTPUT
+          fi
       - name: Checkout
+        if: ${{ steps.check_version.outputs.version_exists == 'false' }}
```
eduwercamacaro: I wonder if we can do this at a higher level. I'm thinking maybe we can add a condition to the `publish-sdk-js` job level

eduwercamacaro (approved):
Thanks for this PR!

.github/workflows/release.yml
mijailr: Yes, we can definitely do that. I'll change it in that way.

Issue #1759, opened 2025-07-21
Validate SleepNode timestamp


coltmcnealy-lh:
`SleepNodeModel#validate()` needs work.

coltmcnealy-lh:
Right now it does nothing. We need to validate that the provided values evaluate to a time.

```
    @Override
    public void validate(MetadataProcessorContext ctx) {
        // TODO (#1759)
    }
```

For example, we need to check:
* `rawSeconds` evaluates to `INT`
* `isoDate` evaluates to `STR` (at the least)
* `timestamp` evaluates to `TIMESTAMP` or `INT`

Additionally we need to make this `WfSpec` valid:

```
WfRunVariable timeToSleepTo = wf.declareTimestamp("my-timestamp");

wf.sleepUntil(timeToSleepTo);
```

Right now, it fails.

PR #2012, opened 2026-01-06
feat(server): Validate input for SleepNode

eduwercamacaro:
solves #1759


server/src/main/java/io/littlehorse/common/util/TypeCastingUtils.java
```diff
@@ -52,6 +52,9 @@ public static boolean canBeType(VariableType sourceType, VariableType targetType
         if (sourceType == null) {
             return true;
         }
+        if (sourceType.equals(targetType)) {
```
eduwercamacaro, line 55: @l4crito what do you think about this change? This is mostly covering the case where the targetType and the sourceType are equals when evaluating if a variable can be another type.

l4crito: Isn't that case covered with this
`        if (sourceType == targetType) {
            return true;
        }
`
i thought that we could evaluate it with == since is an enum

coltmcnealy-lh (approved):

PR #2013, opened 2026-01-07
feat(sdk-python): Wait for condition

eduwercamacaro:


examples/python/user_tasks/user_tasks.py
```diff
@@ -48,7 +48,7 @@ def my_entrypoint(wf: WorkflowThread) -> None:
             user_task_output, delay_in_seconds, task_def_name, arg1, arg2
         )
 
-    return Workflow("example-user-tasks", my_entrypoint)
```
Snarr, line 51: I think this is a typo -- should still be named `example-user-tasks`

examples/python/wait_for_condition/wait_for_condition.py
```diff
@@ -0,0 +1,55 @@
...
+    return config
+
+
+def get_workflow() -> Workflow:
+
+
+    def my_entrypoint(wf: WorkflowThread) -> None:
+        def substract(handler: WorkflowThread) -> None:
+            handler.subtract(counter, 1)
+        counter = wf.declare_int("counter", VariableType.INT).with_default(2)
```
Snarr: Can we remove the `VariableType.INT` argument?

examples/python/wait_for_condition/wait_for_condition.py
```diff
@@ -0,0 +1,55 @@
...
+        wf.wait_for_condition(wf.condition(counter, Comparator.EQUALS, 0))
+        wf.add_interrupt_handler("subtract", substract)
+
+    return Workflow("example-basic", my_entrypoint)
```
Snarr: This name should be "example-wait-for-condition"

examples/python/user_tasks/user_tasks.py
eduwercamacaro: Yep. this shouldn't be here.

PR #2015, opened 2026-01-08
proposal(server): Metrics on littlehorse 

l4crito:


proposals/012-workflow-metrics.md
```diff
@@ -0,0 +1,329 @@
...
+#### Workflow Metrics
+
+* Workflow execution count
+* End-to-end workflow latency
+* Error and exception counts
+
+#### Task Metrics
+
+* Task execution count
+* Queue latency
```
mitchell-h: Could we add Queue Depth to this?  How many tasks executions are waiting.

proposals/012-workflow-metrics.md
```diff
@@ -0,0 +1,329 @@
...
+
+* Inconsistent across teams
+* Hard to correlate with workflow structure
+* Operationally expensive
+* Detached from LittleHorse’s execution model
+
+
+## Proposed Solution
+
+We propose introducing a **Workflow Metrics API** that allows users to explicitly define, collect, and query metrics related to workflow execution.
```
mitchell-h: Will this API be compatible with existing standardized tooling such as Promethus?  
optionally can we have these metrics delivered via output-topic?

l4crito: I believe that, for now, the scope is to access this only through gRPC or lhctl.

mitchell-h: I'd like to see it be a requirement it be delivered to promethus(prefered) or output topic.  but more importantly I'd like to understand the reason for gRPC.

eduwercamacaro: @mitchell-h These metrics are part of the littlehorse API and can become available via output topic in the future. 

We won't be exporting workflow metrics to Prometheus as we consider them to be a domain-specific metrics. We only export technical metrics to prometheus.

proposals/012-workflow-metrics.md
```diff
@@ -0,0 +1,463 @@
...
+message StatusRange {
+  oneof type {
+    LHStatusRange lh_status = 1;
+    TaskRunStatusRange task_run = 2;
+    UserTaskRunStatusRange user_task_run = 3;
+  }
+}
+
+message LHStatusRange {
+  LHStatus starts = 1;
```
coltmcnealy-lh: I don't know if we should rely on the ordering of the `LHStatus` enum here.

proposals/012-workflow-metrics.md
coltmcnealy-lh: That's part of COUNTED_TAGS, which is not covered by this. Queue depth isn't really addressable by this architecture

coltmcnealy-lh (changes requested):
I would prefer to do away with the `MetricSpec` entirely, but it could be a decent user experience if we address the two biggest concerns I have: window size configurations and also metric definition configurations.

## Window Size

We should make our windows _mergeable_ so that you can merge adjacent windows together into a larger window. For the case of a `latency_avg` metric, that would mean that the window stores `total_observations` and `total_latency` rather than storing "average latency." This allows merging two adjacent windows together.

Given that capability, I'd like to remove the window size configuration from the API entirely, and just standardize on a 5-minute (or other value, but I think 2 or 5 minutes is reasonable) window at the GRPC level.

The Dashboard should allow users to select their window size. When the user selects "2-hours" as their window size, the dashboard will select 24 adjacent 5-minute windows (two hours of data), and aggregate them together, for each hour.

This would improve the following:
* The process for defining metrics from the user perspective would be less complicated.
* The server doesn't need logic for handling different window lengths.

And users can still view metrics with whatever window size they want.

## Metrics Definitions

### Opt In

I understand the desire to make metrics "opt-in" for performance reasons. However, recent feedback from users of the output topic is that "opt-in" is really annoying (even there, opt-in is at the `Tenant` level. Here, it's much more labor-intensive: opt-in is at the `WfSpec` or even `Node` level).

### Defining By Status Transitions

The proposal is highly flexible by allowing users to choose to measure time from two arbitrary statuses. However, there are only so many transitions which make sense to measure, and I believe this API makes it harder for users to use / configure these metrics.

proposals/012-workflow-metrics.md
l4crito: Is this the major concern that we have? , letting the user to configure metrics that could not exists  like:
workflowLatency starts: `completed` end: `running`?

coltmcnealy-lh (commented):
After offline discsusions we will combine the best from this and #2020

proposals/012-workflow-metrics.md
```diff
@@ -0,0 +1,575 @@
...
+* Detached from LittleHorse’s execution model
+
+
+## Proposed Solution
+
+We propose introducing **workflow metrics** with a server-defined set of default `MetricSpec`s that are loaded at startup. Clients can query collected metrics via read-only APIs (gRPC), but cannot create or mutate metric definitions at runtime;
+
+Metrics are:
+
+* Pre-defined in `DefaultMetricsRegistry` (server-side)
```
coltmcnealy-lh: Is this an internal class?

proposals/012-workflow-metrics.md
```diff
@@ -0,0 +1,575 @@
...
+* Workflow level (`WfSpec`)
+* Thread level
+* Node level
+* Task level
+
+
+
+## Design Principles
+
+### Explicit Metric Definition
```
coltmcnealy-lh: This seems to conflict with what you said with the `DefaultMetricRegistry`.

proposals/012-workflow-metrics.md
```diff
@@ -0,0 +1,575 @@
...
+* Repartitioning for global aggregation
+
+
+
+### Minimal Performance Impact
+
+* Metrics are aggregated in time windows
+* Computation happens per partition first
+* Only finalized events produce metric updates
+* **Minimum window length**: 1 minute
```
coltmcnealy-lh: Thought we were setting the window length to 5?

proposals/012-workflow-metrics.md
```diff
@@ -0,0 +1,575 @@
...
+### Non-Intrusive Execution
+
+Metric collection does **not** affect workflow semantics or execution order.
+Existing workflows continue to work unchanged.
+
+
+### Mergeable windows & retention
+
+* Default windows are **5 minute windows** aligned to the epoch for operational simplicity.
+* Each window is persisted for **14 days** (configurable) to allow for historical queries.
```
coltmcnealy-lh: How is this configurable? At the tenant level, LH Server config, or somewhere else?

proposals/012-workflow-metrics.md
```diff
@@ -0,0 +1,575 @@
...
+```proto
+message Tenant {
+  // ... existing fields ...
+
+  // The default level of metrics to record in a given `Tenant`.
+  MetricRecordingLevel metrics_level = 4;
+}
+```
+
+The server-side `DefaultMetricsRegistry` may also contain per-`WfSpec` or per-`TaskDef` overrides; these overrides are managed by operators via configuration (not public RPCs).
```
coltmcnealy-lh: I don't think we should use environment variables to override metric levels for a specific `WfSpec` or `TaskDef`. Those should be controlled by RPC's.

It makes sense to use some RPC (either `rpc PutWfSpec` or perhaps a dedicated new one `rpc OverrideMetricLevel`) to override the metric recording level for a specific metadata.

proposals/012-workflow-metrics.md
```diff
@@ -0,0 +1,575 @@
...
+
+### Metric Recording Level Resolution Logic
+
+Metric recording levels are resolved hierarchically for each metric collection event:
+
+1. **Check for specific overrides**: Look up `MetricLevelOverride` objects for the specific `WfSpec`, `TaskDef`, or `Node` involved.
```
coltmcnealy-lh: Agreed. This didn't seem to be what the paragraph above said, could you update it to make the writing more clear please?

proposals/012-workflow-metrics.md
```diff
@@ -0,0 +1,575 @@
...
+public class DefaultMetricsRegistry {
+    public static List<MetricSpec> builtIn() {
+        return List.of(
+            MetricSpec.builder("workflow-completed-5m")
+                .aggregation(COUNT)
+                .scope(globalScope())
+                .transition(WF_RUN, "RUNNING", "COMPLETED")
+                .window(5m)
+                .build(),
+            // more defaults...
```
coltmcnealy-lh: The proposal should show these.

PR #2021, opened 2026-01-13
feat(sdk-dotnet): implement checkpoints

Snarr:


sdk-dotnet/LittleHorse.Sdk.Tests/Worker/VariableMappingTest.cs
```diff
@@ -130,12 +131,15 @@ public void VariableMapping_WithAssignIntValue_ShouldReturnInt32Object()
...
+        var mockClient = new Mock<LittleHorseClient>();
+
         foreach (var type in testAllowedTypes)
         {
             var variableMapping = GetVariableMappingForTest(type, paramName, position);
             VariableValue variableValue = new VariableValue {Int = expectedValue};
             ScheduledTask taskInstance = GetScheduledTaskForTest(variableValue, paramName);
-            var mockWorkerContext = new Mock<LHWorkerContext>(taskInstance, new DateTime());
+
+            var mockWorkerContext = new Mock<LHWorkerContext>(taskInstance, new DateTime(), mockClient.Object);
```
eduwercamacaro, line 142: I don't think we should create a mock for `LHWorkerContext`. Instead, I would suggest creating a dummy instance like this:

`var mockWorkerContext = new LHWorkerContext(taskInstance, new DateTime(), mockClient.Object);`

eduwercamacaro (approved):

sdk-dotnet/LittleHorse.Sdk.Tests/Worker/VariableMappingTest.cs
Snarr: True, this is just based on the existing test code that was in there but you're right we don't really need a mock in this test ✅🤔

Issue #1940, opened 2025-11-11
Pagination for List RPC does not work


eduwercamacaro:
I'm trying to list all task run within a wfrun, but I'm getting this error:

```
❯ lhctl list taskRun 1666dead95474534a4456b905c41003f
2025/11/11 11:47:17 rpc error: code = Internal desc = Unexpected internal failure: java.lang.NoSuchMethodException: io.littlehorse.sdk.common.proto.TaskRunList$Builder.setBookmark(com.google.protobuf.ByteString)
``` 

I think we are missing a `bookmark` field in most of the `List*` RPC.

PR #2027, opened 2026-01-19
fix(server): Fix pagination for List rpc

eduwercamacaro:
solves #1940

### DIFF
--- a/.github/workflows/branch.yml
+++ b/.github/workflows/branch.yml
@@ -33,7 +33,7 @@ jobs:
           java-version: 25
 
       - name: Tests and Build
-        run: ./gradlew server:build server:installDist
+        run: ./gradlew server:installDist
 
       - uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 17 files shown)


coltmcnealy-lh (approved):

server/src/main/java/io/littlehorse/server/streams/BackendInternalComms.java
```diff
@@ -725,7 +725,7 @@ private InternalScanResponse objectIdPrefixScanOnThisHost(InternalScan req) {
             while (iter.hasNext()) {
                 LHIterKeyValue<? extends Storeable<?>> next = iter.next();
                 if (--curLimit < 0) {
-                    bookmarkKey = next.getKey();
+                    bookmarkKey = next.getValue().getStoreKey();
```
eduwercamacaro, line 728: @coltmcnealy-lh 
This is needed as we already put a prefix when processing the next `list` request using this bookmark. 

`next.getKey()` is the full storeKey including the tenantId.

Snarr (approved):

PR #2028, opened 2026-01-20
fix(server): Prefix scan for metadata objects

eduwercamacaro:
This PR fixes a problem in the BackendInternalComms class when executing prefix scans for metadata objects. 

https://github.com/littlehorse-enterprises/littlehorse/pull/1842 Introduced this bug after assuming that the metadata objects were also grouped by a WfRun.

### DIFF
--- a/examples/go/wait-for-condition/README.md
+++ b/examples/go/wait-for-condition/README.md
@@ -1,0 +1,32 @@
+## Running WaitForConditionExample
+
+This is a wait for condition example, which does three things:
+1. Declare an "counter" variable of type Integer
+2. Waits until the counter reaches 0.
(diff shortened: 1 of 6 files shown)


server/src/main/java/io/littlehorse/server/streams/store/LHKeyValueIterator.java
```diff
@@ -26,7 +29,9 @@ public boolean hasNext() {
 
     @Override
     public LHIterKeyValue<T> next() {
-        return new LHIterKeyValue<>(rawIter.next(), cls, executionContext);
+        KeyValue<String, Bytes> next = rawIter.next();
+        log.info("next key: {}", next.key);
```
l4crito: Think we don't need the log here

PR #2030, opened 2026-01-21
feat(sdk-dotnet): Support Timestamp and WfRunId variables

eduwercamacaro:


sdk-dotnet/LittleHorse.Sdk.Tests/Worker/VariableMappingTest.cs
```diff
@@ -291,6 +292,47 @@ public void VariableMapping_WithAssignBoolValue_ShouldReturnBoolObject()
         
         Assert.Equal(taskInstance.Variables[0].Value.Bool, result);
     }
+
+
+    [Fact]
+    public void VariableMapping_WithAssignTimestampValue_ShouldReturnDateObject()
```
Snarr, line 298: I understand that this framework was in place before your PR, but it seems like overkill testing this functionality based on `VariableMapping.Assign` as opposed to testing all of this with the new `LHMappingHelper.VariableValueToObject` method.

Snarr (approved):
Approved but we should overhaul these tests

PR #2032, opened 2026-01-22
feat(server): Control output topic auto creation with LHS_SHOULD_CREATE_TOPICS config

eduwercamacaro:
This PR proposes a way to control output topics auto creation when `LHS_SHOULD_CREATE_TOPICS` is set to false. The goal here is to avoid sending create topic requests to the kafka clusters when this config is set to false.

Usually, Littlehorse operators set this config to false when to productions like staging or production.

### DIFF
--- a/server/src/test/resources/test.properties
+++ b/server/src/test/resources/test.properties
@@ -1,1 +1,1 @@
-bootstrapper.class = e2e.ExternalTestBootstrapper
+bootstrapper.class = e2e.StandaloneTestBootstrapper


server/src/main/java/io/littlehorse/common/LHServerConfig.java
```diff
@@ -1114,6 +1114,9 @@ public int getBulkJobDelayIntervalSeconds() {
      * @throws ExecutionException if the topic creation callback fails.
      */
     public boolean createKafkaTopic(NewTopic topic) throws InterruptedException, ExecutionException {
+        if (kafkaAdmin == null) {
+            return false;
+        }
         try {
```
mijailr, line 1120: Would be more clear to not call this method at all.

server/src/main/java/io/littlehorse/common/LHServerConfig.java
```diff
@@ -1156,6 +1159,9 @@ private void initRocksdbSingletons() {
     }
 
     private void initKafkaAdmin() {
+        if (!shouldCreateTopics()) {
+            return;
+        }
```
mijailr, line 1164: Also this, shouldn't call initKafkaAdmin at all.

PR #2033, opened 2026-01-22
chore(proposal): refactor edge conditions

coltmcnealy-lh:


coltmcnealy-lh:
I have been debating whether to completely overhaul how the `VariableMutation` structure works. I believe it could look like the following:

```proto
message NewVariableMutation {
  string variable_name = 1;

  optional LHPath path = 2;

  VariableAssignment new_value = 3;
}
```

I think that would greatly simplify our protobuf API, but it would require another deprecation for 2.0 (and until 2.0 release would increase complexity).

proposals/014-conditions-refactor.md
```diff
@@ -0,0 +1,170 @@
...
+WfRunVariable foo = wf.declareInt("foo");
+wf.doIf(foo.isLessThan(10), handler -> {
+    handler.execute("hello");
+});
+```
+
+What is going to be no longer valid is the following code:
+
+We will release these changes in the 1.0 release. Our compatibility promise will be that:
+
```
l4crito, line 170: Looks like we are missing some code here

PR #2042, opened 2026-01-25
fix(server): save timer iterator progress and isolate timers

coltmcnealy-lh:
This PR fixes two problems with the Timer Processor in the core topology.
    
- We iterate from the beginning of the timer store in the first punctuation. When creating a new processor (eg. after a rebalance) that involves iterating over tombstones, and that has been observed to take more than 60 seconds, which increases our downtime window.
- We were not isolating LHTimer's with the Storeable infrastructure but rather storing them as if they were the only thing in the core store. Thus when we put something inside the cluster-scoped store, we get problems as the timer iterator sees those things as timers.
    
We'll also need to guard the Tenant names to prevent them from starting with a digit, as that could result stuff in the tenant-scoped stores being mistaken for a cluster-scoped timer.
    
This PR also disables the timer topology by default.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streams/topology/core/TimerCoreProcessor.java
+++ b/server/src/main/java/io/littlehorse/server/streams/topology/core/TimerCoreProcessor.java
@@ -31,8 +31,6 @@ public class TimerCoreProcessor implements Processor<String, LHTimer, String, Ob
     private long lastSeenTimestampMillis;
     private long lastCheckpointedHintTimeMillis;
 
-    private boolean seenLastKey = false;
-
     private final boolean forwardTimers;
 
     public TimerCoreProcessor(boolean forwardTimers) {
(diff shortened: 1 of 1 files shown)


server/src/main/java/io/littlehorse/server/streams/topology/core/TimerCoreProcessor.java
```diff
@@ -65,15 +79,40 @@ public void process(final Record<String, LHTimer> record) {
 
     private void clearTimers(long timestamp) {
         String end = LHUtil.toLhDbFormat(new Date(timestamp));
+        String lastSeenKey =
+                lastSeenTimestampMillis == 0L ? "000000" : LHUtil.toLhDbFormat(new Date(lastSeenTimestampMillis));
+
+        if (!seenLastKey) {
+            System.out.println(lastSeenKey);
```
eduwercamacaro: We can remove this println

coltmcnealy-lh: Thanks!

PR #2044, opened 2026-01-26
feat(server): Archive ThreadRuns

Snarr:
This PR implements Proposal #012 "Archive Inactive `ThreadRun`s"


examples/java/child-thread/src/main/java/io/littlehorse/examples/ChildThreadExample.java
coltmcnealy-lh, line 1: I would prefer reverting this to the old one, because the educational value of the original example is greater.

This seems more like a test case than an example

PR #2045, opened 2026-01-27
feat(server): Support using WfRunVariable as input for RunChildWfNode

eduwercamacaro:
Implements #2047 
This PR extends the `RunChildWfNode` to support using `WfRunVariable` as the `WfSpec` name when running a child workflow. This allows the child `WfSpec` to be determined dynamically at runtime instead of being fully specified at WfSpec compile time.

Changes include:
 - Support for resolving the child WfSpec from a VariableAssignment instead of only from a static `RunChildWfNode .wfSpecName`.
 - New backward compatible field for `RunChildWfNode`
    - Existing behavior using a static `wfSpecName` continues to work unchanged.
- Validation of child workflow inputs/outputs is not executed at WfSpec compile time when the spec is provided dynamically.

#### Defer validation
The WfSpec name defined via VariableAssignment is validated only at node arrival time when the variable values are resolved and the child spec name is actually known. If resolution or validation fails at runtime (e.g., non-existing spec name or LHVarSubError), the node fails with a `CHILD_FAILURE` error type.

#### Motivation
This PR enables the Dynamic child WfSpec pattern, while preserving compatibility with previous versions.

Dynamic workflow example:

```java
    WfRunVariable someVar = workflowThread.declareStr("some-variable");
    SpawnedChildWf childWf = workflowThread.runWf(someVar, Map.of());
    NodeOutput opWfOutput = workflowThread.waitForChildWf(childWf);
    ...
```

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
@@ -306,8 +306,10 @@ public class ThreadRunModel extends LHSerializable<ThreadRun> {
         ExternalEventDefIdModel extEvtId = e.getId().getExternalEventDefId();
         InterruptDefModel idef = getThreadSpec().getInterruptDefFor(extEvtId.getName());
         if (idef != null) {
-            // trigger interrupt
-            initializeInterrupt(e, idef);
+            if (!isTerminated()) {
+                // trigger interrupt
+                initializeInterrupt(e, idef);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


examples/java/basic/src/main/java/io/littlehorse/examples/BasicExample.java
```diff
@@ -21,6 +22,7 @@ public class BasicExample {
     public static Workflow getWorkflow() {
         return new WorkflowImpl("example-basic", wf -> {
             WfRunVariable theName = wf.declareStr("input-name").searchable();
+            wf.runWf(theName, Map.of());
```
eduwercamacaro: Well, we don't actually need this. wondering if we should create a dedicated example for this particular feature or reuse the existing child-wf example.

coltmcnealy-lh (approved):

PR #2057, opened 2026-02-05
fix(server): allow updating sleep until timestamps

coltmcnealy-lh:


server/src/test/java/e2e/SleepTest.java
```diff
@@ -54,10 +58,59 @@ void shouldCompleteWhenTimestampIsInThePast() {
...
+                .start();
+    }
+
+    @Test
+    void shouldRescheduleSleepWhenInterruptUpdatesTimestampEarlier() {
+        long now = System.currentTimeMillis();
+        long initialTime = now + 5000;
+        long updatedTime = now + 1000;
+        verifier.prepareRun(sleepTest, Arg.of("timestamp-to-wait-for", initialTime))
+                .thenSendExternalEventWithContent(INTERRUPT_EVENT, updatedTime)
```
eduwercamacaro, line 89: can we wait for the running status before sending the external event?

server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/subnoderun/SleepNodeRunModel.java
```diff
@@ -97,6 +99,31 @@ public void processSleepNodeMatured(SleepNodeMaturedModel evt) {
...
+            Objects.requireNonNull(newMaturationTime, "Maturation resolved to null.");
+            if (maturationTime != null && maturationTime.equals(newMaturationTime)) {
+                return;
+            }
+            maturationTime = newMaturationTime;
+            SleepNodeMaturedModel snm = new SleepNodeMaturedModel(nodeRun.getId());
+            CommandModel command = new CommandModel(snm, maturationTime);
+            processorContext.getTaskManager().scheduleTimer(new LHTimer(command));
+        } catch (LHVarSubError exn) {
+            // Best-effort: if the variable can't be resolved, keep the original timer.
```
eduwercamacaro: hmmm, I don't believe this represents the "best effort." Instead, I believe this decision leads to silent failures when rescheduling a sleep node. It would be difficult to know that something failed in this workflow and even harder to debug. My recommendation here is to throw a technical error on the workflow.

eduwercamacaro: And I believe it contradicts what we are doing with the validation above. `Objects.requireNonNull(newMaturationTime, "Maturation resolved to null.");`

coltmcnealy-lh: Okay. I'll update that, then good to merge?

PR #2062, opened 2026-02-07
feat(clients): provide shortcut to register `ExternalEventDef` on interrupt handler

coltmcnealy-lh:


Snarr (approved):
Looks good, could use a small test or two in each of the SDKs but its a simple feature so its not the biggest deal. More-so to prevent it from breaking in the future.

PR #2071, opened 2026-02-10
feat(server): Workflow metrics

l4crito:
## Overview

This pull request introduces a comprehensive workflow metrics system to LittleHorse, enabling detailed observability and monitoring of workflow executions in windows of 1 minute. 

---

## Key Features

- **Metrics Collection:** Automatically tracks workflow lifecycle events for count and latency.
- **Partitioned Metrics:** Metrics are collected per Kafka partition for scalability.
- **Global Aggregation:** Partitioned metrics are aggregated into a centralized location for unified querying.
- **Pre-calculated Queries:** Aggregated metrics are computed during processing, enabling fast query responses.
- **Time-Windowed Aggregation:** Metrics are aggregated in 1-minute time windows for granular monitoring.
- **gRPC API Endpoints:** New `ListWfMetrics` API for querying aggregated metrics with time window filtering.
- **CLI Integration:** Added metrics listing command to `lhctl`.

---

## Example Usage

After deploying workflows and running executions, users can query metrics using:

```bash
lhctl list wfMetrics example-basic
lchtl list taskMetrics greet

### DIFF
--- a/dashboard/src/app/(authenticated)/[tenantId]/(diagram)/wfSpec/[...props]/components/SearchVariableDialog.tsx
+++ b/dashboard/src/app/(authenticated)/[tenantId]/(diagram)/wfSpec/[...props]/components/SearchVariableDialog.tsx
@@ -1,4 +1,5 @@
 import LinkWithTenant from '@/app/(authenticated)/[tenantId]/components/LinkWithTenant'
+import { wfRunIdToPath } from '@/app/utils/wfRun'
 import { SearchFooter } from '@/app/(authenticated)/[tenantId]/components/SearchFooter'
 import { SEARCH_DEFAULT_LIMIT } from '@/app/constants'
 import { getTypedVariableValue, getVariableDefType } from '@/app/utils/variables'
@@ -100,16 +101,18 @@ export const SearchVariableDialog: FC<Props> = ({ spec }) => {
         ) : data?.pages[0].results.length ? (
           data?.pages.map((page, i) => (
             <Fragment key={i}>
-              {page.results.map(variableId => (
-                <div key={variableId.name}>
-                  <LinkWithTenant
-                    className="py-2 text-blue-500 hover:underline"
-                    href={`/wfRun/${variableId.wfRunId?.id}?threadRunNumber=${variableId.threadRunNumber}`}
-                  >
-                    {variableId.wfRunId?.id}
-                  </LinkWithTenant>
-                </div>
-              ))}
+              {page.results.map(variableId =>
+                variableId.wfRunId ? (
+                  <div key={`${wfRunIdToPath(variableId.wfRunId)}-${variableId.threadRunNumber}`}>
+                    <LinkWithTenant
+                      className="py-2 text-blue-500 hover:underline"
+                      href={`/wfRun/${wfRunIdToPath(variableId.wfRunId)}?threadRunNumber=${variableId.threadRunNumber}`}
+                    >
+                      {wfRunIdToPath(variableId.wfRunId)}
+                    </LinkWithTenant>
+                  </div>
+                ) : null
+              )}
             </Fragment>
           ))
         ) : (
--- a/dashboard/src/app/(authenticated)/[tenantId]/components/Header.tsx
+++ b/dashboard/src/app/(authenticated)/[tenantId]/components/Header.tsx
@@ -1,13 +1,25 @@
 'use client'
 
+import { getServerVersion } from '@/app/actions/getServerVersion'
+import { useWhoAmI } from '@/contexts/WhoAmIContext'
 import LhLogo from '@/littlehorse.svg'
 import { FC } from 'react'
+import useSWR from 'swr'
 
 import LinkWithTenant from './LinkWithTenant'
 import { Principal } from './Principal'
 import { TenantSelector } from './TenantSelector'
 
 export const Header: FC = () => {
+  const { tenantId } = useWhoAmI()
+  const { data: serverVersion, error: serverVersionError } = useSWR(
+    tenantId ? (['serverVersion', tenantId] as const) : null,
+    async ([_cacheKey, tenantIdForRequest]) => getServerVersion(tenantIdForRequest)
+  )
+
+  const serverLabel =
+    serverVersionError !== undefined ? '—' : serverVersion === undefined ? '…' : serverVersion
+
   return (
     <nav className="mb-4 border-gray-200 bg-black px-8 dark:bg-gray-900">
       <div className="mx-auto flex flex-wrap items-center justify-between py-4">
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: server/src/main/java/io/littlehorse/server/streams/topology/core/processors/CommandProcessor.java [hunk 1]
```diff
@@ -173,41 +175,59 @@ public void close() {
...
 
-        if (metricsOnCurrentPartition != null) {
-            for (AggregateWfMetricsModel aggregateWfMetrics : metricsOnCurrentPartition.buildWfRepartitionCommands()) {
-                forwardMetricSubcommand(aggregateWfMetrics);
+        long startTime = timestamp - (10 * 60 * 1000L); // 10 minutes ago
+        String startPrefix = PartitionMetricWindowModel.STORE_KEY_PREFIX + "/" + startTime;
+        String endPrefix = PartitionMetricWindowModel.STORE_KEY_PREFIX + "/~";
+
+        if (!metricsHintUsed) {
+            metricsHintUsed = true;
```
eduwercamacaro: I think we need to set back to false on the `close` method

server/src/main/java/io/littlehorse/server/streams/topology/core/processors/CommandProcessor.java
```diff
@@ -173,41 +175,59 @@ public void close() {
...
+                if (hintTime < startTime) {
+                    startPrefix = PartitionMetricWindowModel.STORE_KEY_PREFIX + "/" + hintTime;
+                }
             }
-            for (AggregateTaskMetricsModel aggregateTaskMetrics :
-                    metricsOnCurrentPartition.buildTaskMetricRepartitionCommand()) {
-                forwardMetricSubcommand(aggregateTaskMetrics);
+        }
+
+        try (LHKeyValueIterator<PartitionMetricWindowModel> iter =
```
eduwercamacaro, line 240: We should limit the amount of time this function can iterate over the stores. By doing so, we can prevent any transaction timeouts that may occur when there are several tenants with metrics.

schemas/internal/storage.proto
```diff
@@ -141,11 +145,27 @@ message TimerIteratorHint {
...
 message NodeOutput {
   int32 thread_run_number = 1;
   int32 node_run_position = 2;
   string node_name = 3;
   VariableValue value = 4;
   google.protobuf.Timestamp created_at = 5;
+}
+
+message PartitionMetricWindow {
+  oneof id {
```
eduwercamacaro: For consistency with the other protobuf objects, it is better to use a dedicated object ID for this new storable.

schemas/internal/storage.proto
```diff
@@ -141,11 +145,27 @@ message TimerIteratorHint {
...
+    WfSpecId  wf_spec_id  = 1;
+    TaskDefId task_def_id = 2;
+    UserTaskDefId user_task_def_id = 3;
+  }
+  TenantId tenant_id = 5;
```
eduwercamacaro: I think the `tenant_id` should be part of the object id as it is contained in the partitionKey

server/src/main/java/io/littlehorse/common/model/PartitionMetricWindowModel.java
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    @Override
+    public Class<PartitionMetricWindow> getProtoBaseClass() {
+        return PartitionMetricWindow.class;
+    }
+
+    @Override
+    public String getStoreKey() {
+        String idPart = "";
```
eduwercamacaro: This is where we should use an object id instead. see my comment [here](https://github.com/littlehorse-enterprises/littlehorse/pull/2071#discussion_r2828975774)

schemas/internal/storage.proto
l4crito: Are Storeables allowed to have an object ID? I thought object IDs were only for Gettables.

The current flow is:

PartitionMetricWindow is a Storeable.(This is what is getting stored in each partition)

MetricWindow is a Gettable with an object ID, This will contain the accumulation of all the partitons and this is what is getting called by the rpc

schemas/internal/command.proto
```diff
@@ -115,6 +116,11 @@ message AggregateWfMetrics {
     repeated WfMetricUpdate metric_updates = 3;
 }
 
+message AggregateWindowMetrics {
+    TenantId tenant_id = 2;
```
coltmcnealy-lh: Can't this be inferred by the kafka streams record headers? This would be the first time that we're passing around `TenantId`

schemas/internal/storage.proto
coltmcnealy-lh: I sort of agree with Eduwer here. Putting it into a dedicated messsage would be best practice for proto message design, even though this is just a `Storeable`.

In fact, for this update, you could just reuse the actual object id of the Getable itself.

schemas/internal/storage.proto
coltmcnealy-lh: Oh, I see...not in the `partitionKey` but rather in the RocksDB key. See previous comment, this makes me grumpy that we're not using the `TenantScopedStore` here but rather the `ClusterScopedStore` because now we have to manage tenant id's manually (this is the firsst place where we do that).

I won't block the review over it, though.

schemas/littlehorse/metrics.proto
```diff
@@ -0,0 +1,43 @@
...
+option csharp_namespace = "LittleHorse.Sdk.Common.Proto";
+
+message MetricWindowId {
+  oneof id {
+    WfSpecId  wf_spec_id  = 1;
+    TaskDefId task_def_id = 2;
+    UserTaskDefId user_task_def_id = 3;
+  }
+  google.protobuf.Timestamp window_start = 5;
+  MetricWindowType          metric_type  = 6;
```
coltmcnealy-lh: This shouldn't be needed because you can infer it from the `id` oneof field, right?

schemas/littlehorse/metrics.proto
```diff
@@ -0,0 +1,43 @@
...
+message CountAndTiming {
+  int32 count            = 1;
+  int64 min_latency_ms   = 2;
+  int64 max_latency_ms   = 3;
+  int64 total_latency_ms = 4;
+}
+
+message MetricWindow {
+  MetricWindowId              id      = 1;
+  map<string, CountAndTiming> metrics = 2;
```
coltmcnealy-lh: Is there a way we could make the metric names hard-coded into the protobuf itself rather than having the strings be dynamic? I know we discussed this offline (see my closed proposal for metrics)...I know we simplify the implementation by being dynamic but it makes it harder for the client side.

That said, the client is the dashboard, so maybe that's a discussion for you and bryson.

l4crito: Let me think on this

schemas/littlehorse/metrics.proto
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// Workflow-level aggregates for lifecycle transitions; fields are `CountAndTiming`.
+message WfMetrics{
+  CountAndTiming started = 1;
+  CountAndTiming running_to_completed = 2;
+  CountAndTiming running_to_error = 3;
+  CountAndTiming running_to_exception = 4;
+  CountAndTiming running_to_halting = 5;
+  CountAndTiming halting_to_halted = 6;
```
coltmcnealy-lh: Also add `CountAndTiming halted_to_running = 7;`

PR #2074, opened 2026-02-13
feat(server): allow waking up parent wfRun when child is rescued

coltmcnealy-lh:


server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/WfRunModel.java
```diff
@@ -576,6 +584,13 @@ public void rescueThreadRun(int threadRunNumber, boolean skipCurrentNode, CorePr
         validateCanRescueThreadRun(threadRunNumber, ctx);
         ThreadRunModel toRescue = getThreadRun(threadRunNumber);
         toRescue.rescue(skipCurrentNode, ctx);
+
+        // If this WfRun is itself a child and we rescued its entrypoint, re-open the
+        // parent WAIT_FOR_CHILD_WF node when the failure is still unhandled.
+        if (threadRunNumber == 0 && parentTrigger != null && parentTrigger.getWaitingNodeRun() != null) {
+            WfRunModel parentWfRun = ctx.getableManager().get(id.getParentWfRunId());
+            parentWfRun.rescueThreadRun(parentTrigger.getWaitingNodeRun().getThreadRunNumber(), false, ctx);
```
eduwercamacaro, line 592: I believe this behavior requires a proposal, as it is an important decision we are making regarding rescue commands. We automatically rescue parent workflows from a child; this is different from what we have right now. At first, this scenario sounds to me like two different commands that the users should intentionally send. IMO, child workflows should care nothing about the parent lifecycle.

PR #2080, opened 2026-02-16
feat(server): Edge condition refactor

eduwercamacaro:
implements https://github.com/littlehorse-enterprises/littlehorse/pull/2033

### DIFF
--- a/sdk-python/littlehorse/workflow.py
+++ b/sdk-python/littlehorse/workflow.py
@@ -152,7 +152,7 @@ def to_variable_assignment(value: Any) -> VariableAssignment:
         return VariableAssignment(
             expression=VariableAssignment.Expression(
                 lhs=to_variable_assignment(expression.lhs()),
-                operation=expression.operation(),
+                mutation_type=expression.operation(),
                 rhs=to_variable_assignment(expression.rhs()),
             )
         )


examples/java/basic/src/main/java/io/littlehorse/examples/BasicExample.java
coltmcnealy-lh, line 1: Not sure we should edit this file. Editing conditionals example is a great one. Maybe another one `example-boolean-logic` too 😄

schemas/littlehorse/common_wfspec.proto
```diff
@@ -43,8 +43,13 @@ message VariableAssignment {
   message Expression {
     // The left-hand-side of the expression.
     VariableAssignment lhs = 1;
-    // The operator in the expression.
-    VariableMutationType operation = 2;
+
+    oneof operation {
```
coltmcnealy-lh, line 54: A comment here explaining the `oneof`, and that it's for backwards compatibility, and how it works with `UNSET` would be great

schemas/littlehorse/wf_spec.proto
```diff
@@ -441,7 +441,10 @@ message Node {
 // There is no output.
 message WaitForConditionNode {
   // The condition that this node will block for.
-  EdgeCondition condition = 1;
+  oneof node_condition {
```
coltmcnealy-lh, line 447: Same here 😄

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/WorkflowCondition.java
coltmcnealy-lh, line 1: Can we remove this class?

schemas/littlehorse/common_wfspec.proto
eduwercamacaro: Well `UNSET` is not a valid value here. We control backwards compatibility with LegacyEdgeCondition

PR #2081, opened 2026-02-17
fix(server): improve StructDef validation message chain

Snarr:
Throws LHApiException at the top level of StructDef validation and catches internal `StructDefValidationException`s for lower levels. Combines context from lower level exceptions to give the user a more accurate reading of where their Struct error is coming from. 

## Before
For example, before a `StructFieldDef` validation exception might have read:

`INVALID_ARGUMENT: StructFieldDef field type JSON_OBJ is not compatible with the provided default value.`

## After
Now, a `StructFieldDef` validation exception will include the full context of which StructField the error occurred in:

`INVALID_ARGUMENT: StructDef validation failed: StructDef field 'createdAt' invalid: StructFieldDef field type JSON_OBJ is not compatible with the provided default value.`

## Other

PR also handles null case in Java SDK

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/ValidateStructDefEvolutionRequestModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/ValidateStructDefEvolutionRequestModel.java
@@ -6,7 +6,7 @@ import io.littlehorse.common.LHSerializable;
 import io.littlehorse.common.exceptions.LHApiException;
 import io.littlehorse.common.model.getable.global.structdef.InlineStructDefModel;
 import io.littlehorse.common.model.getable.global.structdef.StructDefModel;
-import io.littlehorse.common.model.getable.global.structdef.StructValidationException;
+import io.littlehorse.common.model.getable.global.structdef.StructDefValidationException;
 import io.littlehorse.common.model.getable.objectId.StructDefIdModel;
 import io.littlehorse.common.util.InlineStructDefUtil;
 import io.littlehorse.sdk.common.exception.LHSerdeException;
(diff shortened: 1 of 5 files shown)


sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/FixedSpawnedThreads.java
```diff
@@ -23,7 +24,8 @@ public WaitForThreadsNode buildNode(WaitForThreadsStrategy strategy) {
         WaitForThreadsNode.Builder waitNode = WaitForThreadsNode.newBuilder().setStrategy(strategy);
         for (SpawnedThread spawnedThread : spawnedThreads) {
             WfRunVariableImpl threadNumberVariable = (WfRunVariableImpl) spawnedThread.getThreadNumberVariable();
-            if (!threadNumberVariable.getTypeDef().getPrimitiveType().equals(VariableType.INT)) {
+            if (threadNumberVariable.getTypeDef().getDefinedTypeCase() != DefinedTypeCase.PRIMITIVE_TYPE
```
coltmcnealy-lh, line 27: Shouldn't we support casting here? As in `.isCompatibleWith()`?

coltmcnealy-lh: Oh, sorry, silly comment. This is in the WFSDK not in the Server Side.

Snarr: Yep, sorry this was little change I had to sneak in that doesnt quite make sense in a server PR but isnt big enough to justify its own 😅

PR #2086, opened 2026-02-20
fix(test-utils): Cleans up the worker from the context after each execution

MateoRojas:
The test-utils library is closing the worker after each test execution but it's not removing it from the test context store. When another test uses the same worker, it cannot because it's closed and it cannot recreate it one because it's in the test context store.

image

image


coltmcnealy-lh (approved):
I like it, but I think we're doing way too much in the `BasicTest`. IMO that test should just verify that we can run a single `WfRun` with a single task in it. It's to say "hey is the server working?"

PR #2088, opened 2026-02-23
fix(sdk-java): reopen poll thread after error

coltmcnealy-lh:


sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/PollTaskStub.java
```diff
@@ -77,6 +77,7 @@ public void acquireNextPermit() throws InterruptedException {
     @Override
     public void close() {
         try {
+            closed.set(true);
```
eduwercamacaro, line 80: This makes sense. Thanks!

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/RebalanceThread.java
```diff
@@ -152,6 +159,17 @@ private void maybeCleanupDanglingNonRunningThreads(List<LHHostInfo> availableHos
...
+
+            for (LHHostInfo deadHost : deadHosts) {
+                List<PollThread> threadsOnDeadHost = runningConnections.get(deadHost);
+                for (PollThread threadOnDeadHost : threadsOnDeadHost) {
+                    if (threadOnDeadHost.isRunning()) {
+                        threadOnDeadHost.interrupt();
+                        threadOnDeadHost.close();
+                    }
+                }
+                runningConnections.remove(deadHost);
```
eduwercamacaro, line 171: I don't see a reason for this because the `cleanupDanglingThreadsFromAvailableHosts` method already removes the host with no active connection.

coltmcnealy-lh: Check it again. That method only removes poll threads when you no longer need to look at that host.

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/PollTaskStub.java
coltmcnealy-lh: Actually I dont think this line had an impact. I just put it there for completeness

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/RebalanceThread.java
eduwercamacaro: Ok I see.
I thought `deadHost` was a list of hosts with no running threads, but it's actually a list of all the hosts with at least one thread that is not currently running. 

Here, I think we can do better. and remove the inactive thread as soon as we detect it in order to prevent repeating the iteration over the `deadHosts` (see line 143). Although we can't eliminate an element from the list we are iterating, we can utilize some Streams API tools or make a copy of the list before iterating so you can remove an element in the original `List`.

I know that this is more of a code style thing, and the code already solves the problem of removing inactive threads. So feel free to ignore this comment if you want.

eduwercamacaro (approved):

PR #2101, opened 2026-03-03
fix(server): fix broken masking feature on Structs

Snarr:
This PR fixes the broken Masked Values feature on Structs and in doing so updates the `LHStructField` annotation to better support fields & getter/setter methods.

### DIFF
--- a/sdk-go/littlehorse/task_worker_internal.go
+++ b/sdk-go/littlehorse/task_worker_internal.go
@@ -2,6 +2,7 @@ package littlehorse
 
 import (
 	"context"
+	"fmt"
 	"log"
 	"reflect"
 	"runtime/debug"
@@ -11,6 +12,8 @@ import (
 
 	"github.com/littlehorse-enterprises/littlehorse/sdk-go/lhproto"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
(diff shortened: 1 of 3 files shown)


server/src/main/java/io/littlehorse/common/model/getable/core/variable/StructModel.java
```diff
@@ -20,20 +27,59 @@ public class StructModel extends LHSerializable<Struct> implements Comparable<St
...
     public void initFrom(Message proto, ExecutionContext context) throws LHSerdeException {
         Struct p = (Struct) proto;
 
         this.structDefId = StructDefIdModel.fromProto(p.getStructDefId(), context);
         this.inlineStruct = InlineStructModel.fromProto(p.getStruct(), InlineStructModel.class, context);
+        this.context = context;
+        this.maskOnSerialize = context != null && context.support(RequestExecutionContext.class);
+    }
+
+    private InlineStruct.Builder maskedInlineStructProto() {
```
eduwercamacaro: At first, I thought this method was always masking the struct's content. 

can we name it like `maybeMask...`?

server/src/main/java/io/littlehorse/common/model/getable/core/variable/StructModel.java
```diff
@@ -20,20 +27,59 @@ public class StructModel extends LHSerializable<Struct> implements Comparable<St
...
+
+        return maskedStruct;
+    }
+
+    private boolean shouldMaskField(String fieldName) {
+        if (!maskOnSerialize || context == null || !context.support(RequestExecutionContext.class)) {
+            return false;
+        }
+
+        StructDefModel structDef = context.service().getStructDef(structDefId);
```
eduwercamacaro: This `getStructDef` method can be only called once for the entire iteration. I know that this ends up using the cache, but still it is worth saving extra transformation from `proto` to `model`

PR #2104, opened 2026-03-03
fix(sdk-java, sdk-go): throw exception if TaskDef not found

Snarr:
Throws exception if TaskDef not found on server to avoid worker crashing and NPE


eduwercamacaro:
Wondering if we should also cover this on the server side to prevent an NPE if the task def is missing.

sdk-go/littlehorse/task_worker_internal.go
```diff
@@ -193,7 +195,26 @@ func (controller *serverConnectionManager) notifyCallSuccess(response *lhproto.R
 	controller.workerHealthy = true
 }
 
-func (m *serverConnectionManager) start() {
+func (m *serverConnectionManager) validateTaskDef() error {
+	_, err := (*m.tw.grpcStub).GetTaskDef(context.Background(), m.tw.taskDefId)
```
coltmcnealy-lh: In Java we used an await loop with retries because after `.registerTaskDef()` there is some time for the metadata to propagate through the cluster. Do we need to add a wait() here?

coltmcnealy-lh (approved):

PR #2112, opened 2026-03-04
feature(sdk-dotnet): casting

l4crito:
## Add Explicit Casting Support to .NET SDK

LittleHorse workflows rely on strong typing for variables and task inputs. While some conversions can be inferred automatically by the server, others require explicit intent from the developer. This change adds casting helpers directly to the .NET workflow DSL, allowing developers to explicitly control type conversions during workflow definition.

## What’s Included

The .NET SDK now provides methods for explicit casting on workflow variables, node outputs, and expression results, allowing users to convert values to a target `VariableType`.

Supported casting helpers include:

- `CastToInt()`
- `CastToDouble()`
- `CastToBool()`
- `CastToString()`

These methods can be used anywhere a variable, expression, JSON path result, or node output is accepted as input to another node.

## Example

```csharp
var stringInput = wf.DeclareStr("string-number").WithDefault("3.14");

// STR -> DOUBLE (explicit)
var doubleResult = wf.Execute(
    "double-method",
    stringInput.CastTo(VariableType.Double)
);

// DOUBLE -> INT (explicit)
var intResult = wf.Execute(
    "int-method",
    doubleResult.CastToInt()
);

// JsonPath + cast
wf.Execute(
    "int-method",
    jsonInput.WithJsonPath("$.int").CastToInt()
);

### CODE REVIEW

File: examples/dotnet/casting/Program.cs [hunk 1]
```diff
@@ -0,0 +1,114 @@
...
+    {
+        void Entrypoint(WorkflowThread wf)
+        {
+            var stringInput = wf.DeclareStr("string-number").WithDefault("3.14");
+            var stringBool = wf.DeclareStr("string-bool").WithDefault("false");
+            var jsonInput = wf.DeclareJsonObj("json-input")
+                .WithDefault(new Dictionary<string, object> { { "int", "1" }, { "string", "hello" } });
+
+            // STR -> DOUBLE (explicit)
+            var doubleResult = wf.Execute("double-method", stringInput.CastTo(VariableType.Double));
```
coltmcnealy-lh, line 55: Hmm. Can we use a `.class` instead of using `VariableType`? As much as possible I'd like to avoid using that proto enum in our SDK's public API.

l4crito: sure, let me try it , but other sdks uses that enum, amm may be remove the `castTo(enum)` method? and let only this guys alive 
CastToInt()
CastToDouble()
CastToBool()
CastToString()

PR #2127, opened 2026-03-06
chore(pipeline): split server e2e and unit tests for speedup

coltmcnealy-lh:


.github/workflows/tests.yml
```diff
@@ -101,7 +101,7 @@ jobs:
       - name: Tests
         run: ./gradlew canary:spotlessCheck canary:test canary:build
 
-  tests-server:
+  tests-server-unit:
```
mijailr, line 104: As you're changing the test name, would be required to update the branch protection rules.

coltmcnealy-lh: Yes, and add the e2e as a required check too (:

PR #2132, opened 2026-03-07
feat(sdk-java): allow interaction with raw `InlineStruct`

coltmcnealy-lh:


examples/java/inline-struct-placeholder/src/main/java/io/littlehorse/examples/InlineStructPlaceholderExample.java
```diff
@@ -0,0 +1,158 @@
...
+        Runtime.getRuntime()
+                .addShutdownHook(new Thread(() -> workers.forEach(worker -> {
+                    log.debug("Closing {}", worker.getTaskDefName());
+                    worker.close();
+                })));
+
+        return workers;
+    }
+
+    private static void registerInlineStructDef(LittleHorseBlockingStub client) {
```
Snarr: Can we build this using a class and the LHStructDef annotation?

coltmcnealy-lh: That might cause confusion. The whole point of this example is to use structs without that annotation

Snarr: I think the same can be said vice versa, I don't think the README makes it clear why it is being registered this way

Snarr: Also we only have 2 total examples now that use StructDefs. I want to make it abundantly clear to users that this is the _harder_ way to use StructDefs and in what scenario it would be valuable

examples/java/inline-struct-placeholder/src/main/java/io/littlehorse/examples/InlineStructPlaceholderExample.java
```diff
@@ -0,0 +1,158 @@
...
+        PutStructDefRequest request = PutStructDefRequest.newBuilder()
+                .setName(CUSTOMER_STRUCT)
+                .setDescription("Inline customer StructDef used by placeholder example")
```
Snarr: We should probably remove the word `inline` here, a StructDef's a StructDef right ?

examples/java/inline-struct-placeholder/src/main/java/io/littlehorse/examples/InlineStructPlaceholderExample.java
coltmcnealy-lh: Okay, the README does need to improve. I'll work on that and add some disclaimers. That said, it is probably useful to show users how they can rawdog it and create structdefs on their own.

examples/java/inline-struct-placeholder/src/main/java/io/littlehorse/examples/InlineStructPlaceholderExample.java
coltmcnealy-lh: Yes, agreed.

examples/java/inline-struct-placeholder/README.md
```diff
@@ -6,6 +6,8 @@ This example shows how to:
 - Accept a raw `InlineStruct` input with `@LHType(structDefName = ...)`.
 - Resolve `@LHTaskMethod` and `@LHType(structDefName = ...)` placeholders from a single map.
 
+**_NOTE_: This example is for advanced use-cases only, in which you don't know the exact struct you're using when writing the code. In almost all cases, it's simpler and better to use `@LHStructDef` annotation as shown in the [normal Struct example](../struct-def/README.md)**
```
Snarr, line 9: Nice :)

examples/java/inline-struct-placeholder/README.md
```diff
@@ -8,6 +8,8 @@ This example shows how to:
 
 **_NOTE_: This example is for advanced use-cases only, in which you don't know the exact struct you're using when writing the code. In almost all cases, it's simpler and better to use `@LHStructDef` annotation as shown in the [normal Struct example](../struct-def/README.md)**
 
+**_NOTE_: This feature is EXPERIMENTAL in the 1.0.0 release**
```
Snarr, line 11: 👍

PR #2135, opened 2026-03-09
feat(server): allow runtime injection of external jars on application startup

MateoRojas:
### Summary
This PR updates the server module’s Gradle build to customize the generated start scripts so they accept an optional `EXTRA_CLASSPATH` environment variable. This enables attaching additional jars (e.g., `aws-msk-iam-auth`) at runtime without rebuilding the image.

### Background & Motivation
When connecting Kafka clients (Producer/Streams) to AWS MSK using `AWS_MSK_IAM`, the `aws-msk-iam-auth` jar must be available on the application classpath. While the recommended approach would be to add it as a Gradle `runtimeOnly` dependency, this cannot be done every time a customer needs to add a differen runtime dependency to connect to their Kafka cluster. This change provides a simple, reproducible way to prepend an external classpath via environment variable.

### What Changed
- File: `server/build.gradle`
  - Modified the generated Unix and Windows start scripts to prepend `EXTRA_CLASSPATH` to the application classpath if provided.

### Implementation Details
- Unix script now sets:
  - `CLASSPATH="${EXTRA_CLASSPATH:-}:$APP_HOME/lib/"` (prepends `EXTRA_CLASSPATH` when present; defaults to empty string otherwise)
- Windows script now sets:
  - `set CLASSPATH=%EXTRA_CLASSPATH%;%APP_HOME%\lib\`
- Behavior is fully backward‑compatible: if `EXTRA_CLASSPATH` is not set, startup behavior remains unchanged.


coltmcnealy-lh (approved):
Seems fine; our security model assumes that you can trust the filesystme that's running the container.

PR #2140, opened 2026-03-11
docs(proto): add comments to all messages

Snarr:
Adds comments to all RPCs, Messages, and Enums that were missing them.

## Breaking Changes

This updates the `SearchPrincipalRequest` `principal_criteria` `oneof` cases to use snake case instead of camel case, which is a breaking change on any users depending on this request. We don't have to make this change, but if we think it is low risk it would be nice to push this in 1.0

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streams/lhinternalscan/publicrequests/SearchPrincipalRequestModel.java
+++ b/server/src/main/java/io/littlehorse/server/streams/lhinternalscan/publicrequests/SearchPrincipalRequestModel.java
@@ -73,10 +73,10 @@ public class SearchPrincipalRequestModel
         if (latestStart != null) builder.setLatestStart(LHUtil.fromDate(latestStart));
 
         switch (type) {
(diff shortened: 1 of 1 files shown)


schemas/littlehorse/common_wfspec.proto
```diff
@@ -207,8 +207,10 @@ message TypeDefinition {
   // Note: if no `defined_type` is set, defaults to a `primitive_type` of `JSON_OBJ`
   // to support old clients from 0.14.1 or lower.
   oneof defined_type {
+    // The primitive type definition.
     VariableType primitive_type = 1;
-
+    
+    // The ID of the StructDef that defines this type.
     StructDefId struct_def_id = 5;
```
coltmcnealy-lh, line 214: This makes me wonder whether we can support just passing a `structDefName` so that it can handle multiple versions of the `StructDef`.

Snarr: Good idea

PR #2141, opened 2026-03-12
feat(server): implement `Quota` proposal

coltmcnealy-lh:
Implements the proposal in #2142 . Heavily aided by opencode with GPT 5.4


coltmcnealy-lh:
This repo includes tests in other SDK's: https://github.com/coltmcnealy-lh/lh-quotas-test

sdk-python/littlehorse/client_interceptors.py
coltmcnealy-lh, line 1: I'm not sure about this right here. I think maybe it should be renamed to `grpc_interceptors.py`...?

server/src/main/java/io/littlehorse/server/quotas/QuotaState.java
```diff
@@ -0,0 +1,80 @@
...
+        double numberOfWindowsToWait = 1.0 + ((double) permitDebt / permitsPerWindow);
+        long delayMs = (long) (WINDOW_MS * numberOfWindowsToWait);
+        return delayMs;
+    }
+
+    private void maybeRefreshWindow(QuotaModel quota, int serverCount) {
+        long nowNanos = System.nanoTime();
+
+        // The Quota or the number of servers may have changed!
+        double requestsPerSec = Math.max(1.0, quota.getWriteRequestsPerSecond() / (double) serverCount);
```
l4crito, line 53: I am understanding here that:  if i have a quota for `tenatn-a` set to 300 req/s with 3 servers each server can handle 100 req/s only, is that correct?
Do we have a load balancer on top of servers so we know that the requests are evenly distributed?

coltmcnealy-lh: Yes, that first part is correct. The second part is correct, and we mostly do have a load balancer; in fact, for task workers, we enforce load balancing because the `rpc ReportTask` is sent directly to the server that the task lives on.

There is a slight tradeoff for applications where there's just a single client running workflows, because in that case the `rpc RunWf` would be lopsided, but the task workers would do most of the heavy lifting, and they're spread around.

proposals/019-quotas.md
```diff
@@ -0,0 +1,207 @@
+# Request Quotas
```
eduwercamacaro, line 1: Is this a duplicated file?

schemas/littlehorse/acls.proto
```diff
@@ -50,6 +49,28 @@ message Tenant {
...
+    int32 write_requests_per_second = 3;
+}
+
+// Identifies a `Quota`.
+message QuotaId {
+    // The governed `Tenant`.
+    TenantId tenant = 1;
+
+    // If unset, the quota applies to all `Principal`s in the `Tenant`.
+    PrincipalId principal = 2;
```
eduwercamacaro: Proposal says this is optional. I think we need to do that in order to support tenant-specific quotas

schemas/littlehorse/acls.proto
```diff
@@ -197,4 +222,22 @@ message PutTenantRequest {
     optional OutputTopicConfig output_topic_config = 2;
 }
 
+// Creates or updates a `Quota`.
+message PutQuotaRequest {
+    // The governed `Tenant`.
+    TenantId tenant = 1;
+
+    // If unset, the quota applies to all `Principal`s in the `Tenant`.
+    PrincipalId principal = 2;
```
eduwercamacaro, line 231: Same comment as [here](https://github.com/littlehorse-enterprises/littlehorse/pull/2141/changes#r3119212295).

proposals/019-quotas.md
coltmcnealy-lh: This was a poorly-done merge conflict. Thanks.

schemas/littlehorse/acls.proto
```diff
@@ -50,6 +49,28 @@ message Tenant {
     optional OutputTopicConfig output_topic_config = 3;
 }
 
+// A Quota defines limits for requests made within a certain `Tenant`, optionally
+// scoped to one specific `Principal`.
+message Quota {
```
eduwercamacaro, line 54: I think it is worth adding quota-specific objects into its own `quota.proto` file

schemas/littlehorse/acls.proto
```diff
@@ -50,6 +49,28 @@ message Tenant {
...
+
+    // The time at which the Quota was created.
+    google.protobuf.Timestamp created_at = 2;
+
+    // The maximum number of mutating unary gRPC requests allowed per second.
+    int32 write_requests_per_second = 3;
+}
+
+// Identifies a `Quota`.
+message QuotaId {
```
eduwercamacaro, line 66: I'm not fully convinced that this should be a composite ID between tenant and principal.

While I understand a quota governs a principal within a tenant or a tenant itself, I think this `QuotaId` object limits the API by preventing it from modifying existing quota tenant or principal data. 

If this is intended, then we should mention it in the proposal, as this might introduce some friction when working with quotas. 

Picture a scenario where a user wants to change the quota information from one principal to another. With this structure, the user would need to first delete the quota for the previous principal and create a new one for the other principal.

An alternative would be to create a unique identifier that users could use to mutate the data for a quota.

schemas/littlehorse/service.proto
```diff
@@ -1141,6 +1154,30 @@ message PrincipalIdList {
...
+// Search for Quotas based on certain criteria.
+message SearchQuotaRequest {
+  // Bookmark for cursor-based pagination; pass if applicable.
+  optional bytes bookmark = 1;
+
+  // Maximum results to return in one request.
+  optional int32 limit = 2;
+
+  // If set, return only Quotas for this Tenant.
+  optional string tenant_id = 3;
```
eduwercamacaro, line 1166: I think `tenant_id` should not be optional here because users will always need to specify the tenant they are querying.

canary/src/main/java/io/littlehorse/canary/metronome/MetronomeRunWfExecutor.java
```diff
@@ -45,7 +45,7 @@ public class MetronomeRunWfExecutor implements HealthStatusBinder {
     private final int sampleRate;
     private ScheduledFuture<?> scheduledFuture;
     private final Semaphore inflightRequests;
-    private final Deadline expiredDeadline = Deadline.after(0, TimeUnit.NANOSECONDS);
+    private final Deadline expiredDeadline = Deadline.after(30, TimeUnit.SECONDS);
```
eduwercamacaro, line 48: I don't think canary should do transparent retries when quota is exceeded. Instead, canary should report how many `RESOURCE_EXHAUSTED ` responses it got in a specific period.

schemas/littlehorse/service.proto
coltmcnealy-lh: Ah, it lets you search quotas across all tenants. It's actually pretty useful sometimes.

canary/canary.properties
coltmcnealy-lh, line 1: Oops, I should probably revert this. This was used to help me debug some issues I had.

schemas/littlehorse/acls.proto
```diff
@@ -37,7 +37,6 @@ message Principal {
 // associated data (WfSpec, WfRun, TaskDef, TaskRun, NodeRun, etc) are scoped to within
 // a Tenant.
 //
-// Future versions will include quotas on a per-Tenant basis.
```
eduwercamacaro, line 40: It didn't 😂

sdk-java/src/main/java/io/littlehorse/sdk/common/config/retryinterceptor/ResourceExhaustedRetryInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+import io.grpc.protobuf.StatusProto;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+
```
l4crito, line 16: Could this be a breaking change? 
What will happen if customers are using old versions of the sdk and the server already has quotas?

server/src/main/java/io/littlehorse/common/model/getable/global/acl/PrincipalModel.java
```diff
@@ -151,4 +151,8 @@ public List<TenantIdModel> getTenantsThatPrincipalHasPermissionOver() {
     public boolean canCreateTenants() {
         return globalAcls.allows(ACLResource.ACL_TENANT, ACLAction.WRITE_METADATA);
     }
+
+    public boolean canEditQuotas() {
+        return isAdmin() || globalAcls.allows(ACLResource.ACL_QUOTA, ACLAction.WRITE_METADATA);
```
eduwercamacaro, line 156:
```java
globalAcls.allows(ACLResource.ACL_QUOTA, ACLAction.WRITE_METADATA)
```

I believe that part is not necessary because it should already be being validated at the interceptor level.

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutQuotaRequestModel.java
```diff
@@ -0,0 +1,101 @@
...
+        return PutQuotaRequest.class;
+    }
+
+    @Override
+    public Quota process(MetadataProcessorContext context) {
+        MetadataManager metadataManager = context.metadataManager();
+        PrincipalModel caller =
+                context.service().getPrincipal(context.authorization().principalId());
+
+        if (!caller.canEditQuotas()) {
```
eduwercamacaro, line 69: This is being done at interceptor level, am I missing something?

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/DeleteQuotaRequestModel.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public Class<DeleteQuotaRequest> getProtoBaseClass() {
+        return DeleteQuotaRequest.class;
+    }
+
+    @Override
+    public Empty process(MetadataProcessorContext context) {
+        PrincipalModel caller =
+                context.service().getPrincipal(context.authorization().principalId());
+        if (!caller.canEditQuotas()) {
```
eduwercamacaro, line 46: Same as [here](https://github.com/littlehorse-enterprises/littlehorse/pull/2141/changes#r3120241768)

server/src/main/java/io/littlehorse/server/GlobalExceptionHandler.java
```diff
@@ -30,13 +30,11 @@ public <ReqT, RespT> ServerCall.Listener<ReqT> interceptCall(
     private static class InternalCallListener<ReqT, RespT>
             extends ForwardingServerCallListener.SimpleForwardingServerCallListener<ReqT> {
         private final ServerCall<ReqT, RespT> call;
-        private final Metadata metadata;
 
         public InternalCallListener(
                 ServerCall<ReqT, RespT> call, ServerCallHandler<ReqT, RespT> next, Metadata metadata) {
             super(next.startCall(call, metadata));
             this.call = call;
-            this.metadata = metadata;
```
eduwercamacaro, line 39: I'm going to need to think more about this.

I believe it is incorrect because it doesn't keep the metadata passed from the interceptor chain.

eduwercamacaro: What this interceptor should do is to append new trailers to the metadata object, not completely replace it. This interceptor is part of a chain, and we should act accordingly.

server/src/main/java/io/littlehorse/server/GlobalExceptionHandler.java
```diff
@@ -45,16 +43,26 @@ public void onHalfClose() {
                 // Proceed with the call
                 super.onHalfClose();
             } catch (LHApiException apiException) {
-                call.close(apiException.getStatus(), metadata);
+                call.close(apiException.getStatus(), trailersFor(apiException));
```
eduwercamacaro, line 46: Why do we need the `trailersFor`? As long as I understand, our `LHApiException` doesn't add any extra trailer to the throwable.

server/src/main/java/io/littlehorse/server/quotas/RequestQuotaManager.java
```diff
@@ -0,0 +1,68 @@
...
+        long retryDelayMs = 0L;
+        if (tenantQuota != null) {
+            retryDelayMs = consumeAndGetDelay(tenantQuota, serverCount);
+        }
+        if (principalQuota != null) {
+            retryDelayMs = Math.max(retryDelayMs, consumeAndGetDelay(principalQuota, serverCount));
+        }
+
+        if (retryDelayMs > 0) {
+            throw StatusProto.toStatusRuntimeException(resourceExhaustedStatus(retryDelayMs));
```
eduwercamacaro, line 46: I'm confused because we pass zero extra metadata trailers here, but our GlobalExceptionHandler interceptor expects the exception to contain extra metadata trailers.

eduwercamacaro: See my comment [here](https://github.com/littlehorse-enterprises/littlehorse/pull/2141/changes#r3120464129)

server/src/main/java/io/littlehorse/server/quotas/RequestQuotaManager.java
```diff
@@ -0,0 +1,68 @@
...
+import io.littlehorse.common.model.getable.global.acl.QuotaModel;
+import io.littlehorse.common.model.getable.objectId.PrincipalIdModel;
+import io.littlehorse.common.model.getable.objectId.QuotaIdModel;
+import io.littlehorse.common.model.getable.objectId.TenantIdModel;
+import io.littlehorse.server.streams.BackendInternalComms;
+import io.littlehorse.server.streams.topology.core.RequestExecutionContext;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class RequestQuotaManager {
```
eduwercamacaro, line 18: There is a lot to test in this class. Can you add test cases that simulate different scenarios for the quota manager?

server/src/test/java/io/littlehorse/server/GlobalExceptionHandlerTest.java
```diff
@@ -44,23 +43,27 @@ public void shouldHandleLHValidationExceptions(Status.Code expectedCode) {
                 .onHalfClose();
         globalExceptionHandler.interceptCall(serverCall, metadata, next).onHalfClose();
```
eduwercamacaro, line 44: I think we can improve test cases here by avoiding passing a mock object to the metadata argument. We can pass an actual Metadata instance.

schemas/littlehorse/acls.proto
eduwercamacaro: As we discussed in our internal call. 

Using the tenant and principal makes things easier when resolving quotas at the interceptor level because we have enough information there to resolve the quota id.

I believe we simply need to document this somewhere.

sdk-java/src/main/java/io/littlehorse/sdk/common/config/retryinterceptor/ResourceExhaustedRetryInterceptor.java
coltmcnealy-lh: I'm not sure that's a "breaking" change, in theory, it will improve the quality of service, enforce our contracts, and allow us to provide a multi-tenant service.

server/src/main/java/io/littlehorse/server/GlobalExceptionHandler.java
coltmcnealy-lh: Yeah, I had doubts about that too...if I'm not the only one having doubts about it, I'll revert this to the old architecture.

canary/src/main/java/io/littlehorse/canary/metronome/MetronomeRunWfExecutor.java
eduwercamacaro: @coltmcnealy-lh Can you elaborate on why this is needed? 

The RunWfExecutor uses this deadline to simply throw an immediate 'DEADLINE_EXCEEDED' error code when it does not receive enough server responses to process new RunWfRequests.

eduwercamacaro: This field does not represent any real impact on regular requests. It only helps to immediately reject requests when the server is stuck or with high delays in the command processor.

canary/canary.properties
eduwercamacaro: 1 request every 15 seconds is a conservative value to not overwhelm local deployments.

eduwercamacaro (approved):

eduwercamacaro:
[suggestion]:

At the proposal level, I would expect to understand what the new error codes are that the canary will report on metrics when it is being throttled.

PR #2148, opened 2026-03-14
refactor(dashboard): support new edge conditions

bryson-g:
image


dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/EdgeTypes/EdgeConditionLabel.tsx
```diff
@@ -0,0 +1,48 @@
...
+          <span className="text-[10px] text-gray-500">{operatorSymbol}</span>
+          {rightOperand != null ? <span className={variableBadgeClass}>{getVariable(rightOperand)}</span> : null}
+        </span>
+      )
+    }
+
+    return <span className={variableBadgeClass}>{getVariable(variableAssignment)}</span>
+  }
+
+  return null
```
Comment (steinsleger) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/EdgeTypes/EdgeConditionLabel.tsx]: Do we have any other cases than `condition` and `legacyCondition`? If not, this return can be removed.

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/EdgeTypes/index.ts
```diff
@@ -1,5 +1,8 @@
 import { EdgeTypes } from 'reactflow'
 import CustomEdge from './Default'
+import { extractEdges } from './extractEdges'
+
+export { extractEdges }
```
Comment (steinsleger) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/EdgeTypes/index.ts]: Do we actually need this?

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/EdgeTypes/EdgeConditionLabel.tsx
Comment (Snarr) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/EdgeTypes/EdgeConditionLabel.tsx]: @steinsleger @bryson-g  These are the only two cases now, but as with any protobuf `oneof` case, we can always add more cases down the road.

I would prefer that for the `default` case (no case specified), we fail fast rather than fail silently. Maybe instead of `null` here we can return something that indicates an unknown case to the user.

PR #2159, opened 2026-03-18
fix(dashboard): use wfrunid build util

bryson-g:
https://github.com/user-attachments/assets/ce910f76-c085-4d24-9f82-4067e9c51195

https://littlehorsecommunity.slack.com/archives/C0628N0HXGU/p1773844437508639?thread_ts=1773771934.424239&cid=C0628N0HXGU

### DIFF
--- a/canary/src/main/resources/log4j2.properties
+++ b/canary/src/main/resources/log4j2.properties
@@ -20,19 +20,19 @@ appender.kafka.layout.pattern = %d{HH:mm:ss} %highlight{%-5p} [KAFKA] %c - %m%n
 appender.canary.type = Console
 appender.canary.name = canary
 appender.canary.layout.type = PatternLayout
-appender.canary.layout.pattern = %d{HH:mm:ss} %highlight{%-5p}{TRACE=blue} [CANARY] %c - %m%n
+appender.canary.layout.pattern = %d{HH:mm:ss} %highlight{%-5p}{TRACE=blue} [%t] [CANARY] %c - %m%n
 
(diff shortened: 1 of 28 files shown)


dashboard/src/app/(authenticated)/[tenantId]/(diagram)/wfSpec/[...props]/components/SearchVariableDialog.tsx
```diff
@@ -100,16 +101,18 @@ export const SearchVariableDialog: FC<Props> = ({ spec }) => {
...
-                    {variableId.wfRunId?.id}
-                  </LinkWithTenant>
-                </div>
-              ))}
+              {page.results.map(variableId =>
+                variableId.wfRunId ? (
+                  <div key={`${wfRunIdToPath(variableId.wfRunId)}-${variableId.threadRunNumber}`}>
+                    <LinkWithTenant
+                      className="py-2 text-blue-500 hover:underline"
+                      href={`/wfRun/${wfRunIdToPath(variableId.wfRunId)}?threadRunNumber=${variableId.threadRunNumber}`}
```
Comment (Magol28) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/wfSpec/[...props]/components/SearchVariableDialog.tsx:109]: if we are creating more links, it would be better to have a constructor or something, so if it changes we can change them all

Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/wfSpec/[...props]/components/SearchVariableDialog.tsx:109]: I will leave this to another issue for refactoring, what solution do you recommend?

PR #2169, opened 2026-03-24
refactor(sdk-java): Refactor TaskDef inference utils

Snarr:
This PR refactors TaskDef inference and Variable mapping internals. It removes legacy builders and redundant code for a more class-based approach.

- Removed the old TaskDefBuilder, replaced everywhere with LHTaskSignature since TaskDefBuilder didn't add much on its own
- Added parameter and return type helpers (LHTaskParameter and LHTaskReturnType)
- Added LHType handling helper  (LHTypeMetadata)
- Updated worker variable mapping to use VariableDef-based assignments and fixed argument wiring so WorkerContext logic is moved outside of VariableMapping

### DIFF
--- a/sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTypeMetadata.java
+++ b/sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTypeMetadata.java
@@ -16,27 +16,33 @@ final class LHTypeMetadata {
     }
 
     private final boolean masked;
-    private final Optional<String> name;
-    private final Optional<String> structDefName;
+    private final String name;
+    private final String structDefName;
 
-    private LHTypeMetadata(boolean masked, Optional<String> name, Optional<String> structDefName) {
+    private LHTypeMetadata(boolean masked, String name, String structDefName) {
         this.masked = masked;
         this.name = name;
         this.structDefName = structDefName;
+    }
+
+    private LHTypeMetadata() {
+        this(false, null, null);
     }
 
     static LHTypeMetadata from(AnnotatedElement element, Map<String, String> placeholderValues) {
         Map<String, String> placeholders = placeholderValues == null ? Map.of() : placeholderValues;
 
         if (!element.isAnnotationPresent(LHType.class)) {
-            return new LHTypeMetadata(false, Optional.empty(), Optional.empty());
+            return new LHTypeMetadata();
         }
 
         LHType typeAnnotation = element.getAnnotation(LHType.class);
 
-        Optional<String> parsedName = normalize(typeAnnotation.name());
-        Optional<String> parsedStructDefName = normalize(typeAnnotation.structDefName())
-                .map(value -> PlaceholderUtil.replacePlaceholders(value, placeholders));
+        String parsedName = normalize(typeAnnotation.name());
+        String parsedStructDefName = normalize(typeAnnotation.structDefName());
+        if (parsedStructDefName != null) {
+            parsedStructDefName = PlaceholderUtil.replacePlaceholders(parsedStructDefName, placeholders);
+        }
 
         return new LHTypeMetadata(typeAnnotation.masked(), parsedName, parsedStructDefName);
     }
@@ -46,31 +52,31 @@ final class LHTypeMetadata {
     }
 
     Optional<String> getName() {
-        return name;
+        return Optional.ofNullable(name);
     }
 
     Optional<String> getStructDefName() {
-        return structDefName;
+        return Optional.ofNullable(structDefName);
     }
 
     void validateStructDefNameUsage(Class<?> javaType, ValidationContext context, String contextName) {
         boolean isInlineStruct = InlineStruct.class.isAssignableFrom(javaType);
 
-        if (isInlineStruct && structDefName.isEmpty()) {
+        if (isInlineStruct && structDefName == null) {
             throw new TaskSchemaMismatchError(buildMissingStructDefNameMessage(context, contextName, javaType));
         }
 
-        if (!isInlineStruct && structDefName.isPresent()) {
+        if (!isInlineStruct && structDefName != null) {
             throw new TaskSchemaMismatchError(buildUnexpectedStructDefNameMessage(context, contextName, javaType));
         }
     }
 
-    private static Optional<String> normalize(String value) {
+    private static String normalize(String value) {
         if (value == null || value.isBlank()) {
-            return Optional.empty();
+            return null;
         }
 
-        return Optional.of(value);
+        return value;
     }
 
     private static String buildMissingStructDefNameMessage(


sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/structdefutil/LHClassType.java
eduwercamacaro, line 25: Given that this class is an internal class, it is safe to remove this method as it is not being used anymore.

sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/util/VariableMapping.java
```diff
@@ -1,198 +1,136 @@
...
-        Optional<LHTypeAdapter<?>> maybeAdapter = LHLibUtil.getTypeAdapterForClass(type, typeAdapterRegistry);
-        if (maybeAdapter.isPresent()) {
-            LHTypeAdapter<?> adapter = maybeAdapter.get();
-            if (adapter.getVariableType() != input) {
-                return Optional.of("TaskDef provides " + input + ", but adapter for " + type.getName() + " maps to "
-                        + adapter.getVariableType());
-            }
-            return Optional.empty();
-        }
+    private static TypeDefinition toExpectedTypeDefinition(VariableDef expected, LHTaskParameter lhTaskParameter) {
```
eduwercamacaro: I think we can remove `VariableDef expected` arg. It is not being used.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/structdefutil/LHClassType.java
```diff
@@ -75,6 +83,15 @@ public LHClassType getCoreComponentType(LHTypeAdapterRegistry typeAdapterRegistr
         return LHClassType.fromJavaClass(coreType, typeAdapterRegistry);
     }
 
+    public LHClassType getComponentType(LHTypeAdapterRegistry typeAdapterRegistry) {
```
eduwercamacaro, line 86: Unused method?

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/structdefutil/LHClassType.java
```diff
@@ -35,6 +36,9 @@ public static LHClassType fromJavaClass(Class<?> classType) {
     public static LHClassType fromJavaClass(Class<?> classType, LHTypeAdapterRegistry typeAdapterRegistry) {
         if (classType == null) {
             throw new IllegalArgumentException("Class type should not be null");
+        } else if (void.class.equals(classType)) {
```
eduwercamacaro: Here, we are only testing the `void` primitive. I think we should also test the `Void` wrapper class.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTaskReturnType.java
```diff
@@ -0,0 +1,68 @@
...
+import io.littlehorse.sdk.wfsdk.internal.structdefutil.LHClassType;
+import io.littlehorse.sdk.wfsdk.internal.structdefutil.LHStructDefId;
+import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.Optional;
+import lombok.Getter;
+
+public class LHTaskReturnType {
+
+    @Getter
```
eduwercamacaro: I don't think we need a getter here. It is not being used externally.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTaskReturnType.java
```diff
@@ -0,0 +1,68 @@
...
+        Map<String, String> resolvedPlaceholderValues =
+                placeholderValues == null ? Map.of() : Map.copyOf(placeholderValues);
+
+        LHTypeMetadata metadata = LHTypeMetadata.from(method, resolvedPlaceholderValues);
+        this.isMasked = metadata.isMasked();
+
+        metadata.validateStructDefNameUsage(
+                method.getReturnType(), LHTypeMetadata.ValidationContext.RETURN_TYPE, method.getName());
+
+        this.returnClassType =
```
eduwercamacaro: At this point, we could create the ReturnType instance. right? That would reduce the number of fields on this class.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTaskReturnType.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        returnType = ReturnType.newBuilder()
+                .setReturnType(returnClassType.get().getTypeDefinition().toBuilder()
+                        .setMasked(isMasked)
+                        .build())
+                .build();
+    }
+
+    public ReturnType getReturnType() {
```
eduwercamacaro: This can be even simpler: just a getReturnType that returns the returnType field.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTaskParameter.java
```diff
@@ -0,0 +1,86 @@
...
+        LHTypeMetadata metadata = LHTypeMetadata.from(parameter, resolvedPlaceholderValues);
+        Optional<String> structDefName = metadata.getStructDefName();
+
+        this.variableName = metadata.getName().orElseGet(this::getVarNameFromParameterName);
+        this.isMasked = metadata.isMasked();
+
+        metadata.validateStructDefNameUsage(
+                parameter.getType(), LHTypeMetadata.ValidationContext.PARAMETER, parameter.getName());
+
+        this.variableClassType = buildVariableClassType(typeAdapterRegistry, structDefName);
```
eduwercamacaro: Same comment as [here](https://github.com/littlehorse-enterprises/littlehorse/pull/2169/changes#r3010966771).

I see no reason to not build the `variableClassType` in the constructor.

eduwercamacaro: An instance of this class should be immutable

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTypeMetadata.java
```diff
@@ -0,0 +1,105 @@
...
+    enum ValidationContext {
+        PARAMETER,
+        RETURN_TYPE
+    }
+
+    private final boolean masked;
+    private final Optional<String> name;
+    private final Optional<String> structDefName;
+
+    private LHTypeMetadata(boolean masked, Optional<String> name, Optional<String> structDefName) {
```
eduwercamacaro: Instead of having optional arguments here, I would recommend overloading the constructors with different arguments. 
`private LHTypeMetadata(boolean masked, String name, String structDefName)`
and
`private LHTypeMetadata(boolean masked)`

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTypeMetadata.java
```diff
@@ -0,0 +1,105 @@
...
+    static LHTypeMetadata from(AnnotatedElement element, Map<String, String> placeholderValues) {
+        Map<String, String> placeholders = placeholderValues == null ? Map.of() : placeholderValues;
+
+        if (!element.isAnnotationPresent(LHType.class)) {
+            return new LHTypeMetadata(false, Optional.empty(), Optional.empty());
+        }
+
+        LHType typeAnnotation = element.getAnnotation(LHType.class);
+
+        Optional<String> parsedName = normalize(typeAnnotation.name());
```
eduwercamacaro: Should we also replace placeholders for the type name?

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/structdefutil/LHClassType.java
Snarr: This class is not necessarily internal. It is the easiest way for users to generate `StructDef`s and `PutStructDefRequest`s from a class, and I believe used in an example. But I'd be happy to remove it if we think nobody is using it

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/structdefutil/LHClassType.java
Snarr: It is used in the LH arrays PR, we can remove it for now though

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTaskReturnType.java
Snarr: For sure, I was trying to consider a lazy initialization case early in the design of this code but I realized it was not necessary

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WfRunVariableImpl.java
```diff
@@ -59,44 +58,27 @@ public static WfRunVariableImpl createPrimitiveVar(
...
         WfRunVariableImpl wfRunVar = new WfRunVariableImpl(name, parent);
-        wfRunVar.initializeAsStructDef(structDefName);
-        return wfRunVar;
-    }
 
-    private void initializeAsPrimitive(Object typeOrDefaultVal) {
         if (typeOrDefaultVal instanceof VariableType) {
             VariableType variableType = (VariableType) typeOrDefaultVal;
-            this.typeDef =
+            wfRunVar.typeDef =
```
eduwercamacaro: Instead of just mutating the wfrunVar instance. What do you think about passing the typeDefinition as a constructor arg?

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/structdefutil/LHClassType.java
eduwercamacaro: ~Yep. and make the LH Arrays PR work with the new method instead~

NVM, there is no alternative method for this. We can keep it then.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTypeMetadata.java
Snarr: I am not sure 🤔 I don't think we were replacing placeholders on type names before this PR

PR #2171, opened 2026-03-25
feat(server): Tenant aggregated metrics

l4crito:
### Summary

This PR introduces support for storing and querying **tenant-level aggregated metrics** on the server.

### Details
Previously, metrics were only available at the workflow or task level.

- Previously, each metric punctuator  produced **a single aggregation**, scoped only to a specific `wfSpec` or `taskDef`.
- Now, each punctuator execution produces **two aggregations**:
  1. A **scoped aggregation** for the specific `wfSpec` or `taskDef` (existing)
  2. A **tenant-wide aggregation** across all workflows or tasks (new)

### Example

Given two workflows:

- `example-basic` → 100 executions in the last minute  
- `example-external-event` → 45 executions in the last minute  

You can query metrics as follows:

```bash
lhctl list wfMetrics example-basic
# response:
# started { count: 100 }

lhctl list wfMetrics example-external-event
# response:
# started { count: 45 }
```

With the new aggregation support:
```bash
lhctl list wfMetrics
# response:
# started { count: 145 }
```
The same behavior applies to task metrics:
```bash
lhctl list taskMetrics
```

### DIFF
--- a/server/src/test/java/io/littlehorse/server/streams/topology/core/processors/CommandProcessorTest.java
+++ b/server/src/test/java/io/littlehorse/server/streams/topology/core/processors/CommandProcessorTest.java
@@ -138,9 +138,9 @@ public class CommandProcessorTest {
 
         mockProcessorContext.scheduledPunctuators().get(0).getPunctuator().punctuate(System.currentTimeMillis());
 
-        assertThat(mockProcessorContext.forwarded()).hasSize(1);
+        assertThat(mockProcessorContext.forwarded()).hasSize(2);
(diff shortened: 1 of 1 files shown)


schemas/littlehorse/object_id.proto
```diff
@@ -249,4 +249,6 @@ message MetricWindowId {
   // Start time of the window.
   google.protobuf.Timestamp window_start = 7;
 
+  optional MetricWindowType metric_type = 8;
```
l4crito, line 253: I added this cause it is hard to infer the type when there is no id (in the case of tenant aggregated metrics), this will get stored and passed to the models

PR #2178, opened 2026-03-30
docs(proposals): Update Proposal #000 with latest Arrays implementation

Snarr:
This updates Proposal #000 with the latest LH Arrays implementation details.


proposals/000-struct-and-structdef.md
```diff
@@ -67,51 +68,12 @@ message Struct {
...
-    DoubleArray double_arr = 3;
-    BoolArray bool_arr = 4;
-    IntArray int_arr = 5;
-    BytesArray bytes_arr = 6;
-    WfRunIdArray wf_run_id_arr = 7;
-    StructArray struct_arr = 8;
-    ArrayArray array_arr = 9;
-  }
+  // Set and validated at VariableValue ingress to prevent unnecessary lookups
+  TypeDefinition element_type = 2;
```
eduwercamacaro, line 76: Can an array have null VariableValues?

Snarr: I think so, yes. At least based on this proposal and its current implementation, with the server accepting `null` as a valid value for any TypeDefinition.

PR #2189, opened 2026-04-03
fix(server): pin StructDef version on metadata put requests

Snarr:
- Allows clients to configure StructDef versions locally and set them in `declareStruct` methods
- Adds a bit of logic to the `validateStructDefExist`s method called by metadata requests that pins the StructDef version to the latest version at `Put` time when the version is not specified
- Adds logic to validate incoming Structs against StructDefs using a `superset` validation rule; AKA, only validate whether or not the given Struct has required fields for the StructDef. A Struct with all required fields will not be rejected for having extra fields.
- Refactors some typing exception handling, throws LHApiException only at RequestModel level objects.


server/src/test/java/e2e/PutWfSpecRequest.java
coltmcnealy-lh, line 1: Is this needed?

coltmcnealy-lh (approved):

PR #2193, opened 2026-04-08
feat(sdk-java): Worker task method resolver

sathya-jayapal-dc:
Allows clients (littlehorse workers) to supply a custom worker task method resolver.

### DIFF
--- a/sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTaskSignature.java
+++ b/sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTaskSignature.java
@@ -5,9 +5,8 @@ import io.littlehorse.sdk.common.exception.TaskSchemaMismatchError;
 import io.littlehorse.sdk.common.proto.PutTaskDefRequest;
 import io.littlehorse.sdk.common.proto.ReturnType;
 import io.littlehorse.sdk.wfsdk.internal.structdefutil.LHStructDefType;
-import io.littlehorse.sdk.worker.LHTaskMethod;
+import io.littlehorse.sdk.worker.LHTaskMethodHandle;
 import io.littlehorse.sdk.worker.WorkerContext;
-import io.littlehorse.sdk.worker.internal.util.PlaceholderUtil;
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 import java.util.ArrayList;
@@ -30,13 +29,13 @@ public class LHTaskSignature {
 
     LinkedHashSet<LHStructDefType> structDefClasses;
 
-    Method taskMethod;
+    private final Method taskMethod;
 
     @Getter
-    private String taskDefName;
+    private final String taskDefName;
 
     @Getter
-    private Optional<String> taskDefDescription;
+    private final Optional<String> taskDefDescription;
 
     boolean hasWorkerContext;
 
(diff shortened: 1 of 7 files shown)


eduwercamacaro:
Hi @sathya-jayapal-dc 

Thanks for the PR.

Overall, I really like this proposal, and I understand the problem you are trying to solve here. 

We usually discuss the details of a proposal in a versioned file in the [`/proposals`](https://github.com/littlehorse-enterprises/littlehorse/tree/master/proposals) folder, but I'm ok discussing the details of this proposal using this PR as I don't want to block this. We can create a proposal file for you if that's ok.

I would love to include this in our next minor version, which is 1.1.0. 

I'll make a first pass of review on this PR soon.

sathya-jayapal-dc:
Thanks. Happy to go over it and get through the line.

sdk-java/src/main/java/io/littlehorse/sdk/worker/LHTaskMethodResolver.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Resolves the method to invoke for task execution.
+     *
+     * @param executable the object containing the task implementation
+     * @param taskDefName the resolved task definition name
+     * @param placeholderValues key-value pairs used to resolve placeholders  during method resolution
+     * @return the resolved {@link Method} to invoke
+     * @throws IllegalArgumentException if no matching method is found or if resolution fails
+     */
+    Method resolve(Object executable, String taskDefName, Map<String, String> placeholderValues);
```
eduwercamacaro: I wonder if we should return not only the java Method but also the description and the name of the taskDef. This metadata usually is defined by users at `LHTaskMethod` level and is needed to build an LHTaskSignature. 


I think the point here is to not depend on the `LHTaskMethod` annotation to build an LHTaskSignature instance, and in order to do that, the LHTaskMethodResolver should return this metadata.

It seems that the LHTaskSignature is still depending on reflection to get the LHTaskMethod annotation. see line 56 in LHTaskSignagure class

sathya-jayapal-dc: I have updated the PR with the above changes.

Snarr (approved):

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTaskSignature.java
```diff
@@ -45,20 +44,15 @@ public class LHTaskSignature {
...
+            LHTypeAdapterRegistry typeAdapterRegistry,
+            Map<String, String> placeholderValues) {
+        this.taskMethod = taskMethodHandle.getTaskMethod();
         this.typeAdapterRegistry = Objects.requireNonNull(typeAdapterRegistry);
         this.placeholderValues = placeholderValues == null ? Map.of() : Map.copyOf(placeholderValues);
-        taskDefName = PlaceholderUtil.replacePlaceholders(lhTaskMethod.value(), this.placeholderValues);
-        taskDefDescription = Optional.of(lhTaskMethod.description());
+        this.taskDefName = taskMethodHandle.getTaskDefName();
+        this.taskDefDescription =
+                Optional.ofNullable(taskMethodHandle.getDescription()).filter(desc -> !desc.isEmpty());
```
eduwercamacaro, line 55: We should not add the `filter()` here because that would change the way we register taskdefs's descriptions causing failures during client upgrades. We should continue using the empty string as a default here for compatibility reasons.

eduwercamacaro (commented):
This looks good to me. Just a minor change, and then we can merge it. Thanks!

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/taskdefutil/LHTaskSignature.java
sathya-jayapal-dc: I did this as our workers were failing with
TaskDef already exists and is immutable.

And when i looked at the previous version of LHTaskSignature, there was an empty check.
String annotationDescription = lhTaskMethod.description();
if (annotationDescription != null && !annotationDescription.isEmpty()) {
    taskDefDescription = annotationDescription;
}

eduwercamacaro: Oh! I see it now. 

The actual breaking change was introduced on [this PR](https://github.com/littlehorse-enterprises/littlehorse/pull/2169), and you are fixing it.

I'm glad you caught this. Thank you! ⭐

eduwercamacaro:
@sathya-jayapal-dc Thanks for the PR. I'll merge it when build completes

eduwercamacaro (approved):

PR #2194, opened 2026-04-08
feat(test-utils): Support StructDef registration annotation

Snarr:
This PR adds an annotation to `test-utils` called `@WithStructDefs` that allows test writers to annotate classes with an ordered list of StructDef dependency classes to be registered before all other metadata objects.

## Usage

In the code below you will find an example for registering StructDefs defined by the "Car" class and the "Person" class using this new annotation. In this example, `Car` will be registered before `Person` based on their order.

```java
@LHTest
@WithStructDefs({Car.class, Person.class})
public class MyTestClass {
  // ...
}
```


server/src/test/java/e2e/StructMaskedFieldsTest.java
```diff
@@ -4,35 +4,28 @@
...
 import io.littlehorse.sdk.worker.LHTaskMethod;
 import io.littlehorse.test.LHTest;
 import io.littlehorse.test.LHWorkflow;
+import io.littlehorse.test.WithStructDefs;
 import io.littlehorse.test.WorkflowVerifier;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
 @LHTest
+@WithStructDefs({UserCredentials.class})
```
l4crito, line 21: Does this mean that the Struct will get registered when the test class starts only once?

l4crito (approved):

server/src/test/java/e2e/StructMaskedFieldsTest.java
Snarr: Yes!

eduwercamacaro (approved):
Thanks for improving the e2e framework. :D

PR #2196, opened 2026-04-08
fix(sdk-python): avoid worker crash when signal handlers are unsupported

### LINKED ISSUE #2195: `NotImplementedError` on worker startup due to `asyncio.add_signal_handler` in shutdown hook
### Context

Running a Python worker app with the LittleHorse Python SDK on Windows starts successfully, then crashes during shutdown hook registration. The crash appears to be caused by `asyncio` on Windows not supporting `loop.add_signal_handler(...)`.  

Observed package/runtime details: 
- `littlehorse-client==1.0.2` 
- Python `3.10.20` 
- Windows 11 

```text 
File "...site-packages\littlehorse\worker.py", line 890, in shutdown_hook   
    loop.add_signal_handler(sig, functools.partial(stop_workers, *workers)) 
File "...lib\asyncio\events.py", line 553, in add_signal_handler   
  raise NotImplementedError NotImplementedError 
``` 

### Steps To Reproduce

1. Use Windows 11 with Python 3.10.x.
2. Install `littlehorse-client==1.0.2`.
3. Start one or more workers via::     - `await asyncio.gather(*(littlehorse.start(worker) for worker in workers))` 
4. Observe startup output, then crash with `NotImplementedError`.

### Minimal reproducible snippet
```python
import asyncio
import littlehorse
from littlehorse.worker import LHTaskWorker

async def task_fn(payload: dict) -> dict:
    return {"ok": True}

async def main():
    config = ...  # LHConfig configured for your env
    worker = LHTaskWorker(task_fn, "example-task", config)
    await littlehorse.start(worker)

if __name__ == "__main__":
    asyncio.run(main())
```

### Expected Behavior

Worker startup should succeed on Windows. If POSIX signal handlers are unavailable, the SDK should degrade gracefully (e.g., skip signal registration and use fallback shutdown behavior) rather than crash. 
  

### Affected Components

Python SDK

### Environment

- OS: Windows 11
- Client Version: `littlehorse-client 1.0.2`
- Python: `3.10.20` (Conda env: `validation`)
- Server endpoint: `littlehorse:2023`
- Execution pattern: `asyncio.gather(*(littlehorse.start(worker) for worker in workers))`

### Documentation

Please document Python SDK worker behavior/limitations on Windows regarding signal handling, including expected fallback behavior or recommended setup/workarounds. 

### Additional Context

Suggested fix direction:
- In Python SDK worker shutdown hook, guard `loop.add_signal_handler(...)` with a platform capability check and/or catch `NotImplementedError`.
- Provide fallback shutdown behavior on platforms where signal handlers are unsupported so worker startup remains functional.

jcustenborder:
## Description
Fixes #2195.
This fixes a Windows crash in the Python SDK worker startup path. `shutdown_hook()` previously called `asyncio.add_signal_handler(...)` unconditionally, which raises `NotImplementedError` on Windows event loops and aborted startup.
## Type of change
- Bug fix (non-breaking change which fixes an issue)
- This change requires a documentation update
## Describe the solution
- In `sdk-python/littlehorse/worker.py`, `shutdown_hook()` now:
  - tries `loop.add_signal_handler(...)`,
  - falls back to `signal.signal(...)` on `NotImplementedError`,
  - and skips unsupported registrations (`ValueError`/`OSError`) instead of crashing.
- Added regression tests in `sdk-python/tests/test_worker.py` for supported and fallback paths.
- Added a Windows behavior note in `sdk-python/README.md`.
## Additional context
Validated with rollback/roll-forward test flow and confirmed working on Windows.


sdk-python/littlehorse/worker.py
```diff
@@ -887,7 +887,18 @@ def stop_workers(*workers: LHTaskWorker) -> None:
     loop = asyncio.get_running_loop()
 
     for sig in (signal.SIGTERM, signal.SIGINT):
-        loop.add_signal_handler(sig, functools.partial(stop_workers, *workers))
+        stop_callback = functools.partial(stop_workers, *workers)
+        try:
+            loop.add_signal_handler(sig, stop_callback)
+        except NotImplementedError:
```
sauljabin, line 894: Is it possible to check which SO is running on? and depending on the SO to set the right signal.
Anyways, it looks good 👍🏽

sauljabin: another question, we have multiple github actions (win, linux) wasn't caught there?

coltmcnealy-lh: I don't think the actions catch this because there aren't really any tests in our actions which call `worker.start()` in the python SDK's. That would be an e2e test, we only have e2e in Java unfortunately.

coltmcnealy-lh: Regarding checking which operating system it's on...that's an interesting possibility. There may be more than just linux, mac, windows...I think the `try: catch:` block Jeremy put in here should be sufficient, in the interest of not being as slow-moving as Apache Kafka I'd like to merge this, see if it causes a problem, and react later.

coltmcnealy-lh (approved):

PR #2211, opened 2026-04-16
fix(server): pin StructDef versions on metadata put requests

Snarr:
- Pins StructDef version on metadata put requests (PutWfSpecRequest, PutTaskDefRequest, PutExternalEventDefRequest, etc...)
- When receiving version `-1`, uses latest StructDef version. When receiving any other version number, uses that StructDef version.
- Note: this breaks old clients using new servers, as those old clients will send version `0` for StructDefs instead of `-1`


coltmcnealy-lh (approved):
Structs aren't yet graduated, so this is okay to break.

PR #2215, opened 2026-04-17
feat(server): add `is_nullable` flag for StructFieldDefs

Snarr:
See changes to Proposal #000 for how we will mark some StructFieldDefs as `nullable` and how this will interoperate with default values.


dashboard/src/app/utils/timestamp.ts
```diff
@@ -17,7 +17,15 @@ export function utcParts(d: Date): UtcParts {
   }
 }
 
-export function utcDateFromParts(y: number, mo: number, day: number, h: number, mi: number, s: number, ms: number): Date {
```
Snarr, line 20: Changed caused by running prettier write formatter.

proposals/000-struct-and-structdef.md
```diff
@@ -346,6 +353,39 @@ service LittleHorse {
...
+
+### Design Decision
+
+By default, every `StructField` in a `StructDef` is **non-nullable**: a field's value may not be `null` (i.e. `VALUE_NOT_SET` in the `VariableValue` oneof). This mirrors LittleHorse's recent philosophy of strong-typing and rejecting ambiguity at compile time — the same reason we introduced strong typing for `Struct`s in the first place.
+
+However, there are legitimate real-world use cases where a field value must be allowed to be null (e.g. an optional middle name, an unset expiry date, or a nullable foreign-key reference) and where you don't want a default value either (eg. Strings -> `""`, Integers --> `0`). To support these without compromising the default safety guarantees, we add an `is_nullable` flag to `StructFieldDef`.
+
+### Semantics
+
+**Nullability and requiredness are orthogonal.** A field being nullable controls whether its _value_ can be `null`; a field being required controls whether it must be _present_ in the struct payload. The two axes combine as follows:
```
coltmcnealy-lh: I wouldn't use the word "required" or "not required." I'd use "has default" and "has no default."

With `FULL_COMPATIBILITY` updates, you can only add fields that are either nullable or have a default value.

Ideally, when we add the `BACKWARD_COMPATIBILITY` update type, we will relax that constraint and add the ability to add fields that are non-nullable and do not have a default.

schemas/littlehorse/common_wfspec.proto
```diff
@@ -334,8 +334,11 @@ message StructFieldDef {
   TypeDefinition field_type = 1;
   
   // The default value of the field, which should match the Field Type. If not
-   // provided, then the field is treated as required.
+  // provided, then the field is treated as required.
```
coltmcnealy-lh, line 337: To me, whether a field is "required" or not is determined by the `is_nullable` field. IF we weren't making a (somewhat unnecessary) attempt at backwards compatibiltiy, we would have named field `3` as `required` (but the default woudl be that all previously-created `StructDef`s, or alll `StructDef`s created with old clients, end up having all optional fields.

sdk-java/src/test/java/io/littlehorse/sdk/common/LHLibUtilTest.java
```diff
@@ -410,4 +410,96 @@ public void setName(String name) {
             this.name = name;
         }
     }
+
+    @LHStructDef("nullable-field-struct")
+    public static class NullableFieldStruct {
+        private String required;
+        private String optional;
```
coltmcnealy-lh: How do we tell the difference between something being required or not here?

Snarr: For the purposes of these tests it doesn't matter if this is marked with a default value or as non-nullable, that information isn't used at runtime when ingesting values from the server. All of the validation happens on the server and the client just takes the Struct it receives and converts it to a Java object.

coltmcnealy-lh: I see

coltmcnealy-lh (approved):

PR #2252, opened 2026-05-06
fix(dashboard): honor WfSpec defaults in run workflow form

### LINKED ISSUE #2205: Dashboard overwrites WfRun variable defaults
### Context

1. In `WfSpec`s we can define `VariableDef`s with default values
2. On the dashboard, the `Execute WfRun` form takes in inputs for `WfRun` variables
3. If we don't supply any value to a form input and we execute the `WfRun`, the server should use the default value specified in the `WfSpec`'s `VariableDef`

### Steps To Reproduce

1. Register a `WfSpec` with a `STR` `VariableDef` with a default value. (ex: "Hello")

```java
public static Workflow getWorkflow() {
    return new WorkflowImpl("example-default-bug", wf -> {
        WfRunVariable bookName = wf.declareStr("book-name").withDefault("Harry Potter and the Philosopher's Stone");
    });
}
```

2. Run the dashboard.
3. Navigate to the given `WfSpec` page and go to the `Execute WfRun` form.
4. Without inputing anything into the dashboard form, execute the `WfRun`.
5. Notice that in the `WfRun` `Variable`s section at the bottom of the `WfRun` page, the `Variable` value is blank

### Expected Behavior

The expected behavior is that when a user does not provide an input to a `WfRun` form, the default value should be used by the server.

We should also be able to distinguish between providing no value at all and providing an empty String, since those are two different use cases. An empty String _should_ overwrite the default, while providing no value at all should not.

### Affected Components

Dashboard

### Environment

Client Version: Dashboard 1.1.0-SNAPSHOT
Server Version: 1.1.0-SNAPSHOT
Browser: Safari Version 18.6 (20621.3.11.11.3)

### Documentation

_No response_

### Additional Context

_No response_

bryson-g:
resolves #2205

Blank fields in the run-workflow UI were overwriting variable defaults from the spec. Values from the form get merged with WfSpec defaults before the run starts. Helper is in variables.ts with tests.

### Video

https://github.com/user-attachments/assets/fb41f6fe-2755-446e-a93d-84b72754292c

### DIFF
--- a/dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Forms/WfRunForm.tsx
+++ b/dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Forms/WfRunForm.tsx
@@ -1,5 +1,6 @@
+import { getPrimitiveFormDefaultValue } from '@/app/utils'
 import { Input } from '@/components/ui/input'
-import { ThreadVarDef, VariableType, WfSpec } from 'littlehorse-client/proto'
+import { ThreadVarDef, VariableType, WfRunVariableAccessLevel, WfSpec } from 'littlehorse-client/proto'
 import { forwardRef, useMemo } from 'react'
 import { FormProvider, useForm } from 'react-hook-form'
 import FormField from './components/FormField'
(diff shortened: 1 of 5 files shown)


Snarr (approved):

Snarr:
Looks great, thanks!

PR #2258, opened 2026-05-08
fix(sdk-java): forbid JSON types in StructDefs and InlineArrayDefs

Snarr:
Upon re-evaluating the original intentions of Proposal #000, we came to realize that JSON types do not have a place in `StructDef`s or `InlineArrayDef`s, as they are superseded by `Struct`s and `Array`s and only serve to worsen type-safety and increase complexity within these structures.

New functionality:

- When a `StructDef` field type resolves to `JSON_OBJ` or `JSON_ARR`, throws a descriptive exception.
- When an `InlineArrayDef` type resolves to `JSON_OBJ` or `JSON_ARR`, throws a descriptive exception.
- Array fields in `StructDef`s default to `InlineArrayDef`s
- Java class fields in `StructDef`s where classes use the `@LHStructDef` annotation are interpreted as `StructDef`s still
- Types that adapt to `JSON_OBJ` or `JSON_ARR` but are used in forbidden places (in `StructDef`s or `InlineArrayDef`s) will also be forbidden.


sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/structdefutil/ForbiddenJsonTypeException.java
```diff
@@ -0,0 +1,17 @@
+package io.littlehorse.sdk.wfsdk.internal.structdefutil;
+
+import io.littlehorse.sdk.common.proto.VariableType;
+
+final class ForbiddenJsonTypeException extends Exception {
```
eduwercamacaro, line 5: I like the idea of having an exception for forbidden json types.

Can we make this exception a RuntimeException? Instead of extending from `Exception` I think we can extends from `RuntimeException` because at the end, callers of this exception are wrapping this exception into IllegalArgumentException.


Or we can extend from IllegalArgumentException. wdyt?

Snarr: The reason I made it extend from `Exception` is so that callers have to catch it and wrap it with their own context. For example, `LHArrayType` has a different context than `LHStructProperty` when this error is thrown. I just pushed some refactors so that the shared parts of the message are refactored into ForbiddenJsonTypeException, and the two classes only add the necessary context.

eduwercamacaro: Ohh I see.
ok that makes sense. Thanks for the explanation.

eduwercamacaro (approved):

PR #2261, opened 2026-05-11
feat(sdk-go): forbid JSON types in StructDefs

Snarr:
This PR ports #2258 to `sdk-go`

### DIFF
--- a/sdk-go/littlehorse/lh_struct_def.go
+++ b/sdk-go/littlehorse/lh_struct_def.go
@@ -96,8 +96,8 @@ func resolveFieldName(lhName string, field reflect.StructField) string {
 //   - float32, float64 -> DOUBLE
 //   - bool -> BOOL
 //   - []byte -> BYTES
-//   - slices/arrays -> JSON_ARR
 //   - structs with LHStructDef() -> nested StructDef reference
+//   - slices and arrays of the above -> coming soon: LH native arrays
 func GoStructToInlineStructDef(structInstance interface{}) (*lhproto.InlineStructDef, error) {
 	t := reflect.TypeOf(structInstance)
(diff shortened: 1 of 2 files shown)


sdk-go/littlehorse/lh_struct_def.go
```diff
@@ -191,7 +196,7 @@ func goTypeToStructFieldDef(t reflect.Type) (*lhproto.StructFieldDef, error) {
 	}, nil
 }
 
-func goKindToVariableType(t reflect.Type, isPtr bool) (lhproto.VariableType, error) {
+func goKindToVariableType(t reflect.Type) (lhproto.VariableType, error) {
```
eduwercamacaro: Is this function supposed to resolve json objects? 

I believe yes, we should continue supporting JSON objects. right?

eduwercamacaro: My understanding is that this function not only resolves Struct and inner structs but also json object types.

Snarr: This method is only used in this `lh_struct_def.go` file -- the name could be improved, but it is not used in other places where we do resolve JSON.

eduwercamacaro: I see, it is private. Thanks.!

eduwercamacaro: What's the purpose of adding an if case that handles JSON_OBJ in [the line 183](https://github.com/littlehorse-enterprises/littlehorse/pull/2261/changes#diff-48fffc20c85f0916aabe02d088453eadeb7610b64bd01dad7c6d13b30d5eab87R183)?

eduwercamacaro (approved):

PR #2265, opened 2026-05-12
feat(server): Counted tags

eduwercamacaro:
implements proposal 021-counted-tags.md
### Key Components

| Component | Responsibility |
|-----------|---------------|
| `TagStorageManager` | Detects `COUNTED` tags and increments/decrements the `PartitionCountedTagModel` in memory and local store |
| `PartitionCountedTagModel` | Per-partition accumulator stored in the cluster-scoped store. Holds `tenantId`, `attributeString`, and a running `count` delta |
| `PartitionMetricsMemoryStore` | In-memory map that batches counted tag deltas between punctuations |
| `CommandProcessor` punctuator | Periodically drains the memory store and sends `UpdateCountedTag` commands to the core topic |
| `UpdateCountedTagModel` | A `CoreSubCommand` that applies the delta to the global `CountedTagModel` on the target partition |
| `CountedTagModel` | The authoritative counter stored on the deterministic partition, read at query time |
| `CountRequest<T>` | Abstract base class for count request models; computes the attribute string and routes to the correct partition |


server/src/main/java/io/littlehorse/server/streams/lhinternalscan/count/CountNodeRunRequestModel.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    protected GetableClassEnum getObjectType() {
+        return GetableClassEnum.NODE_RUN;
+    }
+
+    @Override
+    protected List<Attribute> countAttributes() {
+        if (wfSpecName == null) {
+            throw new LHApiException(Status.INVALID_ARGUMENT, "wfSpecName is required");
```
coltmcnealy-lh, line 50: Can we add a new counted tag for all wfspec's in a Tenant? The reason for this is so that in the future, Cluster Administrators can keep track of usage by tenant.

server/src/main/java/io/littlehorse/server/streams/storeinternals/index/CachedTag.java
```diff
@@ -15,31 +15,36 @@ public class CachedTag extends LHSerializable<TagsCachePb.CachedTagPb> {
 
     private String id;
     private boolean isRemote;
+    private boolean isCounted;
+    private String attributeString;
```
coltmcnealy-lh: Naive question. What's the difference between `id` and `attributeString`? Are we storing duplicated info here?

server/src/main/java/io/littlehorse/server/streams/storeinternals/TagStorageManager.java
```diff
@@ -57,51 +53,22 @@ public void store(Collection<Tag> newTags, TagsCache preExistingTags) {
...
+        if (cachedTag.isCounted()) {
+            PartitionCountedTagModel currentAggregation =
+                    partitionMetrics.decrementCounted(authContext.tenantId(), cachedTag.getAttributeString());
+            clusterScopedStore.put(currentAggregation);
         } else {
             lhStore.delete(cachedTag.getId(), StoreableType.TAG);
         }
     }
-
-    private String extractAttributeStringFromStoreKey(String tagStoreKey) {
```
coltmcnealy-lh, line 76: Previous comment suggests that `id` contains the attribute string. I think we may be storing duplicate info now.

server/src/main/java/io/littlehorse/common/model/getable/core/noderun/NodeRunModel.java
```diff
@@ -183,15 +183,33 @@ public Date getCreatedAt() {
 
     @Override
     public List<GetableIndex<? extends AbstractGetable<?>>> getIndexConfigurations() {
+        GetableIndex<? extends AbstractGetable<?>> specNameIndex = new GetableIndex<>(
+                List.of(Pair.of("wfSpecName", GetableIndex.ValueType.SINGLE)), TagStorageType.COUNTED);
+        GetableIndex<? extends AbstractGetable<?>> specNameAndMajorVersionIndex = new GetableIndex<>(
+                List.of(
+                        Pair.of("wfSpecName", GetableIndex.ValueType.SINGLE),
```
coltmcnealy-lh, line 190: I see why we have this for completeness; I'm unsure whether users will need to count noderuns by multiple `WFSpec`s with the same major version but not revision. If we want to get rid of this we can.

coltmcnealy-lh (approved):
Looks good, the biggest concern I have is that we're now duplicating the attribute string in the `Tag` (both in the `id` and now in the new field you added). Yes, compression will help with storage, but it will cost CPU when reading and deserializing it.

The second biggest concern I have is eventually cleaning up the "Count" objects when the `TaskDef` and the `WfSpec` are deleted. That is not critical, so it can become a follow up ticket.

server/src/main/java/io/littlehorse/server/streams/storeinternals/index/CachedTag.java
eduwercamacaro: The `id` contains the full key of a tag, which includes the `createdAt` field. On the other hand, there is the attribute string which is basically a string separated by underscore that contains every attribute of a tag for example (`status__RUNNING_wfSpecName__test`)

The `id` is being used by the partitionKey method

server/src/main/java/io/littlehorse/server/streams/storeinternals/TagStorageManager.java
eduwercamacaro: Oh I see it now.

Thanks for this.

I just removed the attributeString field from the CachedTag storable and added this method so we can extract the attribute string from the partitionKey. 

see [this commit](https://github.com/littlehorse-enterprises/littlehorse/pull/2265/changes/10ffb0e17fe936cda0497dff780d913fceb9e3eb)

server/src/main/java/io/littlehorse/common/model/getable/core/noderun/NodeRunModel.java
eduwercamacaro: The tag `wfSpecName`, `majorVersion` enables user to count the number of noderuns for the *same* wfspec name and version. 

Since revisions can change quickly, I do believe that having this tag is important for both completeness and a use case where users want to monitor the count for a particular major version. Users may want to track only the major versions of a wfspec because, in certain situations, the major version may be more static than revisions.

server/src/main/java/io/littlehorse/server/streams/lhinternalscan/count/CountNodeRunRequestModel.java
eduwercamacaro: Ok, I can do this in a follow-up PR/ticket as this requires modifications in the proposal.

PR #2290, opened 2026-05-20
docs(proposal): Add proposal for counted tags

eduwercamacaro:


l4crito (approved):

proposals/021-counted-tags.md
```diff
@@ -0,0 +1,158 @@
...
+
+```java
+return List.of(new IndexedField(key, wfSpecId.getName(), TagStorageType.COUNTED));
+```
+
+### Consistency Model
+
+Counted tags provide **eventual consistency**:
+
+* Counts are accumulated locally and forwarded periodically (on each punctuation cycle).
```
l4crito, line 120: since we don't have windows here, my only question is how long is it gonna be locally stored ? before it ends in the proper partition

eduwercamacaro: We will clear the local count after forwarding the count to the repartition topic.

proposals/021-counted-tags.md
```diff
@@ -0,0 +1,158 @@
...
+  // Counts the number of TaskRun's currently in TASK_SCHEDULED state for a given TaskDef.
+  // Useful for monitoring task queue depth and detecting backpressure on workers. This is
+  // an eventually consistent count maintained via pre-aggregated counters.
+  rpc CountScheduledTaskRun(CountScheduledTaskRunRequest) returns (Count) {}
+
+// Request to count NodeRun's matching specified criteria. All fields are optional filters
+// that progressively narrow the count. Request will be rejected if no fields are set
+message CountNodeRunRequest {
+
+  // Filter by WfSpec name. If set, only NodeRun's belonging to this WfSpec are counted.
```
coltmcnealy-lh, line 42: So if all are null, then it counts by `Tenant`?

coltmcnealy-lh (approved):

proposals/021-counted-tags.md
eduwercamacaro: No, the proposed indexes for NodeRun does not consider that case.
I put this into the docs:
`Request will be rejected if no fields are set`

Issue #2285, opened 2026-05-19
Possible `MetadataCache` Corruption


coltmcnealy-lh:
It's possible that the `MetadataCache` might be corrupted by `rpc GetTaskDef` running before the Global Store is fully restored. This would put a `null` in the cache, causing intermittent (incorrect) `NOT_FOUND` errors when trying to fetch a `TaskDef`, or causing workflows to fail with "couldn't find TaskDef"

eduwercamacaro:
This could be caused by a check-then-act race condition.

MetadataCache is a singleton object created during instance initialization. 

This behavior could be caused by a race condition that appears when both grpc thread and the global processor thread try to update the metadataCache at the same time for the same key.

eduwercamacaro:
Proposed this change #2291

PR #2291, opened 2026-05-20
fix(server): Solves race condition when updating metadata objects

eduwercamacaro:
Solves #2285


coltmcnealy-lh (approved):
We were able to replicate this issue with a few carefully-placed "sleep()"s and confirm it's solved by this PR. Great work!

PR #2295, opened 2026-05-20
feat(sdk-java): add support for Records to define StructDefs

Snarr:


eduwercamacaro (commented):
I noticed that there are many missing test cases that should verify key behaviors when converting java records into struct defs. I believe the cleanest solution for this would be parameterized tests that run equivalent assertions against both a Java POJO class and a record. This would reduce duplication while explicitly enforcing behavioral parity. In my recent commit, I propose that test. Feel free to either revert the commit or accept it. :)

eduwercamacaro (approved):

PR #2304, opened 2026-05-26
feat(server): Count all the NodeRun in a tenant

eduwercamacaro:
This PR extends the counted-tag proposal to include the ability to count the number of NodeRun on a single tenant.


schemas/littlehorse/service.proto
```diff
@@ -1775,18 +1775,28 @@ message LittleHorseVersion {
...
 message CountNodeRunRequest {
 
-  // Filter by WfSpec name. If set, only NodeRun's belonging to this WfSpec are counted.
-  string wf_spec_name = 1;
+  oneof filter {
+    // Filter by WfSpec attributes.
+    WfSpecFilter wf_spec_filter = 1;
+
+    // Count all NodeRun's in the tenant.
+    google.protobuf.Empty no_filter = 2;
```
Snarr: Could we eliminate the `no_filter` field in this oneof and use the `unspecified`/`unset` filter case to mean "return all in tenant"?

eduwercamacaro: IMO, this makes it more explicit, but I see what you are saying.

What about having both? We count all the tenants if either the field no_filter is set or there is no filter on the request.

Snarr: Assuming this change, maybe `lhctl count nodeRun` without args could return the total count for the tenant. Less explicit, quicker to type, so it has its tradeoffs. Just throwing out ideas.

Snarr: I like both too, just thinking about how we can leverage the `UNSPECIFIED` case more in protobuf because it's easy to forget its there in `oneof`s and easy to forget to include one for `enum`s

eduwercamacaro: Ok, I like it. I just removed the no_filter field on this commit https://github.com/littlehorse-enterprises/littlehorse/pull/2304/changes/42d46303095f7b68820a21f94762d1e5224ada12


thanks :)

Snarr (approved):
Looks good 😁

PR #2333, opened 2026-06-02
feat(sdk-js): minimal workflow WfSpec builder (proposal 018 subset)

GhostArrays:
## Summary

Minimal **TypeScript WfSpec builder**: `Workflow.create` → `compile()` → `PutWfSpecRequest`. Mirrors Java wfsdk internals with a small subset of `WorkflowThread`.

### What's included
- `declareStr` / `declareInt` / `declareDouble` / `declareBool` / `declareJsonObj` / `declareStruct` / `declareArray`
- `execute` (task nodes), `assign`, `mutate`, `format`, `complete`
- `LHExpression` with `add` / `subtract` / `multiply` / `divide` / `pow`
- `TaskNodeOutput` with `withRetries` / `withTimeout` / `withExponentialBackoff`
- `buildStruct` / `buildInlineStruct` for Struct values
- Full `struct.ts` worker-side StructDef definition API

Ref: [proposal 018](https://github.com/littlehorse-enterprises/littlehorse/blob/master/proposals/018-typescript-wfspec-minimal-api.md).


examples/js/expressions/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+```sh
+# Here, we specify that the "quantity" = 1, "price" = 0.8 and "taxes" = 12
+lhctl run example-expressions quantity 1 price 0.8 taxes 12
+```
+
+Or:
+
+```sh
+npm run run-wf -- 1 0.8 12
```
Snarr: I know a lot of the projects in this repo have switched to `pnpm` over `npm`. Should we update this project to use `pnpm` commands as well?

GhostArrays: I mean, I think that makes sense for like actual dependencies and installing, but there's being an open source project and npm is the most widely used package manager so it's the default that everyone has installed so no one has to worry about like doing that stuff but I don't mind changing it as well

examples/js/basic/src/index.ts
```diff
@@ -14,23 +16,23 @@ async function greet(name: string, ctx: WorkerContext): Promise<string> {
 }
 
 async function main() {
-  // Connect to the LH Server (defaults to localhost:2023)
   const config = LHConfig.from({})
+  const client = config.getClient()
 
   const worker = createTaskWorker(greet, 'greet', config, {
     inputVars: { name: z.string() },
```
Snarr: I'm wondering if we can find an alternative design for this, so that the `inputVars` definition is coupled more closely with the Task Method definition.

examples/js/expressions/README.md
steinsleger: We switched to `pnpm` to enforce a minimum release age for dependencies, but it brings other advantages as well. Unlike `pnpm`, `npm` doesn't allow us to exempt specific packages from this rule. This limitation made it impossible to immediately update our own internal dependencies -like the SDK itself or the UI library- following a patch or hotfix. While I would prefer to migrate everything to `pnpm` for consistency, running these existing commands (or others related to registry lookup) remains entirely harmless.

GhostArrays: Okay will do

GhostArrays: Done.

examples/js/basic/src/index.ts
```diff
@@ -15,13 +15,15 @@ async function greet(name: string, ctx: WorkerContext): Promise<string> {
   return msg
 }
 
+const greetTask = defineTask(greet, {
```
Snarr, line 18: Nice this is better

examples/js/basic/src/index.ts
```diff
@@ -1,36 +1,40 @@
...
+
 async function greet(name: string, ctx: WorkerContext): Promise<string> {
   const msg = `Hello, ${name}! (WfRun ${ctx.getWfRunId()?.id})`
   ctx.log(msg)
   console.log(msg)
   return msg
 }
 
+const greetTask = defineTask(greet, {
+  inputVars: { name: z.string() },
```
Snarr, line 19: How/where do we define the return types of the task method?

GhostArrays: fixed

coltmcnealy-lh:
Closing in favor of #2500

PR #2337, opened 2026-06-03
Refactor partition accumulator

eduwercamacaro:
Code refactor for the mechanism of pre-aggregating metrics and counted tags on a partition basis and then forwarding to a repartition store.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streams/topology/core/processors/PartitionDrainScheduler.java
+++ b/server/src/main/java/io/littlehorse/server/streams/topology/core/processors/PartitionDrainScheduler.java
@@ -5,6 +5,7 @@ import static com.google.protobuf.util.Timestamps.toMillis;
 
 import io.littlehorse.common.LHConstants;
 import io.littlehorse.common.LHServerConfig;
+import io.littlehorse.common.Storeable;
 import io.littlehorse.common.model.LHTimer;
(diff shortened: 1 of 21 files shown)


l4crito (approved):

server/src/main/java/io/littlehorse/server/streams/topology/core/processors/PartitionDrainScheduler.java
```diff
@@ -0,0 +1,211 @@
...
+                        return lastSeenWindowTime;
+                    }
+                }
+            }
+        }
+        return lastSeenWindowTime;
+    }
+
+    private void catchUpCountedTags(ClusterScopedStore store) {
+        String prefix = Storeable.getSubstorePrefix(StoreableType.PARTITION_COUNTED_TAG);
```
l4crito, line 149: Dont we need some sort of hint here if for some reason it takes longer to recover when the server fails ? could this lead to a transaction timeout?

l4crito (approved):

server/src/main/java/io/littlehorse/server/streams/topology/core/processors/PartitionDrainScheduler.java
eduwercamacaro: Yes, good observation. I can add it in a follow-up PR to make review easier.

PR #2338, opened 2026-06-04
docs(proposal): workflow migrations

JakeRose03:


proposals/022-workflow-migrations.md
```diff
@@ -0,0 +1,240 @@
...
+### RPC
+```proto 
+  // Register a workflow migration plan with lh server
+  rpc PutWorkflowMigrationPlan(PutWorkflowMigrationPlanRequest) returns (WorkflowMigrationPlan) {}
+
+  // Get a workflow migration plan by ID
+  rpc GetWorkflowMigrationPlan(WorkflowMigrationPlanId) returns (WorkflowMigrationPlan) {}
+
+  // Deletes Workflow Migration Plan Metadata object from the server
+  // Maybe this is not a good idea since wfRun hold migrationPlanId
```
Snarr: Good point and something to consider -- Can a WorkflowMigrationPlan be deleted while a WfRun references the plan? Should we prevent the deletion of a plan while a WfRun refers to it?

proposals/022-workflow-migrations.md
```diff
@@ -0,0 +1,240 @@
...
+### MigrationNodes
+- You have the ability to migrate from and to any node type unless the worklowMigrationPlan
+is applied and a threadRun is currently at a migration node. If the migration node is not a waiting node then the applyMigrationRequest will be rejected
+
+- In the case of which we migrate from an activeNode then the nodeRun
+is put into a halted status when migrated from, but maybe this could be changed to a MIGRATED status.
+
+### Migration Lifecycle
+    - MigrationPlanId will stay stamped on a wfRun
+    till each threadMigration has been satisfied
```
Snarr: Does this imply that MigrationPlanId will be deleted/set to null from a WfRun after a migration is completed? Should we consider keeping the MigrationPlanId to keep an audit trail?

proposals/022-workflow-migrations.md
```diff
@@ -0,0 +1,240 @@
...
+}
+```
+
+
+## Implemention details
+
+
+### Thread Migration Rules 
+- Entrypoint threads can only migrate to entrypoint threads
+- Child threads, interupt, and failure handlers can migrate between
```
Snarr: Does this mean that you can migrate from a Child Thread to an Interrupt?

proposals/022-workflow-migrations.md
JakeRose03: Yes after Migration is completed the migrationPlanId will be deleted, but if the entrypoint thread migrates the wfRuns wfSpec will be updated to the new wfSpec . The now old wfSpec will be added to the current field oldwfSpecVersions. This scenario does not cover the case if the migration plan only migrated a child thread.

I think we should maybe keep migrationPlanId in a list that stores all of the previously applied migrations.
That way if a wfRun migrated more that once we still have an audit trail.

proposals/022-workflow-migrations.md
JakeRose03: Yes, this is because the `threadSpec` does not store information about its type, so we can not infer if a threadSpec we are migrating to is the same type as the current `threadRun`.

eduwercamacaro (approved):

PR #2345, opened 2026-06-12
fix(server): Set tenant id field for repartition timers when creating counted tags

eduwercamacaro:
Add tenant id field to the repartition timer so we ensure that the repartition command lands on the correct tenant. Also, added some unit tests to lock this behavior.

### DIFF
--- a/server/src/main/java/io/littlehorse/server/streams/stores/PartitionMetricsMemoryStore.java
+++ b/server/src/main/java/io/littlehorse/server/streams/stores/PartitionMetricsMemoryStore.java
@@ -2,9 +2,7 @@ package io.littlehorse.server.streams.stores;
 
 import io.littlehorse.common.model.PartitionCountedTagModel;
 import io.littlehorse.common.model.PartitionMetricWindowModel;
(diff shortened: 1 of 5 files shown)


server/src/main/java/io/littlehorse/server/streams/stores/PartitionMetricsMemoryStore.java
```diff
@@ -19,20 +17,6 @@ public void put(PartitionMetricWindowModel windowMetric) {
         metrics.put(windowMetric.getStoreKey(), windowMetric);
     }
 
-    public PartitionCountedTagModel incrementCounted(TenantIdModel tenantId, String tagAttributes) {
```
Snarr, line 22: I guess these were not being used?

Snarr (approved):

server/src/main/java/io/littlehorse/server/streams/stores/PartitionMetricsMemoryStore.java
eduwercamacaro: Yes, not being used. :)

PR #2358, opened 2026-06-18
feat(sdk-java): Allow custom gRPC ClientInterceptors in LHConfig

eduwercamacaro:
Adds support for providing gRPC `ClientInterceptor`s to the Java SDK's `LHConfig`. Every gRPC channel created by the SDK (blocking, future, and async stubs — used both for metadata registration and by the task worker) is now instrumented with these interceptors.

Interceptors can be provided two ways:

- **Varargs constructors** — every `LHConfig` constructor now has an overload that accepts `ClientInterceptor...`:
  - `LHConfig(Properties, ClientInterceptor...)`
  - `LHConfig(Path, ClientInterceptor...)`
  - `LHConfig(String, ClientInterceptor...)`
  - `LHConfig(Map<String, Object>, ClientInterceptor...)`
- **Builder** — `LHConfig.newBuilder().addClientInterceptor(interceptor)` for fluent composition.

Existing constructors are preserved and simply delegate with an empty interceptor array, so this is fully backwards compatible. Null interceptors are filtered out.

## Motivation

The main motivation is to enable **metrics instrumentation** via gRPC client interceptors (e.g. Micrometer's `MetricCollectingClientInterceptor` for Prometheus). More generally, this gives users an extension point to plug in **any** interceptor for their use case — metrics, custom logging, etc. — without having to fork the SDK or build their own channels.

## Interceptor ordering

Order is `[RESOURCE_EXHAUSTED_RETRY_INTERCEPTOR, ...userInterceptors]`. The first entry is the outermost (first to see the outbound call), so retries wrap user interceptors. Per-attempt metrics/tracing are captured as a result.

## Backwards compatibility

- All existing constructors continue to work unchanged.
- When no interceptors are supplied, behavior is identical to before.

## Notes for reviewers

- Should the retry interceptor be outermost or innermost relative to user interceptors? Current choice (outermost) favors per-attempt observability; flagging in case we'd rather count one logical call.

### DIFF
--- a/examples/java/basic/src/main/java/io/littlehorse/examples/BasicExample.java
+++ b/examples/java/basic/src/main/java/io/littlehorse/examples/BasicExample.java
@@ -5,9 +5,6 @@ import io.littlehorse.sdk.wfsdk.WfRunVariable;
 import io.littlehorse.sdk.wfsdk.Workflow;
 import io.littlehorse.sdk.wfsdk.internal.WorkflowImpl;
 import io.littlehorse.sdk.worker.LHTaskWorker;
-import io.micrometer.core.instrument.binder.grpc.MetricCollectingClientInterceptor;
-import io.micrometer.prometheusmetrics.PrometheusConfig;
-import io.micrometer.prometheusmetrics.PrometheusMeterRegistry;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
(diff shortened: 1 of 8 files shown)


Snarr (approved):
Great addition

Snarr:
Can you help me understand this comment more:

> Should the retry interceptor be outermost or innermost relative to user interceptors? Current choice (outermost) favors per-attempt observability; flagging in case we'd rather count one logical call.

What does outermost vs innermost mean? The first or last interceptor in the chain of interceptors? Can you help me put this into that context

PR #2359, opened 2026-06-18
Add write buffer metrics

eduwercamacaro:


mijailr (approved):

server/src/main/java/io/littlehorse/server/monitoring/StandbyMetrics.java
```diff
@@ -48,6 +51,16 @@ void registerMetrics() {
                         .tag(STORE_NAME, storeName)
                         .tag(INSTANCE_ID, instanceId)
                         .register(registry));
+        if (writeBufferManagerSize > 0) {
```
coltmcnealy-lh, line 54: Question—why is this in the standby metrics? Does it not apply to active tasks? I'm asking because the `WriteBufferManager` belongs to all of the rocksdb instances (for both active and standby's).

coltmcnealy-lh (approved):

PR #2360, opened 2026-06-19
fix(server): throw NodeRun failure if try to spawn too many threads

### LINKED ISSUE #2357: `spawnThreadForEach` exceeding `X_MAX_THREAD_RUNS_PER_WF_RUN` leaves orphan child threads in non-terminal state, causing WfRun to appear frozen
### Context

When a `WfSpec` uses `spawnThreadForEach` on an array that results in total thread count exceeding `LHS_X_MAX_THREAD_RUNS_PER_WF_RUN` (default: 65), the server spawns all child threads first via `StartMultipleThreadsRunModel.arrive()`, then detects the overshoot on the next iteration of the `WfRunModel.advance()` while-loop. It force-transitions the WfRun to `ERROR` and places a failure on thread 0's NodeRun, but the already-spawned child threads are never halted or terminated — they remain in `RUNNING` status permanently with their NodeRuns in `STARTING` state.

The WfRun reaches an inconsistent terminal state: it is `ERROR` but contains child threads that are `RUNNING`. Since `advance()` is never called again on a terminated WfRun, those child threads can never progress or be cleaned up. From an operator's perspective monitoring thread or task status (rather than the top-level WfRun status), the workflow appears frozen.

### Steps To Reproduce

1. Deploy a `WfSpec` that calls a task returning a JSON array, then passes that array to `spawnThreadForEach`:
   ```java
   WfRunVariable items = wf.declareJsonArr("items");
   items.assign(wf.execute("get-items"));
   SpawnedThreads threads = wf.spawnThreadForEach(items, "process-item", this::processItem);
   wf.waitForThreads(threads);
   ```
2. Have the `get-items` task return an array with 65 or more elements (i.e., enough to exceed the default limit of 65 total thread runs when the entrypoint thread is included: 65 items + 1 entrypoint = 66 > 65).
3. Run the workflow.
4. **Observe:** The `get-items` task completes successfully. The `START_MULTIPLE_THREADS` node's `arrive()` spawns all 65 child threads. On the next iteration of the `advance()` while-loop, the size check fires. The WfRun transitions to `ERROR` with the message "You exceeded the maximum number of ThreadRuns per WfRun: 65" on thread 0's NodeRun. However, threads 1–65 remain in `RUNNING` status with their entrypoint NodeRuns in `STARTING` — permanently.

### Expected Behavior

One of:

- **Preferred:** The limit should be enforced *before* spawning threads — inside `StartMultipleThreadsRunModel.arrive()`. If the spawn would exceed the limit, the node should fail with a clear `NodeFailureException` explaining the limit was hit, without creating any child threads. This keeps the WfRun state consistent and allows failure handlers on the node to catch the error.

- **Acceptable:** If the limit is detected after spawning, all already-created child threads should receive a halt signal (via `ThreadHaltReasonModel`) and be transitioned to a terminal state before the WfRun itself transitions to `ERROR`. The workflow should reach a fully consistent terminal state.

### Affected Components

Kernel (/server)

### Environment

- **Server Version:** v1.0.1 (confirmed present on latest `main` as well — the logic is unchanged)
- **OS:** Linux (also reproducible on macOS)
- **Java:** Corretto 21

### Documentation

_No response_

## Root Cause

In `WfRunModel.advance()`, the max-thread-run check occurs at the top of the while-loop, but `StartMultipleThreadsRunModel.arrive()` is invoked during thread advancement *within* the same loop iteration. The `arrive()` method unconditionally spawns all child threads:

```java
// StartMultipleThreadsRunModel.arrive() — v1.0.1
VariableValueModel iterable =
        nodeRun.getThreadRun().assignVariable(node.getIterable()).asArr();
for (Object threadInput : iterable.getJsonArrVal()) {
    VariableValueModel iterInput =
            LHSerializable.fromProto(LHLibUtil.objToVarVal(threadInput), VariableValueModel.class, context);
    // ... construct input variables ...
    ThreadRunModel child = nodeRun.getThreadRun()
            .getWfRun()
            .startThread(threadSpecName, time, parentThreadNumber, inputs, ThreadType.CHILD);
    childThreadIds.add(child.getNumber());
}
```

Each `startThread()` call immediately adds the new thread to `threadRunsUseMeCarefully` and puts it in `RUNNING` status:

```java
// WfRunModel.startThread()
threadRunsUseMeCarefully.add(newThread);
newThread.createVariablesAndStart(variables);  // sets status to RUNNING
return newThread;
```

On the next iteration of the `advance()` while-loop, the size check fires:

```java
if (this.threadRunsUseMeCarefully.size()
        > this.executionContext.serverConfig().getMaxThreadRunsPerWfRun()) {
    putFailureOnThreadRun(
            getThreadRun(0),
            new FailureModel(
                    String.format("You exceeded the maximum number of ThreadRuns per WfRun: %s",
                            this.executionContext.serverConfig().getMaxThreadRunsPerWfRun()),
                    LHErrorType.INTERNAL_ERROR.toString()),
            time,
            null);
    transitionTo(LHStatus.ERROR);
    break;
}
```

The `putFailureOnThreadRun` method adds the failure to thread 0's current NodeRun (the `START_MULTIPLE_THREADS` NodeRun), sets thread 0 to `ERROR`, and calls `nodeRun.maybeHalt(null)`. Since `StartMultipleThreadsRunModel` does not override `maybeHalt()`, the base `SubNodeRun.maybeHalt()` returns `!nodeRun.isInProgress()` which is `false` (the NodeRun is in `STARTING` state), so the NodeRun is set to `HALTING`.

Then `transitionTo(LHStatus.ERROR)` sets the WfRun to `ERROR` and `break` exits the loop.

**No halt signal is ever propagated to the child threads.** They remain in `RUNNING` status because:

1. The WfRun is now in a terminal state (`ERROR`), so `advance()` will never be called again.
2. No `ThreadHaltReasonModel` was created for the children.
3. The children were never advanced (their tasks were never scheduled), so no timeout timer exists to clean them up.
4. `archiveCompletedThreadRuns()` only archives `COMPLETED` threads — not `RUNNING` ones.

The result is a permanently inconsistent state: WfRun is `ERROR` but contains `RUNNING` child threads with `STARTING` NodeRuns.

## Impact

- Workflows using `spawnThreadForEach` with dynamically-sized arrays will hit this limit once the array grows past 64 elements (with default config). The boundary is `array_size + 1 > LHS_X_MAX_THREAD_RUNS_PER_WF_RUN`.
- While the top-level WfRun status is `ERROR` and a failure message exists on thread 0's NodeRun, the failure mode is difficult to diagnose in practice: the last successfully-completed task (which returned the array) appears to be the point of failure, the child threads show as `RUNNING`, and there is no warning at `WfSpec` registration time that the limit might be exceeded at runtime.
- Increasing `LHS_X_MAX_THREAD_RUNS_PER_WF_RUN` is a workaround but is bounded by `LHS_PRODUCER_MAX_REQUEST_SIZE` (default: ~1MB) — large numbers of threads will eventually cause `RecordTooLargeException` which is caught by `LHProcessingExceptionHandler` and silently dropped (a separate but compounding issue).
- Affected WfRuns cannot recover without manual intervention.

## Suggested Fix

Check the limit inside `StartMultipleThreadsRunModel.arrive()` before spawning any threads:

```java
@Override
public void arrive(Date time, CoreProcessorContext processorContext) throws NodeFailureException {
    StartMultipleThreadsNodeModel node = getNode().getStartMultipleThreadsNode();
    try {
        VariableValueModel iterable =
                nodeRun.getThreadRun().assignVariable(node.getIterable()).asArr();
        List<?> items = iterable.getJsonArrVal();

        int currentThreadCount = nodeRun.getThreadRun().getWfRun().getThreadRunsUseMeCarefully().size();
        int maxAllowed = processorContext.serverConfig().getMaxThreadRunsPerWfRun();

        if (currentThreadCount + items.size() > maxAllowed) {
            FailureModel failure = new FailureModel();
            failure.message = String.format(
                "spawnThreadForEach would create %d total threads (current: %d + requested: %d), "
                + "exceeding LHS_X_MAX_THREAD_RUNS_PER_WF_RUN (%d). "
                + "Reduce array size or increase the server limit.",
                currentThreadCount + items.size(), currentThreadCount, items.size(), maxAllowed);
            failure.failureName = LHConstants.INTERNAL_ERROR;
            throw new NodeFailureException(failure);
        }

        for (Object threadInput : items) {
            // ... existing spawn logic unchanged ...
        }
    } catch (LHVarSubError | LHSerdeException | LHValidationException e) {
        FailureModel failure = new FailureModel();
        failure.message = "Failed constructing input variables for thread: " + e.getMessage();
        failure.failureName = LHConstants.VAR_SUB_ERROR;
        throw new NodeFailureException(failure);
    }
}
```

This ensures:
- No threads are created if the limit would be exceeded
- The failure is surfaced as a proper `NodeFailureException` with a descriptive message
- The WfRun state remains consistent (no orphan threads)
- Failure handlers on the node can potentially catch and handle the error
- The error message tells the operator exactly what happened and what to do

Snarr:
Resolves #2357 

## Background

See Proposal #12 "Archive ThreadRuns" proposal for information about the ThreadRun Archival mechanism.

The bug: when executing a StartMultipleThreads NodeRun that pushes the active `ThreadRun` count over the `LHS_X_MAX_THREAD_RUNS_PER_WF_RUN` limit, the new ThreadRuns are spawned and then the `WfRun` catches the limit on the next advance call, but now there are orphaned child threads stuck forever.

## Solution

1. I added the limit check within the `startThread()` method so that we can catch a limit pass on a per-ThreadRun basis, making the check as atomic as possible. It's important to have this conditional check at this low level to prevent other edge cases unrelated to `SpawnMultipleThreadsRunModel` from occurring.

2. `StartMultipleThreadsRunModel` now checks if the `currentThreadRuns + requestedThreadRuns > maxThreadRuns` before spawning any child thread, making the NodeRun fail early before even calling the `startThread()` method.

### DIFF
--- a/server/src/test/java/e2e/MaxThreadRunCountTest.java
+++ b/server/src/test/java/e2e/MaxThreadRunCountTest.java
@@ -57,19 +57,28 @@ public class MaxThreadRunCountTest {
 
     @Test
     void shouldNotLeaveOrphanThreadsWhenSpawnExceedsLimit() {
+        int maxThreadRuns = this.serverConfig.getMaxThreadRunsPerWfRun();
         ArrayList<Integer> largeArr = new ArrayList<>();
-        for (int i = 0; i < this.serverConfig.getMaxThreadRunsPerWfRun(); i++) {
+        for (int i = 0; i < maxThreadRuns; i++) {
             largeArr.add(i);
         }
 
+        // 1 entrypoint ThreadRun + maxThreadRuns requested children would exceed the limit.
+        String expectedFailureMessage = String.format(
+                "WfRun would have %d ThreadRuns, exceeding the maximum number of ThreadRuns per WfRun: %d. "
+                        + "Reduce the number of spawned ThreadRuns or increase LHS_X_MAX_THREAD_RUNS_PER_WF_RUN.",
+                maxThreadRuns + 1, maxThreadRuns);
+
         verifier.prepareRun(spawnManyThreadsWf, Arg.of("json-arr", largeArr))
                 .waitForStatus(LHStatus.ERROR)
+                .thenVerifyNodeRun(0, 1, nodeRun -> {
+                    Assertions.assertThat(nodeRun.getStatus()).isEqualTo(LHStatus.ERROR);
+                    Assertions.assertThat(nodeRun.getFailuresList().get(0).getMessage())
+                            .isEqualTo(expectedFailureMessage);
+                })
                 .thenVerifyWfRun(wfRun -> {
                     Assertions.assertThat(wfRun.getThreadRunsCount()).isEqualTo(1);
-
-                    Assertions.assertThat(wfRun.getThreadRunsList())
-                            .noneMatch(threadRun -> threadRun.getStatus() == LHStatus.RUNNING
-                                    || threadRun.getStatus() == LHStatus.STARTING);
+                    Assertions.assertThat(wfRun.getThreadRuns(0).getStatus()).isEqualTo(LHStatus.ERROR);
                 })
                 .start();
     }
--- a/server/src/test/java/e2e/MaxThreadRunCountTest.java
+++ b/server/src/test/java/e2e/MaxThreadRunCountTest.java
@@ -48,9 +48,28 @@ public class MaxThreadRunCountTest {
         verifier.prepareRun(spawnManyThreadsWf, Arg.of("json-arr", largeArr))
                 .waitForStatus(LHStatus.ERROR)
                 .thenVerifyNodeRun(0, 1, nodeRun -> {
-                    Assertions.assertThat(nodeRun.getStatus()).isEqualTo(LHStatus.HALTING);
+                    Assertions.assertThat(nodeRun.getStatus()).isEqualTo(LHStatus.ERROR);
                     Assertions.assertThat(nodeRun.getFailuresList().get(0).getMessage())
-                            .contains("You exceeded the maximum number of ThreadRuns per WfRun");
+                            .contains("exceeding the maximum number of ThreadRuns per WfRun");
+                })
+                .start();
+    }
+
+    @Test
+    void shouldNotLeaveOrphanThreadsWhenSpawnExceedsLimit() {
+        ArrayList<Integer> largeArr = new ArrayList<>();
+        for (int i = 0; i < this.serverConfig.getMaxThreadRunsPerWfRun(); i++) {
+            largeArr.add(i);
+        }
+
+        verifier.prepareRun(spawnManyThreadsWf, Arg.of("json-arr", largeArr))
+                .waitForStatus(LHStatus.ERROR)
+                .thenVerifyWfRun(wfRun -> {
+                    Assertions.assertThat(wfRun.getThreadRunsCount()).isEqualTo(1);
+
+                    Assertions.assertThat(wfRun.getThreadRunsList())
+                            .noneMatch(threadRun -> threadRun.getStatus() == LHStatus.RUNNING
+                                    || threadRun.getStatus() == LHStatus.STARTING);
                 })
                 .start();
     }
--- a/server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/WfRunModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/WfRunModel.java
@@ -20,10 +20,10 @@ import io.littlehorse.common.model.corecommand.subcommand.ResumeWfRunRequestMode
 import io.littlehorse.common.model.corecommand.subcommand.SleepNodeMaturedModel;
 import io.littlehorse.common.model.corecommand.subcommand.StopWfRunRequestModel;
 import io.littlehorse.common.model.getable.core.externalevent.ExternalEventModel;
+import io.littlehorse.common.model.getable.core.noderun.NodeFailureException;
 import io.littlehorse.common.model.getable.core.noderun.NodeRunModel;
 import io.littlehorse.common.model.getable.core.variable.VariableValueModel;
 import io.littlehorse.common.model.getable.core.wfrun.failure.FailureBeingHandledModel;
-import io.littlehorse.common.model.getable.core.noderun.NodeFailureException;
 import io.littlehorse.common.model.getable.core.wfrun.failure.FailureModel;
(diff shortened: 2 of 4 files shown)


server/src/test/java/e2e/MaxThreadRunCountTest.java
```diff
@@ -48,9 +48,28 @@ void shouldFailIfSpawnTooManyWf() {
...
+            largeArr.add(i);
+        }
+
+        verifier.prepareRun(spawnManyThreadsWf, Arg.of("json-arr", largeArr))
+                .waitForStatus(LHStatus.ERROR)
+                .thenVerifyWfRun(wfRun -> {
+                    Assertions.assertThat(wfRun.getThreadRunsCount()).isEqualTo(1);
+
+                    Assertions.assertThat(wfRun.getThreadRunsList())
+                            .noneMatch(threadRun -> threadRun.getStatus() == LHStatus.RUNNING
```
eduwercamacaro: I believe it is better if we assert the exact status we expect when the number of threads run is exceeded. It confuses me because this is basically saying that the only two status that are not expected here are `RUNNING` and `STARTING`

eduwercamacaro (approved):

server/src/test/java/e2e/MaxThreadRunCountTest.java
coltmcnealy-lh: I'd also like to assert the actual `Failure` message is expected and makes some form of sense.

coltmcnealy-lh (approved):

server/src/test/java/e2e/MaxThreadRunCountTest.java
Snarr: I understand your confusion, I was just trying to be as true to the original bug as possible. We want to make sure there are "no orphaned child threads in the  running or starting state when we've reached the max thread run limit". The exact state of the stopped ones does not matter as much.

eduwercamacaro:
This PR needs to be cherry-picked into 1.0 and 1.1 release branches

eduwercamacaro:
I cherry-picked this into 1.1 release branch. 

There are some conflicts when merging this into 1.0. @Snarr I'm not sure if we need this fix in 1.0 branch. Can you confirm please?

PR #2364, opened 2026-06-24
fix(server): close iterators on readonly base store

coltmcnealy-lh:


l4crito (approved):

eduwercamacaro:
I think we need to cherry-pick this into 1.0 and 1.1 as this is a bug that got released since 1.0

coltmcnealy-lh:
We should cherry-pick this to `1.0` and `1.1`.

PR #2368, opened 2026-06-25
feat(server): add LittleHorse Native Maps

Snarr:
This PR adds LittleHorse native maps to the `server` and `sdk-java`. See the Maps proposal for more information (https://github.com/littlehorse-enterprises/littlehorse/blob/feat/add-maps-to-lh/proposals/023-native-maps.md)


server/src/main/java/io/littlehorse/common/model/getable/core/variable/VariableValueModel.java
```diff
@@ -940,6 +1059,14 @@ public VariableValueModel asArray() throws LHVarSubError {
         }
     }
 
+    public VariableValueModel asMap() throws LHVarSubError {
+        if (getTypeDefinition().getDefinedTypeCase() == DefinedTypeCase.INLINE_MAP_DEF) {
+            return new VariableValueModel(map);
```
eduwercamacaro, line 1069: For empty (aka null) VariableValues whose type is a `MAP`, this sentence here will throw a RuntimeException because, according to this constructor, it can't take null arguments. Am I right?

server/src/main/java/io/littlehorse/common/model/getable/core/variable/MapModel.java
```diff
@@ -0,0 +1,110 @@
...
+    public MapModel(MapModel other) {
+        if (other == null) {
+            throw new IllegalArgumentException("Cannot copy-construct MapModel from null");
+        }
+
+        this.entries = new ArrayList<>();
+        if (other.entries != null) {
+            for (MapEntryModel entry : other.entries) {
+                this.entries.add(new MapEntryModel(
+                        entry.getKey() == null ? null : entry.getKey().getCopy(),
```
eduwercamacaro: This doesn't seem to be fully consistent with the initFrom and toProto methods. 

Here we accept `null` keys and values, but we are not adding null safeguards on the `toProto`.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WfRunVariableImpl.java
```diff
@@ -128,19 +128,22 @@ public WfRunVariableImpl get(String field) {
...
                         && typeDef.getPrimitiveType() != VariableType.JSON_OBJ) {
                     throw new LHMisconfigurationException(
-                            "Can only use get() on JSON_OBJ, JSON_ARR, or Struct variables");
+                            "Can only use get(String key) on JSON_OBJ, JSON_ARR, Map, or Struct variables");
                 }
                 break;
             case INLINE_ARRAY_DEF:
-                // Typed inline arrays (e.g. declareArray("x", Long.class)) are allowed to be
-                // indexed with get(int).
+                throw new LHMisconfigurationException(
```
eduwercamacaro, line 135: Can we add this exception to the javadoc of the `get()` method? This could be interpreted as a breaking change because, before this PR, we were silently setting a default value.

server/src/main/java/io/littlehorse/common/model/getable/core/variable/VariableValueModel.java
```diff
@@ -623,6 +661,41 @@ public VariableValueModel extend(VariableValueModel rhs) throws LHVarSubError {
                 } catch (Exception exn) {
                     throw new LHVarSubError(exn, "Failed extending ARRAY");
                 }
+            case MAP:
+                MapModel thisMap = this.map;
+                if (rhs.getValueType() != ValueCase.MAP) {
+                    throw new LHVarSubError(null, "Cannot extend MAP with non-MAP value");
+                }
+                MapModel rhsMap = rhs.getMap();
+                if (rhsMap.getEntries().size() != 1) {
```
eduwercamacaro: Uh, this is hard to reason about as a client. 


I would expect to merge two Map values into a single one when using the `EXTEND` operation. Just like arrays.

Snarr: Sorry, this is something I updated in the proposal but haven't updated here. I've marked this PR as a `work in progress` as I improve the PR to reflect the proposal. Thanks for the early feedback though. Proposal: #2374

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/structdefutil/LHMapType.java
```diff
@@ -0,0 +1,63 @@
...
+                    "Map key type must resolve to a primitive VariableType. Provided key class: " + keyClazz.getName()
+                            + " resolves to " + resolvedKeyType.getDefinedTypeCase()
+                            + ". Only primitive types (STR, INT, DOUBLE, BOOL, BYTES, WF_RUN_ID, TIMESTAMP) are allowed as map keys.");
+        }
+        this.keyType = resolvedKeyType;
+
+        if (valueClazz.isArray()) {
+            this.valueType = new LHArrayType(valueClazz, typeAdapterRegistry);
+        } else {
+            this.valueType = LHClassType.fromJavaClass(valueClazz, typeAdapterRegistry);
```
eduwercamacaro, line 31: For `Map` values, this line will transform it into a `JSON_OBJ`. We would need to modify this method to support `Map` values. am I missing something?

server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/VariableAssignmentModel.java
```diff
@@ -393,6 +393,10 @@ private boolean isValidSizeOfOperand(TypeDefinitionModel operandType) {
             return true;
         }
 
+        if (operandType.getDefinedTypeCase() == DefinedTypeCase.INLINE_MAP_DEF) {
```
eduwercamacaro: There is an exception message that needs to be updated on the `VariableAssignmentModel.getSourceType` method

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/structdefutil/LHStructProperty.java
```diff
@@ -167,13 +174,40 @@ public LHClassType getPropertyType(LHTypeAdapterRegistry typeAdapterRegistry) {
...
+
         return LHClassType.fromJavaClass(pd.getPropertyType(), typeAdapterRegistry);
     }
 
     private boolean isNativeArray() {
         return pd.getPropertyType().isArray() && !byte[].class.equals(pd.getPropertyType());
     }
 
+    private boolean isNativeMap() {
+        return Map.class.isAssignableFrom(pd.getPropertyType());
```
eduwercamacaro, line 189: I think this is a breaking change because any existing struct class will now default to a native map. Before this PR it was resolving to JSON_OBJ, that's why I think this is going to break existing client code.

If this is expected, we would need to add it to the proposal.

Snarr: Previously, Struct classes could not resolve JSON_OBJs because we disallow JSON types in Structs.

server/src/main/java/io/littlehorse/common/model/getable/core/variable/VariableValueModel.java
Snarr: When a VariableValue is null, isn't its returned `TypeDefinition.definedTypeCase` going to be `TypeDefinition.DEFINEDTYPE_NOT_SET`? Specifically, given only a `VariableValue` we cannot tell it's `TypeDefinition` when the `VariableValue` is `null`.

mijailr:
I'm interested in this.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/structdefutil/LHMapType.java
Snarr: I don't think we can support nested Maps using Java reflection at the moment, so that is why it is written this way.

server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/VariableAssignmentModel.java
Snarr: We've since removed this functionality of calling SIZE_OF on Maps, I don't want that to be possible. Down the road, people can call SIZE_OF on `map.keys()` which can return an Array.

Snarr:
> I'm interested in this.

@mijailr Great to hear, on track for this functionality to make it into 1.2.

If you have the time, please check out the Proposal for this PR (https://github.com/littlehorse-enterprises/littlehorse/blob/feat/add-maps-to-lh/proposals/023-native-maps.md) and see if you agree with the semantics of Maps and if they work for your use case.

sdk-java/src/main/java/io/littlehorse/sdk/wfsdk/internal/WfRunVariableImpl.java
Snarr: There are a lot of other parts of the codebase that reference this Exception, and I actually think these parts of the codebase misuse this exception. This `LHMisconfigurationException` is used for 3 different types of exceptions, so I want to create a follow-up PR to refactor this and add the proper javadoc notes.

server/src/main/java/io/littlehorse/common/model/getable/core/variable/VariableValueModel.java
eduwercamacaro: Ohh right! That's the part I was missing.

eduwercamacaro (approved):

PR #2379, opened 2026-06-30
feat(server): thread run queue

JakeRose03:
## Summary

Adds a persistent `ThreadRun` queue so a `WfRun` can create more threads than the configured in-memory limit without failing at runtime.

Previously, exceeding `LHS_X_MAX_THREAD_RUNS_PER_WF_RUN` produced an internal error. This change redefines the limit as `LHS_X_ACTIVE_THREAD_RUNS_PER_WF_RUN`: only that many `ThreadRun`s may be active in memory, while additional threads are queued and started as capacity becomes available.

## Implementation

- Adds an ordered `thread_run_queue` to `WfRun`.
- Extends `InactiveThreadRun` with an `inactive_reason` oneof:
  - `archived` for completed threads removed from the active collection.
  - `queued` for threads waiting for an active slot.
- Stores queued thread input variables in `QueuedThreadRunInfo`.
- Assigns queued threads their final thread-run number immediately.
- Records queued threads in the parent's `child_thread_ids`, ensuring the parent waits for them before completing.
- Drains the queue whenever the `WfRun` advances and active capacity is available.
- Moves activated threads into the active collection, initializes their variables, starts execution, and deletes the queued `InactiveThreadRun`.
- Preserves queue order while allowing a halted queue entry to be deferred so later eligible entries can start.
- Propagates and resolves halt reasons for queued threads without accessing a `NodeRun`, since queued threads have not started one.
- Prevents queued threads from starting after the `WfRun` becomes terminal.
- Treats older `InactiveThreadRun` records without an `inactive_reason` as archived for backward compatibility.
- Renames the server setting from `LHS_X_MAX_THREAD_RUNS_PER_WF_RUN` to `LHS_X_ACTIVE_THREAD_RUNS_PER_WF_RUN`.

## Behavior

When the active limit is reached, newly created child threads are persisted with status `STARTING` instead of being added to the active thread collection. As active threads complete and are archived, queued threads are activated until the queue is empty.

A queued thread with unresolved halt reasons remains queued. It does not block later eligible queue entries from starting.

## Testing

Added end-to-end coverage verifying:

- Excess threads are queued and eventually complete.
- Queued threads are persisted as `InactiveThreadRun` records.
- Halted queued threads are not activated.
- A halted queue head does not block later eligible entries.
- Queued threads activate after parent halt reasons are resolved.
- Parent failures propagate halt reasons to queued children.


proposals/023-thread-run-queue.md
```diff
@@ -0,0 +1,74 @@
...
+
+It stores completed/archived `threadRun`s to free space on the `wfRun` object, allowing more `threadRun`s to be spawned. `InactiveThreadRun` should encompass all `threadRun`s not held on the `wfRun`; in order to do this we need to extend the current proto schema while keeping in mind backwards compatibility. 
+
+The new `InactiveThreadRun` schema:
+
+```proto
+message InactiveThreadRun {
+  ThreadRun thread_run = 1;
+
+  oneof inactive_thread_run_type {
```
eduwercamacaro: nit: 
I see this more as a `reason` a not so much as a `type`. We can define it as `oneof inactive_reason`. what do you think?

eduwercamacaro (commented):
First pass. just reading the proposal.

proposals/023-thread-run-queue.md
JakeRose03: Yea, reason is a better choice of words.

Snarr (approved):
Looks good 👍

Issue #1998, opened 2025-12-21
PutWfSpec creates new revisions for identical workflow definitions (non-deterministic compilation)


jcustenborder:
### Context

When repeatedly calling PutWfSpec with the same workflow definition (no code changes), the revision number sporadically increments. This suggests the compiled WfSpec is not deterministic, causing LittleHorse to detect "changes" that don't exist.

  The revision number increments sporadically even with no changes to the workflow:
  - Runs 1-4: revision=1 (correct)
  - Run 5: revision=2 (WRONG - no change, but revision incremented)
  - Runs 6-7: revision=2 (correct)
  - Run 8: revision=3 (WRONG - no change, but revision incremented)
  - And so on...

### Steps To Reproduce

  1. Save the reproduction script below as lh_version_repro.py
  2. Start LittleHorse server via Docker
  3. Register the mock tasks in one terminal:
  python lh_version_repro.py tasks
  4. In another terminal, run upload repeatedly:
  python lh_version_repro.py upload
  python lh_version_repro.py upload
  python lh_version_repro.py upload
  ... repeat 10+ times
  5. Observe revision numbers incrementing despite no code changes between runs

```
#!/usr/bin/env python3
"""LittleHorse version bug reproduction."""

import argparse
import asyncio
import logging
from typing import Any

import littlehorse
from littlehorse.config import LHConfig
from littlehorse.model import PutWfSpecRequest
from littlehorse.worker import LHTaskWorker, WorkerContext
from littlehorse.workflow import Workflow, WorkflowThread

logging.basicConfig(
    level=logging.INFO,
    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
)
logger = logging.getLogger(__name__)

UPDATE_EMBEDDINGS_TASK = "repro-update-embeddings"
UPDATE_RECORD_TASK = "repro-update-record"
DETECT_FACES_TASK = "repro-detect-faces"
GRADE_IMAGE_TASK = "repro-grade-image"

WORKFLOW_NAME = "repro-image-processing"


async def mock_update_embeddings(
    url: str, record_type: str, tenant_id: str, creator_id: str, record_id: str, ctx: WorkerContext,
) -> dict[str, Any]:
    return {}


async def mock_update_record(
    tenant_id: str, creator_id: str, record_id: str, updates: dict[str, Any], ctx: WorkerContext,
) -> dict[str, Any]:
    return {"updated_fields": list(updates.keys()) if updates else []}


[...]

PR #2381, opened 2026-07-03
fix(server): canonicalize WfSpec equality

jcustenborder:
## Summary

Fixes nondeterministic `PutWfSpec` revision bumps for semantically identical workflow definitions by making `WfSpecUtil.equals()` compare canonicalized protobufs instead of raw serialized bytes.

The old equality check compared `toByteArray()` output after only normalizing `id` and `created_at`. That made equality sensitive to protobuf serialization order, especially for server-derived repeated fields such as `frozen_variables`. Equivalent specs could therefore look different and cause an unnecessary WfSpec revision.

## Changes

- Canonicalize `frozen_variables` by variable name before WfSpec equality comparison.
- Replace raw byte-array comparison with semantic protobuf equality after sanitization.
- Add regression coverage for reordered frozen variables after protobuf serde.
- Add seeded shuffled-order coverage to reproduce order sensitivity deterministically.
- Add a representative Java SDK-compiled workflow similar to issue #1998, including JSON input, public/searchable vars, JSON path assignments, task outputs, and multiple task nodes.

## Test Plan

- [x] `./gradlew server:spotlessCheck`
- [x] `./gradlew server:test --tests io.littlehorse.common.util.WfSpecUtilTest`
- [x] `./gradlew server:test --tests io.littlehorse.common.model.metadatacommand.PutWfSpecRequestModelTest`

## Notes

This keeps the fix scoped to server-side WfSpec equality. No SDK serialization behavior changes are required because the server should treat semantically equivalent protobufs as equal regardless of valid wire/map ordering.


jcustenborder:
@coltmcnealy-lh and @eduwercamacaro, Can you all let me know what you think about this one? I think I have put together a fairly representative test of what was going on. I was able to run `python lh_version_repro.py upload` 50 times without seeing a revision change.

server/src/main/java/io/littlehorse/common/util/WfSpecUtil.java
```diff
@@ -28,11 +31,21 @@ public static boolean equals(WfSpecModel left, WfSpecModel right) {
         sanitize(copy, date);
         WfSpec.Builder toCopy = right.toProto();
         sanitize(toCopy, date);
-        return Arrays.equals(copy.build().toByteArray(), toCopy.build().toByteArray());
+        return copy.build().equals(toCopy.build());
     }
 
     private static void sanitize(WfSpec.Builder spec, Timestamp date) {
         spec.setId(WfSpecId.newBuilder().setName(spec.getId().getName())).setCreatedAt(date);
+        sortFrozenVariables(spec);
```
eduwercamacaro, line 39: I'm wondering if we need to sort every array in the inner objects inside the WfSpec.. For example, `WfSpec.equals` also calls `ThreadSpec.equals` method under the hood. Do we need to also sort the array fields inside the ThreadSpec object?

eduwercamacaro (commented):
Thank you for taking time to work on this. 

I believe comparing WfSpecs like this is the right direction. Currently, we just compare at the byte array level, which brings some unexpected behavior like the one you are trying to solve. However, I'm not sure if we need to do more complex logic to sanitize the WfSpec before comparing it.

eduwercamacaro:
@mijailr Do you have any thoughts on this?

mijailr:
> @mijailr Do you have any thoughts on this?

Nothing from me.

server/src/main/java/io/littlehorse/common/util/WfSpecUtil.java
jcustenborder: Yeah, offhand, I think we need to. I'm wondering if things like HashMaps are doing some weirdness around ordering, which would, of course, mess with the byte array. My main concern is trying to get back to the point where I can use CI to deploy my application without continually cycling the versions on the workflows when nothing changed.

jcustenborder:
@eduwercamacaro @mijailr Do you all need anything from me?

server/src/main/java/io/littlehorse/common/util/WfSpecUtil.java
eduwercamacaro: We have had some internal discussions about this. I agree that this solves the issue of bumping the WfSpec's version when the order on `fronzenVariables` collection changes, but there are other collections that need to also be sorted before comparing the two wfspecs. I'm going to propose an implementation where we check every field at Model level to determine if we need to bump the version of an existing WfSpec when processing `PutWfSpec` requests.

eduwercamacaro (approved):
I'm approving this PR as it solves the problem of ordering of the `frozenVaraiables` collections. However, this issue will also be present on more complex wfspecs as we are not sorting the rest of collections that are present in a WfSpec.

PR #2384, opened 2026-07-07
feat(lhctl): add `Map`, `Array`, and `Struct` support

Snarr:
This PR adds `Map`, `Array`, and `Struct` support to `lhctl`. 

It updates Proposal 023 to add details regarding how we will support Maps, Arrays, and Structs in `lhctl` alongside the existing `JSON_OBJ` and `JSON_ARR` types.

This PR also updates the Java and .NET Maps, Arrays, and StructDefs examples to use `lhctl` commands instead of manual inputs.


eduwercamacaro (approved):
Thanks! Good improvement

PR #2385, opened 2026-07-07
feat(server): support Array concatenation using EXTEND

### LINKED ISSUE #2388: Support Array Concatenation Using `EXTEND`
### Background

1. We have native LittleHorse Arrays with fixed types
2. Right now, when you call `EXTEND` on an `Array` it functions like an `APPEND` operation (it should probably be renamed).
3. Calling `EXTEND` on a `JSON_ARR` with an `RHS` of another `JSON_ARR` concatenates the two `JSON_ARR`s.
4. Calling `EXTEND` on an `Array` with an `RHS` of a single item of that `Array`s type appends the item to the `Array`.
5. Calling `EXTEND` on an `Array` with an `RHS` of another `Array` does not work currently.

### Proposal

We should make it possible to call `EXTEND` on an `Array` with an `RHS` of another `Array` of the same type. This would concatenate the two `Array`s with the content of the `LHS` preceding the `RHS`.

For example:
```java
WfRunVariable intArr1 = wf.declareArray("my-arr-1", Integer[].class);
WfRunVariable intArr2 = wf.declareArray("my-arr-2", Integer[].class);

intArr1.assign(intArr1.extend(intArr2));
```

If you pass `[1,2,3]` into `my-arr-1` and `[7,8,9]` into `my-arr-2` in the above WfSpec, the final value of `my-arr-1` should resolve to `[1,2,3,7,8,9]`.

### Acceptance Criteria

1. Support concatenating same-type Arrays using the `EXTEND` operation
2. The LHS should precede the RHS in the concatenation

### Out of Scope

_No response_

### Affected Components

Kernel (/server)

### Documentation

_No response_

### Additional Context

_No response_

Snarr:
This PR updates the Array functionality to support concatenation using two like Arrays with the `EXTEND` operation.

Resolves #2388


server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/expression/ArrayReturnTypeStrategy.java
```diff
@@ -22,13 +22,13 @@ public TypeDefinitionModel getIdentity() {
     @Override
     public Optional<TypeDefinitionModel> add(ReadOnlyMetadataManager manager, LHTypeStrategy other)
             throws InvalidExpressionException {
-        throw new UnsupportedOperationException("Cannot add to an Array. Use EXTEND instead.");
+        throw new InvalidExpressionException("Cannot add to an Array. Use EXTEND instead.");
```
Snarr, line 25: Previously, when an UnsupportedOperationException was thrown here, the error did not propagate back to the client and the server returned an internal server error.

eduwercamacaro (approved):

server/src/test/java/e2e/ArraysTest.java
```diff
@@ -195,6 +198,26 @@ public void shouldAppendItemOnExtend() {
                 .start();
     }
 
+    @Test
+    public void shouldConcatenateArraysOnExtend() {
```
eduwercamacaro, line 205: I believe would be useful to test the case where we EXTEND an array of arrays.

Snarr: @eduwercamacaro Testing this is hard because of our inability to handle nested Arrays due to the limitations of Java Generics. I will try to add an e2e test that tests this by manually registering and sending a nested array.

PR #2391, opened 2026-07-08
fix(server): gate BigDecimal JSONPath conversion

### LINKED ISSUE #2390: JSONPath numeric field returns BigDecimal and crashes server during REPORT_TASK_RUN
## Context

LittleHorse server `v1.1.1` throws an unhandled `RuntimeException` while processing `REPORT_TASK_RUN` when a workflow assigns a JSONPath-selected decimal value from task output.

Confirmed from server logs:

```text
ERROR [LH] VariableValueModel - Not possible to get this from jsonpath 0.47380000352859497=class java.math.BigDecimal
ERROR [LH] LHProcessingExceptionHandler - Caught exception processing REPORT_TASK_RUN command:
java.lang.RuntimeException: Not possible to get this from jsonpath
```

The failure repeats with multiple decimal values such as:

```text
0.45170000195503235=class java.math.BigDecimal
0.49570000171661377=class java.math.BigDecimal
0.46309998631477356=class java.math.BigDecimal
```

The server code path appears to reject `java.math.BigDecimal` in `VariableValueModel.jsonPath()`, even though JSONPath can return `BigDecimal` for decimal JSON numbers.

Possibly related: https://github.com/littlehorse-enterprises/littlehorse/issues/1106

## Steps To Reproduce

1. Register a WfSpec that executes a task returning a `JSON_OBJ` containing a decimal value:

   ```json
   {
     "score": 0.47380000352859497
   }
   ```

2. In the WfSpec, assign that task output field using JSONPath:

   ```python
   result = wf.execute("return-score", ...)
   score = wf.declare_double("score")
   score.assign(result.with_json_path("$.score"))
   ```

3. Run the workflow and let the worker report the task run successfully.

4. Observe the LittleHorse server log:

   ```text
   VariableValueModel - Not possible to get this from jsonpath 0.47380000352859497=class java.math.BigDecimal
   LHProcessingExceptionHandler - Caught exception processing REPORT_TASK_RUN command:
   java.lang.RuntimeException: Not possible to get this from jsonpath
   ```

## Expected Behavior

LittleHorse should not crash command processing with an unhandled `RuntimeException`.

Expected behavior should be one of:

- Treat JSONPath `BigDecimal` numeric results as `DOUBLE`, or
- fail the WfRun/task with a clear `LHVarSubError` explaining that the JSONPath result type is unsupported.

The error should include the JSONPath, value, and Java type.

## Affected Components

- Kernel (/server)

## Environment

```text
OS: Linux container
Client Version: Python SDK / littlehorse-client 1.1.1
Server Version: v1.1.1 / server-1.1.1.jar
Java: unknown
```

## Documentation

If `BigDecimal` is intentionally unsupported, JSONPath documentation should state which scalar result types are supported for variable assignment.

## Additional Context

Stack trace:

```text
java.lang.RuntimeException: Not possible to get this from jsonpath
  at io.littlehorse.common.model.getable.core.variable.VariableValueModel.jsonPath(VariableValueModel.java:448)
  at io.littlehorse.common.model.getable.core.wfrun.ThreadRunModel.assignVariable(ThreadRunModel.java:909)
  at io.littlehorse.common.model.getable.global.wfspec.variable.VariableMutationModel.getRhsValue(VariableMutationModel.java:129)
  at io.littlehorse.common.model.getable.global.wfspec.variable.VariableMutationModel.execute(VariableMutationModel.java:150)
  at io.littlehorse.common.model.getable.global.wfspec.node.EdgeModel.mutateVariables(EdgeModel.java:164)
  at io.littlehorse.common.model.getable.core.noderun.NodeRunModel.evaluateOutgoingEdgesAndMaybeMutateVariables(NodeRunModel.java:648)
  at io.littlehorse.common.model.getable.core.wfrun.ThreadRunModel.advance(ThreadRunModel.java:566)
  at io.littlehorse.common.model.getable.core.wfrun.WfRunModel.advance(WfRunModel.java:555)
  at io.littlehorse.common.model.getable.core.taskrun.TaskRunModel.onTaskAttemptResultReported(TaskRunModel.java:392)
  at io.littlehorse.common.model.corecommand.subcommand.ReportTaskRunModel.process(ReportTaskRunModel.java:55)
```

The immediate failing value/type log confirms the unsupported type is `java.math.BigDecimal`.

jcustenborder:
## Summary
- Add opt-in `LHS_X_JSONPATH_BIG_DECIMAL_TO_DOUBLE` server config, cached as a boolean.
- Convert JsonPath `BigDecimal` results to `DOUBLE` only when enabled.
- Keep disabled behavior as `RuntimeException` with type-only message to avoid value leakage.

Closes #2390

## Test Plan
- [x] `./gradlew server:test --tests io.littlehorse.common.LHServerConfigTest --tests io.littlehorse.common.model.wfrun.VariableValueModelTest`
- [x] `./gradlew server:test`


jcustenborder:
Operational follow-up: the new JSONPath error message avoids leaking customer JSON, which is good, but it still does not tell operators where the failure happened. At minimum we need wfRunId and step context (threadRunNumber, nodeRunPosition/nodeName, maybe threadSpecName) when JSONPath evaluation fails.

I do not think VariableValueModel#jsonPath(String) should take workflow context directly; it is a low-level value helper. Better place is the runtime boundary/callers that have workflow context, e.g. ThreadRunModel.assignVariable(...) and VariableMutationModel.execute(...), catching LHVarSubError | RuntimeException and rethrowing LHVarSubError with safe context:

- wfRunId
- threadRunNumber
- nodeRunPosition / nodeName
- threadSpecName
- JSONPath
- assignment/mutation role

Do not include the JSON/customer value in the message.

I would treat this as a follow-up unless we want to expand this PR scope.

server/src/main/java/io/littlehorse/common/model/getable/core/variable/VariableValueModel.java
```diff
@@ -480,13 +485,17 @@ public VariableValueModel jsonPath(String path) throws LHVarSubError {
             return new VariableValueModel((Boolean) val);
         } else if (Double.class.isAssignableFrom(val.getClass())) {
             return new VariableValueModel((Double) val);
+        } else if (convertBigDecimal && val instanceof BigDecimal bigDecimal) {
```
Snarr: Thanks for the issue and the PR, this is an edge case we haven't seen yet but your analysis and fix look great.

From the looks of it, our jsonPath resolver dependency used under the hood by our `com.jayway.jsonpath` (`json-smart`) has the capability of returning `BigDecimal`s and `BigInteger`s, so I think we might need to account for both of those: https://github.com/netplex/json-smart-v2/blob/master/json-smart/src/main/java/net/minidev/json/parser/JSONParserBase.java#L246.

Could you add another `Else If` case and configuration for the `BigInteger` case as well? Not sure if that should be separate configurations or the same configuration, I'd err on the side of caution and make them separate experimental configurations.

eduwercamacaro: @Snarr Technically, it is also possible to get a `JSONObject` and `JSONArray` instance at this point.

eduwercamacaro:
> Operational follow-up: the new JSONPath error message avoids leaking customer JSON, which is good, but it still does not tell operators where the failure happened. At minimum we need wfRunId and step context (threadRunNumber, nodeRunPosition/nodeName, maybe threadSpecName) when JSONPath evaluation fails.
> 
> I do not think VariableValueModel#jsonPath(String) should take workflow context directly; it is a low-level value helper. Better place is the runtime boundary/callers that have workflow context, e.g. ThreadRunModel.assignVariable(...) and VariableMutationModel.execute(...), catching LHVarSubError | RuntimeException and rethrowing LHVarSubError with safe context:
> 
> * wfRunId
> * threadRunNumber
> * nodeRunPosition / nodeName
> * threadSpecName
> * JSONPath
> * assignment/mutation role
> 
> Do not include the JSON/customer value in the message.
> 
> I would treat this as a follow-up unless we want to expand this PR scope.


@jcustenborder I agree, this is a good change as we don't leak the customer json anymore.. 

I'm ok with a follow-up PR to do this refactor. This PR solves the issue you described, but we can improve the log and error message propagating this exception to the caller and then print more useful error messages.

eduwercamacaro (approved):
Thanks for the fix! Looks good to me :)

PR #2393, opened 2026-07-08
fix(dashboard): wire up Map/Array variable input paths

bryson-g:
The native Maps feature (#2368) completed the display direction (proto -> JSON) but left the input direction expecting raw proto-JSON, so Map variables either rendered no input field or crashed the page.

- Add getTypedVariableValueFromTypeDef: converts the friendly JSON the dashboard displays ({"k":v}) back into a typed proto Map/Array using the declared key/element types (recursive), plus a shared getVariableFilterValue helper.
- getVariableCaseFromTypeDef now handles inlineArrayDef/inlineMapDef, and getVariableDefType delegates to it so the two type-dispatch switches can't drift apart again (fixes the ExternalEventDef page crashing on a Map return type).
- The Execute WfRun form renders a validated textarea for Map/Array variables in the live dispatcher (VariableFormField) and converts the value on submit, instead of leaving the field unrendered and silently sending an empty container.
- The WfRun variable filter no longer throws during render (which was white-screening the whole page); the dialog validates the value and surfaces the error inline.

Adds unit tests covering map/array/nested conversion and malformed input.


Snarr (approved):
Thanks for this quick fix Bryson!

PR #2398, opened 2026-07-09
fix(server): pin authoritative element type on nested arrays and maps

Snarr:


eduwercamacaro (approved):

eduwercamacaro:
We can cherry-pick this into the 1.0 and 1.1 release branches so we solve this issue on these versions.

PR #2401, opened 2026-07-10
fix(dashboard): add clipboard copy fallback when API is unavailable

### LINKED ISSUE #2300: Dashboard Copy Button Broken
The "copy wfrunid" button and the "copy" button for the JSON_OBJ output popup are both not working.

bryson-g:
resolves #2300 

Use execCommand when navigator.clipboard is undefined so copy works in environments where the Clipboard API is blocked.


steinsleger (approved):

dashboard/src/app/utils/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from './copyToClipboard'
```
steinsleger, line 1: Is this being imported from the barrel anywhere?

bryson-g: No, I'll drop it 👍

PR #2409, opened 2026-07-14
fix(sdk-java): resolve InlineStruct StructDef type before Array or Map

Snarr:
This PR reorders the handling of Task Method result serialization so that `InlineStruct`s are serialized before `Array`s and `Maps.

## Background

Initially, `InlineStruct`s were being serialized after `Array`s and `Maps`, which led to a code path where `LHLibUtil.varValToObj` was called without a map of `placeholderValues` and the result serialization would fail.

## Further Work

Further work needs to be done to make `LHLibUtil` easier to read. For the sake of preventing breaking changes, there are a lot of methods added around `varValToObj` and `objToVarVal` that add Struct, Array, and Map support, but as time goes on we should work towards combining these methods again and reducing the complexity of this class. That is out of scope for this PR, as this is just a quick fix.

### DIFF
--- a/sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/ScheduledTaskExecutor.java
+++ b/sdk-java/src/main/java/io/littlehorse/sdk/worker/internal/ScheduledTaskExecutor.java
@@ -175,13 +175,25 @@ public class ScheduledTaskExecutor {
     }
 
     /**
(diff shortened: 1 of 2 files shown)


eduwercamacaro (approved):

eduwercamacaro:
@Snarr Can you please cherry-pick this into 1.2 after merging to master?

PR #2423, opened 2026-07-21
fix(dashboard): keep node selection and viewport across WfRun polling refreshes

### LINKED ISSUE #2421: Dashboard: WfRun polling clears graph node selection and details panel
## Context

On a running WfRun page, the Dashboard refreshes backend data approximately every three seconds. Each refresh rebuilds the workflow graph.

When a user selects a node to inspect its details, the next refresh clears the selection and returns the sidebar to "Select a node to view its details." This prevents sustained investigation of a NodeRun.

## Steps To Reproduce

1. Open LittleHorse Dashboard.
2. Navigate to a running WfRun containing NodeRuns.
3. Select a node in the workflow graph.
4. Inspect the node details in the sidebar.
5. Wait approximately three seconds.
6. Observe that the Dashboard refetches the WfRun, repaints the graph, unselects the node, and clears the details panel.

The issue occurs consistently.

## Expected Behavior

Background polling may continue updating the WfRun, but the selected node and details panel should remain open while that node still exists in the current graph and thread.

Refreshing runtime data should not reconstruct or reposition an unchanged graph. Updated diagnostics should be displayed without losing the user selection.

## Affected Components

Dashboard

## Environment

```text
OS: macOS 26.5.1
Client Version: Dashboard UI 1.2.0
Server Version: 1.2.0
Browser: Google Chrome 150.0.7871.127
Deployment: Internal/self-hosted
```

## Documentation

No documentation changes expected.

## Additional Context

Chrome DevTools MCP confirmed the behavior:

- Node selection appeared 51 ms after clicking.
- The next WfRun fetch began 2,494 ms later.
- The selected graph node was replaced and selection cleared at 2,568 ms.
- The main response transferred approximately 9.5 KB.
- Multiple requests were issued to the WfRun route during each refresh cycle.

Version history indicates this is likely a `v1.2.0` regression:

- Global `refreshInterval: 3000` polling was introduced on January 18, 2025 by [#1254](https://github.com/littlehorse-enterprises/littlehorse/pull/1254). This polling configuration is not recent.
- [`Diagram.tsx`](https://github.com/littlehorse-enterprises/littlehorse/blob/v1.2.0/dashboard/src/app/(authenticated)/%5BtenantId%5D/(diagram)/components/Diagram.tsx) changed on July 10, 2026 in [#2301](https://github.com/littlehorse-enterprises/littlehorse/pull/2301), five days before `v1.2.0`.
- Before that change, the graph-reset layout effect depended on the selected thread.
- The change added `wfRun` to the effect dependencies and added `setNode(undefined)` before reconstructing the nodes and edges.
- Each SWR response supplies a new `wfRun` object, causing the effect to run, clear the selected node, and rebuild the graph.
- The relevant change is absent from `v1.1.2` and present in `v1.2.0`.
- Neither `Diagram.tsx` nor the polling configuration has changed since `v1.2.0` on `master`.
- The `v1.1.2` comparison is source-level only; its UI was not run during this investigation.

The internal reproduction URL was omitted from this public report and can be shared privately if needed.

bryson-g:
Fixes #2421

The Diagram reset effect keyed on the wfRun/spec object identities, which change on every SWR poll that returns updated data. Each refresh therefore cleared the selected node, closed the details sidebar, and re-laid-out and re-positioned the graph while a WfRun was being inspected.

Implemented guarding the reset with the immutable identities (WfSpec id, WfRun id, thread name) so the graph is only rebuilt when the user actually navigates, and refresh nodeRun data in place on the laid-out nodes and the selected node so statuses and sidebar diagnostics keep updating between rebuilds. And now clear the selected node from context when the canvas selection empties, so the sidebar no longer relied on the periodic reset to close after deselecting or navigating.

### DIFF
--- a/dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Diagram.tsx
+++ b/dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Diagram.tsx
@@ -20,12 +20,19 @@ import { LHStatus, WfRun, WfSpec } from 'littlehorse-client/proto'
 import { PlayCircleIcon, RotateCcwIcon, StopCircleIcon } from 'lucide-react'
 import { usePathname, useSearchParams } from 'next/navigation'
 import { FC, ReactNode, useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react'
-import ReactFlow, { Controls, Node as RFNode, useEdgesState, useNodesState, type Viewport } from 'reactflow'
+import ReactFlow, {
+  Controls,
+  Node as RFNode,
+  useEdgesState,
+  useNodesState,
+  type OnSelectionChangeParams,
+  type Viewport,
+} from 'reactflow'
 import 'reactflow/dist/base.css'
 import { DiagramProvider, NodeInContext, ThreadType } from '../context'
 import edgeTypes from './EdgeTypes'
 import { extractEdges } from './EdgeTypes/extractEdges'
-import { LayoutManager } from './LayoutManager'
+import { getNodeRunsList, LayoutManager } from './LayoutManager'
 import nodeTypes from './NodeTypes'
 import { extractNodes, getCycleNodes, getNodeAfterEntrypoint } from './NodeTypes/extractNodes'
 import { Sidebar } from './Sidebar'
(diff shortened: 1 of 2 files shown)


coltmcnealy-lh (approved):

steinsleger (approved):

eduwercamacaro:
@bryson-g Thanks for fixing this. I cherry-picked this into 1.2 👍

Issue #2303, opened 2026-05-26
RecordTooLargeException silently drops state instead of failing the WfRun


clipinski:
### Context

When a workflow variable (or other getable) exceeds the configured `LHS_PRODUCER_MAX_REQUEST_SIZE` limit during `GetableManager.commit()`, the `BoundedBytesSerde` throws a `RecordTooLargeException` which is caught by `LHProcessingExceptionHandler` and silently swallowed. The workflow continues executing with stale or missing variable data and no error is surfaced to the user.

I believe this should instead fail the affected `WfRun` (transition it to `ERROR` status) so that the problem is visible and actionable, rather than allowing the workflow to continue in a corrupted state.

We discovered this in production when a workflow that passes large JSON payloads (serialized domain objects) through `STR` variables began silently producing incomplete results. The workflow appeared healthy (`RUNNING` status) but downstream tasks were receiving empty or stale variable values. The only indication was a `WARN`-level server log line: `"Record dropped: "`.

### Steps To Reproduce

1. Define a workflow with a `STR` variable that receives a large task output (e.g. a JSON-serialized list of domain objects, or a Base64-encoded file).
2. Run the workflow with enough data that the serialized `Variable` getable exceeds the `LHS_PRODUCER_MAX_REQUEST_SIZE` limit (default ~1MB).
3. The task itself completes successfully.
4. The variable assignment appears to succeed from the workflow's perspective — execution continues to the next node.
5. However, the variable's new value is never persisted. Downstream tasks that read the variable receive an empty or stale value.
6. The only indication is a `WARN`-level server log line: `"Record dropped: "` with a `RecordTooLargeException` stack trace.

### Expected Behavior

When a workflow variable exceeds the storeable size limit, the `WfRun` should transition to `ERROR` status with a clear failure message indicating which record exceeded the limit and its size. The server should remain stable (no crash).

### Affected Components

Kernel (/server)

### Environment

- **Server Version:** `1.0.1`
- **Java:** Corretto 21
- **OS:** Linux (containerized deployment)
- **Configuration:** `LHS_PRODUCER_MAX_REQUEST_SIZE` at default `1047000` (we have temporarily raised ours to ~5MB as a workaround)

### Root cause analysis

I traced through the code on the latest `master` branch and it appears the issue is in how `RecordTooLargeException` propagates through the command processing pipeline.

**The call chain:**

When a task result comes back (e.g. `ReportTaskRun`), the processing flows through:

[...]

PR #2429, opened 2026-07-22
feat(server): Send failed ReportTaskRun when these subcommands fail at the command producer level

eduwercamacaro:
Kafka producers might throw `RecordTooLarge` exceptions when sending batch records. This is the case to the server's command producers; sometimes users send commands that exceed the maximum request size configured via `LHServerConfig.PRODUCER_MAX_REQUEST_SIZE gRPC. The gRPC listeners transform these exceptions into internal errors, and they get propagated to the clients. 

Previously the exception was turned into an internal error and propagated to the client with no actionable meaning. Worse, for large commands like `ReportTaskRun` with a big task output, the command was never processed, so the WfRun never received its task result and hung until the task node handled the timeout command.

This PR only addresses the case when the command producers catch the RecordTooLarge exception. I'll create separate PRs to catch these errors in other places where these exceptions can be thrown. 

Solves one of the problem described in this issue #2303

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/ReportTaskRunModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/ReportTaskRunModel.java
@@ -13,6 +13,7 @@ import io.littlehorse.common.model.getable.core.taskrun.TaskRunModel;
 import io.littlehorse.common.model.getable.core.variable.VariableValueModel;
 import io.littlehorse.common.model.getable.objectId.TaskRunIdModel;
 import io.littlehorse.common.util.LHUtil;
+import io.littlehorse.sdk.common.proto.LHErrorType;
 import io.littlehorse.sdk.common.proto.ReportTaskRun;
 import io.littlehorse.sdk.common.proto.TaskStatus;
 import io.littlehorse.server.streams.topology.core.CoreProcessorContext;
@@ -102,6 +103,18 @@ public class ReportTaskRunModel extends CoreSubCommand<ReportTaskRun> {
     public static ReportTaskRunModel fromProto(ReportTaskRun proto, ExecutionContext context) {
         ReportTaskRunModel out = new ReportTaskRunModel();
         out.initFrom(proto, context);
+        return out;
+    }
+
+    public ReportTaskRunModel toErrorReport(LHApiException apiException) {
+        ReportTaskRunModel out = new ReportTaskRunModel();
+        out.taskRunId = this.taskRunId;
+        out.time = this.time;
+        out.status = TaskStatus.TASK_OUTPUT_SERDE_ERROR;
+        out.attemptNumber = this.attemptNumber;
+        out.error = new LHTaskErrorModel(apiException.getMessage(), LHErrorType.TASK_ERROR);
+        out.totalCheckpoints = this.totalCheckpoints;
+        out.logOutput = new VariableValueModel(apiException.getMessage());
         return out;
     }
 }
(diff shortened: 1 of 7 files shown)


server/src/main/java/io/littlehorse/common/model/corecommand/subcommand/ReportTaskRunModel.java
```diff
@@ -104,4 +105,16 @@ public static ReportTaskRunModel fromProto(ReportTaskRun proto, ExecutionContext
         out.initFrom(proto, context);
         return out;
     }
+
+    public ReportTaskRunModel toErrorReport(LHApiException apiException) {
+        ReportTaskRunModel out = new ReportTaskRunModel();
+        out.taskRunId = this.taskRunId;
```
coltmcnealy-lh, line 111: Oh man, no setters? I like this, it's like 22-year-old Colt Code

server/src/main/java/io/littlehorse/common/util/LHProducer.java
```diff
@@ -44,7 +47,7 @@ public CompletableFuture<RecordMetadata> send(String key, AbstractCommand<?> t,
      */
     private CompletableFuture<RecordMetadata> sendRecord(ProducerRecord<String, Bytes> record) {
         CompletableFutureCallback out = new CompletableFutureCallback();
-        CompletableFuture.runAsync(() -> prod.send(record, out));
+        prod.send(record, out);
```
coltmcnealy-lh, line 50: I think this is okay because the `Producer#send()` is async anyways, right? (`linger.ms`)

server/src/main/java/io/littlehorse/server/LHServerListener.java
```diff
@@ -740,7 +743,27 @@ public void reportTask(ReportTaskRun req, StreamObserver<Empty> ctx) {
         ReportTaskRunModel reqModel = LHSerializable.fromProto(req, ReportTaskRunModel.class, requestContext);
         TenantIdModel tenantId = requestContext.authorization().tenantId();
         PrincipalIdModel principalId = requestContext.authorization().principalId();
-        commandSender.reportTaskAndDontWaitForResponse(reqModel, ctx, principalId, tenantId);
+        CompletableFuture<RecordMetadata> futureResponse =
+                commandSender.reportTaskAndDontWaitForResponse(reqModel, principalId, tenantId);
+        try {
+            waitForProcessing(futureResponse, Optional.empty());
```
coltmcnealy-lh, line 749: Can we "defer" this to a background thread so that it doesn't block the `ReportTask` rpc? That way we don't slow down the throughput of the task worker.

Unless you want the task worker to synchronously receive an error?

coltmcnealy-lh (approved):

server/src/main/java/io/littlehorse/common/util/LHProducer.java
eduwercamacaro: Yes, this was needed because there was a bug on Java 21 on VirtualThreads with synchronized methods. 
It is now solved on java 25

server/src/main/java/io/littlehorse/server/LHServerListener.java
eduwercamacaro: Worker clients don't wait for this command to report in order to continue executing more tasks, so this shouldn't slow down throughput.

eduwercamacaro: The worker sdk uses a callback to handle retries

PR #2435, opened 2026-07-22
feat(server): Automatically retry failed ReportTaskRun commands

eduwercamacaro:
LH server doesn't surface API errors when processing ReportTaskRun commands, that causes unexpected behaviors on WfRun that are waiting on a task node.

This modifies that behavior by automatically retrying a ReportTaskRun when the streams topology fails to process the command (i.e RecordTooLargeException).

### DIFF
--- a/server/src/test/java/e2e/VarSerdeTest.java
+++ b/server/src/test/java/e2e/VarSerdeTest.java
@@ -8,9 +8,11 @@ import io.littlehorse.sdk.common.proto.TaskStatus;
 import io.littlehorse.sdk.common.proto.VariableValue;
 import io.littlehorse.test.LHTest;
 import java.util.Date;
+import org.junit.jupiter.api.Disabled;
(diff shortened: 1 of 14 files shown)


server/src/test/java/e2e/VarSerdeTest.java
```diff
@@ -8,9 +8,11 @@
 import io.littlehorse.sdk.common.proto.VariableValue;
 import io.littlehorse.test.LHTest;
 import java.util.Date;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 @LHTest
+@Disabled("These tests are silently failing because the TaskRun is not being created in the arrange phase.")
```
eduwercamacaro, line 15: @Snarr @l4crito I disabled this test until we find a proper way to test this.

coltmcnealy-lh (approved):
I like this version more. Good work!

PR #2439, opened 2026-07-28
fix(server): Avoid sending output topic records when the output topic is missing 

eduwercamacaro:
Littlehorse server has the ability to send events to an output topic, so analytics users can consume these events and perform aggregations over these events (See proposal `001-output-topic.md`). This behavior is controlled by a configuration at the tenant level. When a tenant is configured to send events to the output topic, the Kafka Streams processor hopes that the topic was already created by Littlehorse server operators.

On local development, Littlehorse server automatically creates these topics, making it easy to test the output topic feature locally. However, this behavior is not recommended for production.

In production environments, operators must create the output topics before enabling the tenant-level configuration to send events. 

A missing output topic is catastrophic for the entire LHCluster, causing transaction timeouts and poison pill scenarios.

This PR verifies if the output topic actually exists on the kafka cluster before attempting to produce a record. The proposal includes an asynchronous call to the brokers using the kafka Admin API, the idea is to not block the Stream processors with a time-consuming operation.

### DIFF
--- a/server/src/main/java/io/littlehorse/common/LHServerConfig.java
+++ b/server/src/main/java/io/littlehorse/common/LHServerConfig.java
@@ -40,9 +40,12 @@ import org.apache.commons.lang3.tuple.Pair;
 import org.apache.kafka.clients.CommonClientConfigs;
 import org.apache.kafka.clients.admin.Admin;
 import org.apache.kafka.clients.admin.AdminClientConfig;
+import org.apache.kafka.clients.admin.DescribeTopicsResult;
 import org.apache.kafka.clients.admin.NewTopic;
+import org.apache.kafka.clients.admin.TopicDescription;
 import org.apache.kafka.clients.consumer.ConsumerConfig;
 import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.common.KafkaFuture;
 import org.apache.kafka.common.config.TopicConfig;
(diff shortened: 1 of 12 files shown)


server/src/main/java/io/littlehorse/common/LHServerConfig.java
```diff
@@ -1160,6 +1163,14 @@ public boolean createKafkaTopic(NewTopic topic) throws InterruptedException, Exe
         }
     }
 
+    public Map<String, KafkaFuture<TopicDescription>> outputTopicExistsFor(TenantIdModel tenant) {
```
Snarr, line 1166: Do we need to include the executionOutputTopic too? I see `getMetadataOutputTopicName`, `getCoreCmdTopicName`, but no `getExecutionOutputTopicName` call.

coltmcnealy-lh (commented):
>This PR verifies if the output topic actually exists on the kafka cluster before attempting to produce a record. The proposal includes an asynchronous call to the brokers using the kafka Admin API, the idea is to not block the Stream processors with a time-consuming operation.

The problem here is that the Stream Thread does block when waiting for the futures. 

We should refactor this to check at most once for each output topic. I understand the vulnerability that if the output topic is deleted we end up with a problem; however, that's a less-common case than when the tenant is created with output topic enabled before the topic exists.

Issue #2441, opened 2026-07-29
Support `map.put(key, value)` on Native Maps


Snarr:
## Background

1. We recently added native LittleHorse Maps to the server, along with the `EXTEND` operation for merging two native Maps.
2. There is no first-class way to insert or update a single key/value pair in a native Map. The complementary removal operation already exists (`REMOVE_KEY`, `VariableMutationType` #8), leaving the Map mutation surface asymmetric.
3.  LittleHorse cannot yet construct a Map with dynamic (runtime-resolved) keys in a WfSpec. Structs have a dynamic constructor (`StructBuilder` / `InlineStructBuilder`, exposed as a `VariableAssignment` source at `common_wfspec.proto:94`), but Maps have no analog.

## Proposal

- Do not add a new `VariableMutationType`, and do not modify the `VariableMutation` message.
- Instead, add a `MapBuilder` as a `VariableAssignment` source
- Implement `map.put(key, value)` as SDK sugar that compiles to an `EXTEND` mutation whose RHS is a single-entry `MapBuilder`.

### Why not extend `VariableMutation` with a new two-operand `PUT` feature?

An earlier idea I had involved adding `PUT` to `VariableMutationType` plus an optional `key` field on `VariableMutation`. That was rejected because:

- `VariableMutation` (`common_wfspec.proto:167`) is deliberately single-operand: `LHS = LHS <op> RHS`, with one `rhs_value` oneof. Adding a second operand field that is meaningful for exactly one operation pollutes a core, widely-used message.
- It would duplicate merge logic that `extend`'s MAP branch already implements correctly.
- It would leave the dynamic-Map-construction gap unsolved.

### Protobuf

Add a `MapBuilder` message and wire it into the `VariableAssignment` `source` oneof in `schemas/littlehorse/common_wfspec.proto`.

```proto
message VariableAssignment {
  // ... existing ...
  oneof source {
    string variable_name = 2;
    VariableValue literal_value = 3;
    FormatString format_string = 4;
    NodeOutputReference node_output = 5;
    Expression expression = 6;
    StructBuilder struct_builder = 9;
    SizeOf size_of = 10;

    // Builds a native Map from values available in the ThreadRun.
    MapBuilder map_builder = 11;
  }
}

// Builds a native Map using data available in the context of this ThreadRun.
// This is the Map analog of StructBuilder, and permits dynamic (runtime-resolved)
// keys and values, which a literal `Map` cannot express.
message MapBuilder {
[...]

PR #2442, opened 2026-07-30
feat(proto): add MapBuilder support for inline map construction

Snarr:
This PR adds MapBuilder to the proto so we can construct Maps dynamically within `WfSpec`s.

This fulfills acceptance criterions 1 and 2 for issue #2441.


server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/variable/VariableAssignmentModel.java
```diff
@@ -226,6 +238,14 @@ public boolean canBeType(TypeDefinitionModel type, ThreadSpecModel tspec) {
             case SIZE_OF:
                 baseType = new TypeDefinitionModel(VariableType.INT);
                 break;
+            case MAP_BUILDER:
+                if (mapBuilder.getMapType() != null) {
+                    baseType = new TypeDefinitionModel(mapBuilder.getMapType());
+                } else {
+                    // unknowable without manager/wfSpec context
+                    return true;
```
eduwercamacaro, line 248: If we can't infer the Map type, wouldn't this line cause cast exceptions at runtime?

eduwercamacaro (approved):

PR #2443, opened 2026-07-30
feat(sdk-java): add Map `put` and `LHMapBuilder` semantics

### LINKED ISSUE #2441: Support `map.put(key, value)` on Native Maps

Snarr:
This PR rounds out the Map functionality in the `sdk-java` `WfSpec` DSL:
- Adds `map.put(key,value)` functionality to the WfSpec DSL
- Adds `LHMapBuilder` for dynamically generating Maps in WfSpecs
- Adds `lhs_lh_path` to `VariableMutation`s to support the `put` functionality under the hood

Resolves acceptance criteria 3 and 4 of issue #2441.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


examples/java/maps/src/main/java/io/littlehorse/examples/MapExample.java
```diff
@@ -16,21 +17,37 @@ public class MapExample {
...
+        return new WorkflowImpl("reserve-inventory", wf -> {
+            WfRunVariable inventory = wf.declareMap("inventory", String.class, Integer.class)
+                    .withDefault(Map.of("apples", 3, "bananas", 5, "cherries", 12));
+            WfRunVariable sku = wf.declareStr("sku").withDefault("apples");
+            WfRunVariable quantity = wf.declareInt("quantity").withDefault(2);
+            WfRunVariable available = wf.declareInt("available");
+            WfRunVariable remaining = wf.declareInt("remaining");
 
-            wf.execute("consume-map", mapVar);
+            available.assign(wf.execute("check-availability", inventory, sku));
```
eduwercamacaro: proposal:
Can we replace the task worker with a `inventory.get(sku)`? Not sure if we support that yet.

proposals/023-native-maps.md
```diff
@@ -183,6 +225,39 @@ wf.execute("report", wordCounts.get("hello"));
...
+// Insert/overwrite with a literal key.
+wordCounts.put("hello", 42);
+
+// Insert/overwrite with a variable key.
+wordCounts.put(theKey, 100);
+```
+
+`put` on a key that already exists replaces the previous value (consistent with `EXTEND` semantics).
+
+### Building Maps Inline (`buildMap` / `LHMapBuilder`)
```
eduwercamacaro, line 294: This section is mainly an implementation detail. I think we can remove it.

eduwercamacaro (approved):

coltmcnealy-lh (commented):
Question: right now, we use the `ASSIGN` operation with a `lhsJsonPath`. Should we continue to use `ASSIGN` but change it to use an `LHPath`, rather than use `EXTEND`? When I wrote `ASSIGN` and `EXTEND` it felt like Extend was just for concatenating strings & arrays.

Snarr:
> Question: right now, we use the `ASSIGN` operation with a `lhsJsonPath`. Should we continue to use `ASSIGN` but change it to use an `LHPath`, rather than use `EXTEND`? When I wrote `ASSIGN` and `EXTEND` it felt like Extend was just for concatenating strings & arrays.

Can you produce an idea of what your DSL version would look like? I prefer EXTEND in this case, but I want to understand your proposal more

examples/java/maps/src/main/java/io/littlehorse/examples/MapExample.java
Snarr: Added support for that now :)

bryson-g (approved):

proposals/023-native-maps.md
```diff
@@ -148,6 +152,83 @@ Every native `Map` is always typed. A `Map` may never exist without a known key/
...
+### `LHPath` and `VariableMutation` Changes
+
+To update a Map entry selected at runtime, such as `wordCounts.put(theKey, value)`, we will complete the typed mutation path design from [Proposal 011](./011-type-safe-structpaths.md) by adding an `LHPath` option for the left-hand side of a `VariableMutation`. The deprecated JSONPath field remains for wire compatibility:
+
+```proto
+message VariableMutation {
+  // ... existing fields ...
+
+  oneof lhs_path {
+    // DEPRECATED: use lhs_lh_path.
```
coltmcnealy-lh, line 203: Love to see this. Thanks for continuing to undo one of the big mistakes I made in the first draft of LH

proposals/023-native-maps.md
```diff
@@ -189,6 +270,47 @@ wf.execute("report", wordCounts.get("hello"));
...
+wordCounts.put(theKey, 100);
+```
+
+### Building Maps Inline (`buildMap` / `LHMapBuilder`)
+
+`WorkflowThread` gains a `buildMap()` factory that returns an `LHMapBuilder`:
+
+```java
+// Construct a Map<STR, INT> at runtime from a mix of literals and variables.
+WfRunVariable wordCounts = wf.declareMap("word-counts", String.class, Integer.class);
```
coltmcnealy-lh, line 300: How would I declare a nested map? Is that possible, or would I require a struct to do that?

Additionally, how do I put a Map inside a Struct? 
```
@LHStruct("obiwan")
public Class ObiWan{
  public String lightsaber;
  public Map<String, Integer> planetsToTimesVisitedOnEachPlanet;
}
```

proposals/023-native-maps.md
```diff
@@ -239,15 +361,18 @@ This keeps the wildcard/untyped representation out of the engine entirely for al
 
 The following `VariableMutationType` operations are supported on `Map` values:
 
-- **`ASSIGN`**: replace the entire map.
+- **`ASSIGN`**: replace the entire Map when no LHS path is present, or assign one nested value when `lhs_lh_path` is present.
```
coltmcnealy-lh, line 364: Better, thanks!

coltmcnealy-lh (approved):

PR #2445, opened 2026-07-30
fix(dashboard): send optional variables when executing a wfspec

bryson-g:
The Execute WfRun form dropped every optional variable, so workflows started with them NULL regardless of what the user typed in, while `lhctl run` passed the same values fine. `WfRunForm` read `formState.dirtyFields` only inside its submit handler, and react-hook-form only tracks a slice of `formState` once it has been read during render, so `dirtyFields` was always empty and `ExecuteWorkflowRun` filtered out every non-required variable. Two older bugs in the same form became reachable once optional variables actually reached the server, each fixed in its own commit: `Textarea` never attached its forwarded ref, keeping Map, Array and JSON values out of the submitted values, and DOUBLE fields had no `step`, so any decimal failed constraint validation and silently blocked submit. Verified against a local cluster across every primitive type plus Map and JSON; 180 tests pass, and each of the three fixes has a test confirmed to fail when that fix alone is reverted.


Snarr (approved):
Ran and validated the fix on my end 👍

PR #2447, opened 2026-07-31
fix(dashboard): prevent sidebar crash from stale attempt/noderun index

bryson-g:
Fixes the sidebar crash on completed WfRuns: open a TaskRun with multiple attempts, switch to a later attempt, then click a different node and the sidebar dies with "Something went wrong".

The attempt index lives in `TaskNodeRun`, which isn't keyed by node, so React reuses the instance and the old index carries into a TaskRun with fewer attempts. `nodeRunIndex` in `Sidebar` has the same issue when you jump between nodes with different noderun counts.

Added key `TaskNodeRun` by taskGuid so it remounts, reset `nodeRunIndex` when the selected node changes, and guard the three list reads (`Attempts`, `NodeRunComponent`, `Failures`) so a stale index returns null instead of white screening.


coltmcnealy-lh (approved):

coltmcnealy-lh (approved):

coltmcnealy-lh:
@eduwercamacaro once this is merged, we should cherry-pick it into the `1.2` branch.

PR #2448, opened 2026-08-03
fix(server): evaluate edge compare expressions against buffered mutations

### LINKED ISSUE #2181: Server Performs Variable Mutations in Inconsistent Order
### Context

When assigning a variable and using the assigned value in another assignment operation on the same edge, the server may process the two variable mutations in an inconsistent order, leading to unexpected mutations and or variable mutation errors.

This is the equivalent of running the following code and it producing an error because `my_array` is null when trying to use it with the `in` operand:

```python
def return_arr():
  return [1,2,3]

def main():
  my_array = return_arr()
  found = 2 in my_array

main()
```

### Steps To Reproduce

1. Declare the following `WfSpec`:

```java
WfRunVariable arrVar = wf.declareJsonArr("my-array");
WfRunVariable found = wf.declareBool("found");
TaskNodeOutput produced = wf.execute("produce-array");
arrVar.assign(produced);
found.assign(arrVar.doesContain(2));
```

Where the `produce-array` task produces an Array of Integers `[1, 2, 3]`

2. Execute a WfRun

3. Observe the Variable Mutation error

> "Failed evaluating edge with sink node 2-exit-EXIT: Mutating variable found with operation ASSIGN: Failed evaluating CONTAINS"

### Expected Behavior

In the above `WfSpec` example, the mutation `arrVar = produced` should happen before the mutation `found = 2 in arrVar`, meaning that the `WfRun` should not error and the `found` variable should be set to `true`.

### Affected Components

_No response_

### Environment

OS: macOS 15.1.1
Server Version: 1.0.0

### Documentation

_No response_

### Additional Context

_No response_

Snarr:
Resolves #2181

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/getable/core/wfrun/ThreadRunModel.java
@@ -911,7 +911,7 @@ public class ThreadRunModel extends LHSerializable<ThreadRun> {
                 break;
             case EXPRESSION:
                 ExpressionModel expression = assn.getExpression();
-                val = expression.evaluate(this, varAssn -> assignVariable(varAssn, txnCache));
+                val = expression.evaluate(varAssn -> assignVariable(varAssn, txnCache));
                 break;
             case SIZE_OF:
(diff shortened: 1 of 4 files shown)


coltmcnealy-lh (approved):
Nice!

PR #2454, opened 2026-08-06
fix(server): bump StructDef version when description changes

Snarr:
This PR bumps the StructDef version when the description changes, making StructDef descriptions behave more similarly to other schema registries.

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutStructDefRequestModel.java
+++ b/server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutStructDefRequestModel.java
@@ -86,19 +86,17 @@ public class PutStructDefRequestModel extends MetadataSubCommand<PutStructDefReq
         if (latestVersion == null) {
             spec.setId(new StructDefIdModel(name, 0));
         } else {
-            if (InlineStructDefUtil.equals(spec.getStructDef(), latestVersion.getStructDef())) {
-                boolean descriptionChanged = !Objects.equals(description, latestVersion.getDescription());
+            boolean schemaChanged = !InlineStructDefUtil.equals(spec.getStructDef(), latestVersion.getStructDef());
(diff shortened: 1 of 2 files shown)


server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutStructDefRequestModel.java
```diff
@@ -86,19 +86,17 @@ public Message process(MetadataProcessorContext context) {
         if (latestVersion == null) {
             spec.setId(new StructDefIdModel(name, 0));
         } else {
-            if (InlineStructDefUtil.equals(spec.getStructDef(), latestVersion.getStructDef())) {
-                boolean descriptionChanged = !Objects.equals(description, latestVersion.getDescription());
+            boolean schemaChanged = !InlineStructDefUtil.equals(spec.getStructDef(), latestVersion.getStructDef());
+            boolean descriptionChanged = !Objects.equals(description, latestVersion.getDescription());
```
eduwercamacaro, line 90: nit:
I'm thinking that we should probably trim whitespaces from the beginning or the end of the `description` field before storing it so we reduce users making accidental changes on the description. 

A counter-argument to that would be to store the exact same value that users write into the schema. 

Any thoughts? We can ignore this if you want

eduwercamacaro (approved):

server/src/main/java/io/littlehorse/common/model/metadatacommand/subcommand/PutStructDefRequestModel.java
Snarr: @eduwercamacaro This is a good point. I'd err on the side of caution by not trimming the user's description of any whitespace. We can add that in the future if it causes user confusion, but for now I want to fully honor the user's provided description and not provide any unexpected mutations to it.

PR #2456, opened 2026-08-07
feat(ci): generate GitHub release notes from release_notes/ instead of git-cliff

eduwercamacaro:
Generate release notes by copying the release_notes files for each release.


mijailr (approved):
Looks good, I guess you can remove all references to git-cliff now that you have a different approach.

eduwercamacaro:
> Looks good, I guess you can remove all references to git-cliff now that you have a different approach.

I will push a new PR to remove git-cliff dependency from these pipelines.

PR #2461, opened 2026-08-12
feat(server): add SearchWorkflowMigrationPlan RPC and lhctl support

JakeRose03:
Adds an EXPERIMENTAL `SearchWorkflowMigrationPlan` gRPC endpoint to list `WorkflowMigrationPlan`s, with `prefix`, `limit`, and `bookmark`.

Assisted-by: Claude Opus 4.8


schemas/littlehorse/service.proto
```diff
@@ -1151,6 +1154,28 @@ message ExternalEventDefIdList {
   optional bytes bookmark = 2;
 }
 
+// EXPERIMENTAL: Search for WorkflowMigrationPlan's.
+message SearchWorkflowMigrationPlanRequest {
```
coltmcnealy-lh, line 1160: Suggestion: we should have an optional parameter that specifies the `wf_spec_name` that the `WorkflowMigrationPlan` refers to.

coltmcnealy-lh (approved):

PR #2469, opened 2026-08-16
chore(proposal): update and fix quotas proposal

coltmcnealy-lh:


coltmcnealy-lh:
This is the prompt I used:

> The quotas proposal ( @proposals/020-quotas.md) is a bit off. Update it so that:

> the protobuf for the metrics is fixed.
> - Should matchwhat's in @proposals/017-workflow-metrics.md and in @schemas/littlehorse/service.proto and @schemas/littlehorse/metrics.proto 

> the Quota resource actually is cluster-scoped, not tenant-scoped
> - the quotaId has a tenant id, so one quota affects only one tenant (this matches the proposal)
> - The reason why we have it as cluster-scoped is because of ACL's. In some multi-tenant deployments, cluster admins want to give users of a tenant full access (ALL_ACTIONS ACL_ALL_RESOURCES) to the tenant. 
> - If the Quota belonged to the Tenant not the cluster-scope, then the users inside a tenant could alter their own quotas
> - The current model is such that you need root (global_acls) access to create quotas, tenants, and principals
> - The part where the quota usage metrics show up as getables shown inside the tenant is actually correct

Snarr (approved):

PR #2472, opened 2026-08-17
feat(sdk-java): allow base64-encoded certificates to be passed as env vars

coltmcnealy-lh:
- Introduces LHC_CLIENT_CERT_VALUE, LHC_CA_CERT_VALUE, and LHC_CLIENT_KEY_VALUE as new configurations for LHConfig in java
- Old file-based configurations still take precedence

### DIFF
--- a/sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
+++ b/sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
@@ -22,11 +22,14 @@ import io.littlehorse.sdk.common.proto.LittleHorseGrpc.LittleHorseStub;
 import io.littlehorse.sdk.common.proto.TaskDef;
 import io.littlehorse.sdk.common.proto.TaskDefId;
 import io.littlehorse.sdk.common.proto.TenantId;
+import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


sdk-java/src/main/java/io/littlehorse/sdk/common/config/LHConfig.java
```diff
@@ -511,6 +525,28 @@ private Channel getChannel(String host, int port) {
         return out;
     }
 
+    private byte[] getCertificateBytes(String fileKey, String valueKey) {
```
eduwercamacaro, line 528: The values for these arguments are always the constant that is available at this class level. We can use them instead of receiving arguments.

Snarr: @eduwercamacaro the keys are constant but there are 3 call sites where we call `getCertificateBytes()` receives different key/value pairs.

```java
private Channel getChannel(String host, int port) {
  // ...
  byte[] caCert = getCertificateBytes(CA_CERT_KEY, CA_CERT_VALUE_KEY);
  // ...
  byte[] clientCert = getCertificateBytes(CLIENT_CERT_KEY, CLIENT_CERT_VALUE_KEY);
  byte[] clientKey = getCertificateBytes(CLIENT_KEY_KEY, CLIENT_KEY_VALUE_KEY);
  // ...
}
```

Am I understanding this right? Maybe we can make some methods that codify the named pairs:

```java
private byte[] getCaCertBytes() {
    return getCertificateBytes(CA_CERT_KEY, CA_CERT_VALUE_KEY);
}

private byte[] getClientCertBytes() {
    return getCertificateBytes(CLIENT_CERT_KEY, CLIENT_CERT_VALUE_KEY);
}

private byte[] getClientKeyBytes() {
    return getCertificateBytes(CLIENT_KEY_KEY, CLIENT_KEY_VALUE_KEY);
}
```

Snarr: @eduwercamacaro I guess we could also set these constants on LHConfig instantiation?

PR #2473, opened 2026-08-18
perf(server): warm RocksDB block cache before forwarding bulk WfRun deletes

eduwercamacaro:
`BulkDeleteWfRunModel.process` forwards multiple `InternalDeleteWfRunRequestModel` when processing bulk wfrun deletes.


coltmcnealy-lh (approved):

eduwercamacaro:
@coltmcnealy-lh do we want to cherry-pick this into 1.2?

PR #2481, opened 2026-08-19
fix(dashboard): end the keycloak session when signing out

bryson-g:
Sign out now redirects the browser through Keycloak's end-session endpoint so Keycloak clears its own SSO cookies, instead of calling that endpoint from the server where it cannot.

Adds an `/api/auth/federated-logout` route that builds the logout URL from the session's ID token, and tests for it.


jhosep-marin (approved):

mijailr:
Think about this for a second, when you sign in with an OIDC provider (Google) in an application, and you click logout, does it close your Google session? 

What I've seen is no, it won't close your OIDC provider session, just the application session. So I believe we shouldn't close the OIDC provider session.

mijailr (changes requested):
I would like to find an alternative that doesn't close the OIDC session.

mijailr:
OpenID specs says there are 4 options: https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
 
- none: default
- login: ask to login again
- consent: ask for consent
- select_account: allows to select account

I'm not sure if Keycloak supports all, but you can pass the parameter as https://next-auth.js.org/getting-started/client#additional-parameters

Using `{prompt: "login"}` or `{prompt: "consent"}` may fix it.

dashboard/src/app/api/auth/signin/LoginButton.tsx

```diff
'use client'
import { ClientSafeProvider, signIn } from 'next-auth/react'
import { useSearchParams } from 'next/navigation'
import { FC } from 'react'

export const LoginButton: FC<Pick<ClientSafeProvider, 'id' | 'name'>> = ({ id, name }) => {
  const searchParams = useSearchParams()
  const callbackUrl = searchParams.get('callbackUrl') || '/'
  return (
    <button
-      onClick={() => signIn(id, { callbackUrl })}
+     onClick={() => signIn(id, { callbackUrl, prompt: "login" })}
      className="w-full rounded-md bg-blue-500 px-4 py-4 font-semibold text-white md:w-96 lg:w-96"
    >
      Login with {name}
    </button>
  )
}
```

mijailr (approved):

PR #2485, opened 2026-08-20
feat: add StructFieldDef descriptions

### LINKED ISSUE #2486: Add Descriptions to StructFieldDefs
This issue will track the progress of implementing Proposal 025 "StructFieldDef Descriptions". 

For more information about the feature and the requirements for the implementation, see the proposal [here](https://github.com/littlehorse-enterprises/littlehorse/blob/master/proposals/025-structfielddef-descriptions.md).

Snarr:
This PR adds StructFieldDef descriptions to LittleHorse based on Proposal 25. 

## Purpose

StructFieldDef descriptions allow you to add an optional String description to a StructFieldDef so its usage can be documented. This feature will help both humans and agents better understand your StructDef.

## Immutability

StructDef versions are still immutable so successfully issuing a description update to a StructDef will bump the version.

## Backwards Compatibility

I manually validated that old StructDefs created before fields descriptions were added to the proto are equivalent to new StructDefs without field descriptions after field descriptions were added to the proto. This means users existing StructDefs won't bump a version upon upgrading to this change if they don't add StructFieldDef descriptions.

## Dashboard

This PR adds
- StructFieldDef descriptions to the dashboard, specifically on the dedicated StructDef pages and also in the "Execute WfRun" form. 
- It also adds StructDef descriptions to the "Execute WfRun" form, as they were missing
- It also truncates long StructDef descriptions and StructFieldDef descriptions but allows users to open a modal to reveal them fully

**Execute WfRun Form Screenshot**:
image

**StructDef page Screenshot**:
image

## Examples

This PR updates the Java, .NET, Go, and Python examples to use StructFieldDef descriptions and also StructDef descriptions.


dashboard/src/components/ui/field.tsx
```diff
@@ -11,7 +11,7 @@ function FieldGroup({ className, ...props }: React.ComponentProps<'div'>) {
     <div
       data-slot="field-group"
       className={cn(
-        'group/field-group @container/field-group flex w-full flex-col gap-7 data-[slot=checkbox-group]:gap-3 [&>[data-slot=field-group]]:gap-4',
+        'group/field-group @container/field-group flex w-full flex-col gap-2 data-[slot=checkbox-group]:gap-3 [&>[data-slot=field-group]]:gap-4',
```
bryson-g: shadcn components should not be edited due to regressions that may occur throughout the whole project

coltmcnealy-lh (approved):

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Forms/components/StructDefGroup.tsx
Comment (bryson-g) [inline dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Forms/components/StructDefGroup.tsx]: the `description` field is not currently being passed to fields which are struct-typed. 

however, the description of a struct-typed field and the description of a struct itself seems redundant in this type of scenario. how do you think we should handle this?

screenshots for reference:

execute wfrun form:
Image

structdef metadata page:
Image

dashboard/pnpm-lock.yaml
bryson-g, line 1: all these changes in this lock file appear to be whitespace differences. checkout main's and we should probably add this file to `.prettierignore`.

dashboard/src/components/ui/field.tsx
bryson-g: pass the override through the caller instead of at the source, so it affects the specific component you intend.

bryson-g (dismissed):

dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Forms/components/StructDefGroup.tsx
Comment (Snarr) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Forms/components/StructDefGroup.tsx]: This is a great point, thanks Bryson. 

In written programming languages, the equivalent would be a class comment versus a field comment. Usually the class comment would not appear where the field is, you'd have to go into the class to see it.

But, in this case, since we are not limited by the semantics of a written programming language and can design the dashboard however we want, I think it would be best if we showed both descriptions.

Maybe with labels where one says "StructDef description" and another says "FieldDef Description".

What do you think?

Comment (bryson-g) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Forms/components/StructDefGroup.tsx]: it seems a bit inconsistent to have the explicit named label on just the struct-typed field and not the rest of the fields. 

if we don't want the named labels, we could just have a new line so the two descriptions read one after the other, separated by a new line break. I think I prefer this approach than explicitly named labels taking up real estate. 

I'm a proponent of the approach of intuitive readability of the dashboard rather than excessive labeling and over explanation of every data value everywhere.

Comment (Snarr) [reply dashboard/src/app/(authenticated)/[tenantId]/(diagram)/components/Forms/components/StructDefGroup.tsx]: I had an idea in mind a bit different than named labels, where we make the field name and description appear above the Struct input div and include the StructDef description itself inside the div like where it currently is. This will introduce a clear separation between the field metadata and the typedef metadata. I didn't know how to verbalize this until now, but what do you think of this alternative?

PR #2495, opened 2026-08-25
feat(server): Background thread for local partitions as replacement for punctuators

eduwercamacaro:
This PR introduces a new design for background operations that happen inside a punctuator.


coltmcnealy-lh (approved):
After a lot of playing with this, I think it is ready for prime time.

PR #2500, opened 2026-09-01
sdk-js parity via a cross-SDK conformance suite

bryson-g:
## What this is

The 1.3 sdk-js parity deliverable, plus the mechanism that proves the parity claim instead of asserting it.

1. sdk-js is now a complete SDK. It has the full workflow builder, unified serde, OAuth, task workers, user task helpers, and a JS-native options style (`declareStr('email', { required: true, masked: true })`) alongside the Java-style chained methods. Both styles compile to identical protos.

2. A conformance suite (`sdk-conformance/`) proves the parity mechanically. An SDK here is really a compiler. User code goes in, one `PutWfSpecRequest` comes out, and the server only ever sees the proto. The suite freezes 147 expected answers across three areas (workflow compilation, value serde, side registrations), minted from sdk-java, and grades every SDK against them. sdk-java takes the same exam, so drift in the reference reddens its own column. Both SDKs pass 147/147 today (`sdk-conformance/MATRIX.md`).

3. Trust layers beyond the corpus. A seeded random dual-compile requires both SDKs to build identical workflows with no shared canon (`sdk-conformance/FUZZ.md`). The `examples/js/` directory mirrors the Java example set as 32 long-lived workers, and `run-all.mjs` drives each one against a live server to a COMPLETED run. There are 61 unit tests and 32 integration tests against a real server in docker.

Nothing here gates anyone. The conformance CI job is report-only, and the sdk-js test job gates only sdk-js, like every other SDK's job.

## Run it yourself

```bash
node sdk-conformance/runner/build.mjs
node sdk-conformance/runner/suite.mjs
```

The first command runs each SDK's own build. The second runs the whole suite and prints one report. `sdk-conformance/QUICKSTART.md` also walks through breaking the suite on purpose and watching it catch you, which is the fastest way to trust it.

## How to review without reading 273 fixture files

- Start with `sdk-conformance/README.md`, then `QUICKSTART.md` when you want to run it.
- The real review surface is the case definitions (`sdk-java/conformance/.../WfsdkArea.java` and their sdk-js twins) and the sdk-js source. The fixtures are sdk-java's mechanical answers. Spot-check a few and let the mint and freshness gates worry about the rest.
- Every guarantee is wired to something that can fail. Coverage ratchets run in both directions, probe pairs have a vacuity guard, ledgers ratchet, and the matrix has a staleness check. Each was proven able to go red before landing.

## The numbers

- wfsdk, 119 cases covering 123 of 144 reflected public capabilities (21 excused, each with a written reason)
- serde, 20 cases covering 8 of 12 `VariableValue` arms (4 excused, each naming its pending design question)
- registrations, 8 cases covering 10 of 10 capabilities
- fuzz, no corpus by design, cross-SDK agreement on random workflows is the verdict
- examples, the Java set mirrored as long-lived JS workers, driven end to end by `examples/js/run-all.mjs`

## Release notes 1.3 should carry

- `LHConfig.from()` now reads `LHC_*` env vars as the base layer, and explicit args win. 1.2.x ignored env. This matches Java.
- JSON task outputs now strip null object fields, matching Java. Workers that emitted explicit nulls will see different stored documents.
- OAuth tokens are now minted per call instead of frozen at client creation, so long-lived clients and workers survive token expiry.
- Known edge, documented. JS numbers lose int64 precision past 2^53. The `int-beyond-double` conformance case makes the loss visible, and `bigint` is the exact escape hatch.

## Not in 1.3

Worker-runtime conformance (a future area). The worker is covered by the ported implementation, the integration tier, and the live examples. The four unexamined serde arms are inventoried in the corpus exemption files with reasons.


proposals/sdk-conformance/adoption.md
bryson-g: the proposal diffs don't have to be included if it's unnecessary to keep. making a proposal here was more-so for keeping track of my own comprehensions.

Snarr (approved):

coltmcnealy-lh (approved):

PR #2506, opened 2026-09-03
perf(server): reduce cold reads on DeleteWfRunRequests

coltmcnealy-lh:
Kafka Streams KeyValueStore#delete() is a get-then-delete. We don't do anything with the value that it gets, so we circumvent this costly behavior (which reads cold blocks from the disk most times) by instead using put(key, null).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -134,7 +134,7 @@ jobs:
           java-version: 25
 
       - name: E2E Tests
-        run: ./gradlew server:e2e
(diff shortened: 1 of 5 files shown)


coltmcnealy-lh:
This will prevent reading the `Tag` objects during cold deletes. Due to the way our soak test is structured, the savings won't be massive, because all of the tags are sequentially ordered and therefore on adjacent blocks. But a real world use-case would have a higher performance gain from this change.

Snarr (approved):
Good change thanks

Issue #1949, opened 2025-11-14
Improve comparator support for non-standard VariableValue cases


Snarr:
### Background

1. In the `WfSpec` DSL users can write expressions comparing two variables. Right now, that is done using the `condition(Object lhs, Comparator comparator, Object rhs)` method though that may change
2. Under the hood, we use the Java `Comparable` interface on the LH Server to compare `VariableValue`s
3. Not all `VariableValue`s use a Java `Comparable`-compatible class. For instance, `VariableValue.TIMESTAMP` uses the `com.google.protobuf.Timestamp` class which doesn't implement Comparable. This means that comparing Timestamps will result in a `VAR_SUB_ERROR`.

### Proposal

1. We support comparing `Timestamp`s, `Struct`s,`WfRunId`s, `JsonObj`s, `JsonArr`s, and `Bytes`, all of throw errors when you try to compare them currently.

2. We should also support a limited subset of comparators depending on the `VariableValue` case.
For example, `Timestamp`s can be compared using `LESS_THAN` and `GREATER_THAN` comparators, but it doesn't make sense to compare entire `Struct`s using those comparators. For comparing entire `Structs`, it only makes sense to support `EQUALS` and `NOT_EQUALS` comparators.

### Acceptance Criteria

1. We should support identity comparators (`EQUALS` and `NOT_EQUALS`) for `Timestamp`s, `Struct`s,`WfRunId`s, `JsonObj`s, `JsonArr`s
2. We should support magnitude comparators (`LESS_THAN`, `GREATER_THAN_EQ`, etc...) for `Timestamp`s
3. We should reject improper comparators at compile time and throw descriptive `VAR_SUB_ERROR`s if they somehow make it into runtime

### Affected Components

Kernel (/server)

PR #2509, opened 2026-09-04
feat(server): add Map comparison logic, ignore entry order

Snarr:
This PR adds logic to the Comparer class to support Map comparison in LittleHorse. This allows users to compare Map equality in WfSpecs. At the protobuf level Maps store `Entry`s in an ordered list but in this implementation we ignore the ordering during Map comparison.

### DIFF
--- a/server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/node/Comparer.java
+++ b/server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/node/Comparer.java
@@ -6,7 +6,9 @@ import io.littlehorse.common.model.getable.core.variable.MapModel;
 import io.littlehorse.common.model.getable.core.variable.VariableValueModel;
 import io.littlehorse.common.util.LHUtil;
 import io.littlehorse.sdk.common.proto.VariableValue.ValueCase;
+import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


coltmcnealy-lh (approved):

server/src/main/java/io/littlehorse/common/model/getable/global/wfspec/node/Comparer.java
```diff
@@ -18,6 +20,10 @@ public static int compare(VariableValueModel left, VariableValueModel right) thr
             if (right.getVal() == null && left.getVal() != null) return 1;
             if (right.getVal() == null && left.getVal() == null) return 0;
 
+            if (left.getValueType() == ValueCase.MAP && right.getValueType() == ValueCase.MAP) {
```
Snarr, line 23: There's some tech debt with `Comparer` here... a while back during a tech minga I think Christian and I worked on removing this tech debt but haven't gotten back around to it. Tracking this in issue: https://github.com/littlehorse-enterprises/littlehorse/issues/1949

Snarr: Cleaning up this tech debt is of scope for this PR

PR #2467, opened 2026-08-16
feat(server): track quota usage in a grpc-facing metric

coltmcnealy-lh:

PR #2510, opened 2026-09-04
feat(dashboard): add quota usage metrics page

bryson-g:
Follow-up to #2467, which added the quota usage metrics but no way to see them outside `lhctl list quotaMetrics`.

`/{tenantId}/quotas` shows requests observed vs throttled over time, throttle rate, and how much backoff the server handed out, with the configured limit drawn on the chart. Same shape as the WfSpec and TaskDef metrics pages, and it reuses their bucketing helpers.

The page shows the quota you're actually subject to: your principal's own quota if it has one, otherwise the tenant-wide quota. It resolves that with `GetQuota` by id rather than listing quotas, so there's no picker to choose someone else's.

Reaching the page is a `Quotas` item in the principal dropdown. Not a Metadata Search tab, since a quota isn't really metadata you browse.


dashboard/src/app/(authenticated)/[tenantId]/quotas/actions/getApplicableQuota.ts
```diff
@@ -0,0 +1,43 @@
...
+}
+
+/** Reads quotas by id rather than listing them, so a principal only ever sees the quota they are subject to. */
+export const getApplicableQuota = async ({ tenantId }: WithTenant): Promise<ApplicableQuota> => {
+  const client = await lhClient({ tenantId })
+  const principalId = (await client.whoami({})).id?.id
+
+  if (principalId) {
+    const principalQuota = await getQuotaOrNull(client, { tenant: { id: tenantId }, principal: { id: principalId } })
+    if (principalQuota) return { scope: 'principal', quota: principalQuota, principalId }
```
Comment (Snarr) [inline dashboard/src/app/(authenticated)/[tenantId]/quotas/actions/getApplicableQuota.ts]: Let's say the tenantQuota is 2 writes per second, and the principalQuota is 50 writes per second. In this code, aren't we preferring to return the principalQuota first and not returning the tenantQuota? But then what happens if the server throttling the tenant at 2 writes per second then, and the dashboard user only sees "50 writes per second" as their given quota, wouldn't that lead to some confusion? Am I understanding this correctly?

Snarr (approved):
Reviewed code, tested locally, and looks great. Latest changes that include both Tenant and Principal quota graphs are good 👍

PR #2523, opened 2026-09-12
fix(server): do not publish bulkjob shard progress every punctuation

coltmcnealy-lh:
This fixes the root cause of several outages in my soak cluster with bulk jobs spamming the metadata processor with bulk shard updates. Instead, we push updates every 60 seconds.

### DIFF
--- a/server/src/main/java/io/littlehorse/common/LHServerConfig.java
+++ b/server/src/main/java/io/littlehorse/common/LHServerConfig.java
@@ -1018,7 +1018,7 @@ public class LHServerConfig extends ConfigBase {
 
         // We want a request to be able to fail and be handled (if non-fatal) before a transaction times out.
         // Therefore, request timeout should be less than transaction timeout / session timeout.
-        props.put("request.timeout.ms", (int) (getStreamsSessionTimeout() * 0.75));
+        props.put("request.timeout.ms", (int) (getStreamsSessionTimeout() * 0.30));
         props.put("producer.acks", "all");
         props.put("replication.factor", (int) getReplicationFactor());
         props.put("num.standby.replicas", Integer.valueOf(getOrSetDefault(NUM_STANDBY_REPLICAS_KEY, "0")));
--- a/server/src/main/java/io/littlehorse/server/streams/topology/core/processors/BulkJobScanJob.java
+++ b/server/src/main/java/io/littlehorse/server/streams/topology/core/processors/BulkJobScanJob.java
@@ -24,9 +24,12 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BooleanSupplier;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.kafka.streams.processor.api.Record;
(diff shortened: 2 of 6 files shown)


server/src/main/java/io/littlehorse/server/streams/topology/core/processors/BulkJobScanJob.java
```diff
@@ -59,6 +62,12 @@ public class BulkJobScanJob implements PartitionBackgroundJob<CommandProcessorOu
 
     private static final Duration INTERVAL = Duration.ofSeconds(1);
 
+    /**
+     * BulkJob is metadata, so every shard report is sent through the single metadata partition.
+     * Keep the most recent progress in memory and publish it at a bounded rate.
+     */
+    private static final Duration PROGRESS_REPORT_INTERVAL = Duration.ofMinutes(1);
```
Snarr, line 69: PR description says we push updates every 30s but this says 1m, is this ok?

server/src/main/java/io/littlehorse/server/streams/topology/core/processors/BulkJobScanJob.java
```diff
@@ -173,35 +196,93 @@ record -> ctx.forward(uncheckedCast(record)),
...
+        Instant now = clock.get();
+        List<String> dueReports = shardProgress.keySet().stream()
+                .filter(key -> isReportDue(key, now))
+                .toList();
+
+        for (String key : dueReports) {
+            ctx.forward(toRecord(shardProgress.get(key)));
+        }
+        if (!dueReports.isEmpty()) {
+            ctx.submit();
```
Snarr, line 220: Nitpick but if `ctx.submit()` takes long to submit due to backpressure, the `now` instant could drift far from the actual submission time. Perhaps on line 221 we could refetch the `clock.get()` time when we put the `lastReportAt` entries? `clock.get()` should be an inexpensive call, so I don't see why not

server/src/main/java/io/littlehorse/common/LHServerConfig.java
```diff
@@ -1018,7 +1018,7 @@ private Properties getBaseStreamsConfig() {
 
         // We want a request to be able to fail and be handled (if non-fatal) before a transaction times out.
         // Therefore, request timeout should be less than transaction timeout / session timeout.
-        props.put("request.timeout.ms", (int) (getStreamsSessionTimeout() * 0.75));
+        props.put("request.timeout.ms", (int) (getStreamsSessionTimeout() * 0.30));
```
eduwercamacaro: I recommend putting this into a separate PR with its own description and discussion.

eduwercamacaro (approved):

server/src/main/java/io/littlehorse/server/streams/topology/core/processors/BulkJobScanJob.java
eduwercamacaro: This is correct. Under certain circumstances, the drift could make each iteration send a single report.

eduwercamacaro: It is important to add a test case that verifies this behavior. We can simulate that on a unit test by making the `ctx.submit` block for `>PROGRESS_REPORT_INTERVAL`

server/src/main/java/io/littlehorse/server/streams/topology/core/processors/BulkJobScanJob.java
coltmcnealy-lh: Fixed desc. Thanks

server/src/main/java/io/littlehorse/server/streams/topology/core/processors/BulkJobScanJob.java
coltmcnealy-lh: Sending a report on each iteration isn't necessarily the worst, because it is still only one message per minute, which won't over-burden the metadata processor. So I'm not sure we need to guard against that case.

That said, I agree with Jacob's comment.

server/src/main/java/io/littlehorse/common/LHServerConfig.java
coltmcnealy-lh: Agree. I was doing this for my soak cluster. Thanks!

PR #2540, opened 2026-09-18
fix(server): process matured sleep nodes when threadRunNumber > threa…

### LINKED ISSUE #2539: Processing Matured Sleep Nodes
### Context

If you create a workflow spec that spawns child threads, when the number of Thread Runs exceeds the 90% percent threshold of active Thread Runs the server will archive completed Thread Runs, removing them from `threadRunsUseMeCarefully`.

When processing a matured sleep node, the server check this condition:

`    
if (threadRunNumber >= threadRunsUseMeCarefully.size() || threadRunNumber < 0) {
            throw new LHValidationException(null, "Reference to nonexistent thread.");
        }
`

This condition checks if the `threadRunNumber` of the Thread Run that has a matured sleep node and compares it to the size of `threadRunsUseMeCarefully`. The bug arises when Thread Runs have been archived and removed from `threadRunsUseMecarefully`. After archiving Thread Runs the size of the `threadRunsUseMeCarefully` can be less than threadRunNumbers being evaluated.

### Steps To Reproduce

```
 return new WorkflowImpl("matured-sleep-node-bug", wf -> {
            Integer[] values = java.util.stream.IntStream.range(0, 100)
        .boxed()
        .toArray(Integer[]::new);

        WfRunVariable arrVar = wf.declareArray("arr", Integer.class)
            .withDefault(values);
        
          wf.spawnThreadForEach(arrVar, "child-thread", thread ->{
                thread.sleepSeconds(1);
            });
        });
    }
```

### Expected Behavior

Matured Sleep nodes should be processed allowing the Thread Run advance and continue execution.

### Affected Components

Kernel (/server)

### Environment

Server Version: 3.0.0

### Documentation

_No response_

### Additional Context

_No response_

JakeRose03:
This PR provides a fix for #2539

## Problem
Matured sleep nodes were not being processed when the `threadRunNumber` was greater than the size of `threadRunsUseMeCarefully`. This condition would evaluate to true after the server reached the active thread run threshold of 90% and archived completed thread runs.

## Fix
Remove the condition stated above.

## Testing

Added `shouldAdvanceSleepNodesWhenMaturedWArchivedThreadRuns` which spawns a `WfRun` that exceeds the active thread run capacity and forces completed `ThreadRuns` to be archived.  The test then asserts that the `WfRun` completes as expected, this ensures that every Sleep Node  has been completed, allowing all child threads to complete and in return allowing the `WfRun` to complete.

Closes #2539


eduwercamacaro:
Looking at the WfRunModel class, I believe a similar behavior could arise when processing external events. 

The reason is because inside the `WfRunModel.processExternalEvent` method we iterate over the `threadRunsUseMeCarefully` collection to "notify" every existing thread run.

eduwercamacaro (approved):

